PR #370, opened 2022-04-27
feat: websocket endpoint

mathnogueira:
This PR adds an endpoint to handle Websocket connections

## Changes

- This only adds a `/ws` endpoint that does nothing yet;
- It creates the structure to handle message routing based on the message `type` attribute.

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/go/websocket/router.go
+++ b/server/go/websocket/router.go
@@ -1,0 +1,47 @@
+package websocket
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/gorilla/websocket"
+)
+
+type MessageExecutor func(*websocket.Conn, Message)
+
+type Router struct {
+	routes map[string]MessageExecutor
+}
+
+func NewRouter() *Router {
+	return &Router{
+		routes: make(map[string]MessageExecutor),
+	}
+}
+
+func (r *Router) Add(messageType string, executor MessageExecutor) {
+	r.routes[messageType] = executor
+}
+
+func (r *Router) ListenAndServe(addr string) {
+	routingFunction := func(conn *websocket.Conn, message []byte) {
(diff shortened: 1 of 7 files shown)


schoren (approved):
this looks awesome!

server/go/websocket/router.go
```diff
@@ -0,0 +1,47 @@
...
+	return &Router{
+		routes: make(map[string]MessageExecutor),
+	}
+}
+
+func (r *Router) Add(messageType string, executor MessageExecutor) {
+	r.routes[messageType] = executor
+}
+
+func (r *Router) ListenAndServe(addr string) {
```
xoscar, line 28: noob question, where does r come from here? Is it somehow getting injected?

server/go/websocket/router.go
```diff
@@ -0,0 +1,47 @@
...
+
+type MessageExecutor func(*websocket.Conn, Message)
+
+type Router struct {
+	routes map[string]MessageExecutor
+}
+
+func NewRouter() *Router {
+	return &Router{
+		routes: make(map[string]MessageExecutor),
```
xoscar, line 20: Another noob one. What are the route names/message types that would be available here?

server/go/websocket/router.go
schoren: The function signature means that the function is attached to the struct `Router`. That's the go way to create something similar to methods. You can think of the `r` as `this` (though not quite the same).  You can read [a more precise definition](https://go.dev/tour/methods/1)

server/go/websocket/router.go
schoren: What that code means is that you will have a way to map a message type to an "object" that knows how to handle that specifc type

PR #371, opened 2022-04-27
feat: add subscription module

mathnogueira:
This PR adds a module responsible for handling subscriptions and notifications in tracetest. This will be used to send updates from tracetest to the websocket endpoint.

## Changes

- Add a new isolated module
- Enable messages to be sent from a place to another without coupling them. It is based on event-driven programming.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/go/subscription/manager.go
+++ b/server/go/subscription/manager.go
@@ -9,15 +9,7 @@ type Manager struct {
 
 var managerInstance *Manager = nil
 
-func GetManager() *Manager {
-	if managerInstance == nil {
-		managerInstance = newManager()
-	}
-
-	return managerInstance
-}
-
-func newManager() *Manager {
+func NewManager() *Manager {
 	return &Manager{
 		subscriptions: make(map[string][]Subscriber),
 		mutex:         sync.Mutex{},
--- a/server/go/subscription/manager.go
+++ b/server/go/subscription/manager.go
@@ -7,8 +7,6 @@ type Manager struct {
 	mutex         sync.Mutex
 }
 
-var managerInstance *Manager = nil
-
 func NewManager() *Manager {
 	return &Manager{
 		subscriptions: make(map[string][]Subscriber),
--- a/server/go/subscription/manager.go
+++ b/server/go/subscription/manager.go
@@ -38,14 +38,14 @@ func (m *Manager) Subscribe(resourceID string, subscriber Subscriber) {
 	m.subscriptions[resourceID] = array
 }
 
-func (m *Manager) Unsubscribe(resourceID string, subscriber Subscriber) {
+func (m *Manager) Unsubscribe(resourceID string, subscriptionID string) {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 
 	if array, ok := m.subscriptions[resourceID]; ok {
 		newArray := make([]Subscriber, 0, len(array)-1)
 		for _, item := range array {
-			if item.ID() != subscriber.ID() {
+			if item.ID() != subscriptionID {
 				newArray = append(newArray, item)
 			}
 		}
--- a/server/go/subscription/manager.go
+++ b/server/go/subscription/manager.go
@@ -53,7 +53,7 @@ func (m *Manager) PublishUpdate(resourceID string, message Message) {
 
 	if subscribers, ok := m.subscriptions[resourceID]; ok {
 		for _, subscriber := range subscribers {
-			(subscriber).Notify(&message)
+			subscriber.Notify(message)
 		}
 	}
 }
--- a/server/go/subscription/manager.go
+++ b/server/go/subscription/manager.go
@@ -1,0 +1,67 @@
+package subscription
+
+import "sync"
+
+type Manager struct {
+	subscriptions map[string][]Subscriber
+	mutex         sync.Mutex
+}
+
+var managerInstance *Manager = nil
+
+func GetManager() *Manager {
+	if managerInstance == nil {
+		managerInstance = newManager()
+	}
+
+	return managerInstance
+}
+
+func newManager() *Manager {
+	return &Manager{
+		subscriptions: make(map[string][]Subscriber),
+		mutex:         sync.Mutex{},
+	}
+}
+
+func (m *Manager) Subscribe(resourceID string, subscriber Subscriber) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	array := make([]Subscriber, 0)
+
+	if existingArray, ok := m.subscriptions[resourceID]; ok {
+		array = existingArray
+	}
+
+	array = append(array, subscriber)
+	m.subscriptions[resourceID] = array
+}
+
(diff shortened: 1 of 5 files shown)


server/go/subscription/manager.go
```diff
@@ -0,0 +1,67 @@
...
+import "sync"
+
+type Manager struct {
+	subscriptions map[string][]Subscriber
+	mutex         sync.Mutex
+}
+
+var managerInstance *Manager = nil
+
+func GetManager() *Manager {
```
schoren, line 12: I don't think it's a great idea to have a singleton. It would be better to have a `subscription.NewManager()` func, and pass that around as a dependency. We could even export a `Manager` interface (and have the struct unexported) so that the code depending on this is not coupled to the actual implementation.

server/go/subscription/manager.go
```diff
@@ -0,0 +1,67 @@
...
+
+	array = append(array, subscriber)
+	m.subscriptions[resourceID] = array
+}
+
+func (m *Manager) Unsubscribe(resourceID string, subscriptionID string) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	if array, ok := m.subscriptions[resourceID]; ok {
```
schoren, line 45: We could reduce the complexity here by inverting the condition like: 
```
array, exists := m.subscriptions[resourceID]
if !exists {
	return
}

newArray := make([]Subscriber, 0, len(array)-1)
	for _, item := range array {
		if item.ID() != subscriptionID {
			newArray = append(newArray, item)
		}
	}

	m.subscriptions[resourceID] = newArray
```

server/go/subscription/manager.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+		}
+
+		m.subscriptions[resourceID] = newArray
+	}
+
+}
+
+func (m *Manager) PublishUpdate(resourceID string, message Message) {
+	m.mutex.Lock()
```
schoren, line 59: why is this lock needed? I _think_ that since this func doesn't write anything, it can work without lock

server/go/subscription/manager.go
```diff
@@ -0,0 +1,67 @@
...
+	defer m.mutex.Unlock()
+
+	if subscribers, ok := m.subscriptions[resourceID]; ok {
+		for _, subscriber := range subscribers {
+			(subscriber).Notify(&message)
```
schoren, line 64: why the parentheses?

server/go/subscription/watcher.go
```diff
@@ -0,0 +1,28 @@
+package subscription
+
+import "github.com/google/uuid"
+
+type Subscriber interface {
+	ID() string
+	Notify(message *Message) error
```
schoren, line 7: I think it's really dangerous to have the message as a pointer. It makes the message mutable and thread unsafe. Is it really neccesary?

server/go/subscription/watcher.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type SubscriberFunction struct {
+	id       string
+	function func(*Message) error
```
schoren, line 12: same here

server/go/subscription/watcher.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (sf *SubscriberFunction) ID() string {
```
schoren, line 15: again, why does this need to work on a pointer?

schoren (changes requested):

server/go/subscription/manager.go
mathnogueira: If you read and write to a map concurrently, golang might throw this error:

```
fatal error: concurrent map read and map write
```

You can [try it here](https://go.dev/play/p/TIZHgEpME_2)

server/go/subscription/manager.go
```diff
@@ -0,0 +1,67 @@
...
+func newManager() *Manager {
+	return &Manager{
+		subscriptions: make(map[string][]Subscriber),
+		mutex:         sync.Mutex{},
+	}
+}
+
+func (m *Manager) Subscribe(resourceID string, subscriber Subscriber) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
```
xoscar, line 29: Does defer mean doing this after the rest of the code statements are executed?

server/go/subscription/manager.go
```diff
@@ -0,0 +1,67 @@
+package subscription
+
+import "sync"
+
+type Manager struct {
+	subscriptions map[string][]Subscriber
```
xoscar, line 6: Isn't something like `topics` a better match than subscriptions?

server/go/subscription/manager_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/kubeshop/tracetest/server/go/subscription"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestManagerIsSingleton(t *testing.T) {
+	manager1 := subscription.GetManager()
+	manager2 := subscription.GetManager()
+
+	assert.Equal(t, manager1, manager2, "managers should be the same instance")
```
xoscar, line 14: Noob question, when doing an assertion like this, is it done by reference or by something else?

server/go/subscription/manager.go
schoren: Ahh ok. ignore this then!

server/go/subscription/manager.go
schoren: exactly. It is executed just before returning. [More details](https://go.dev/tour/flowcontrol/12)

server/go/subscription/manager.go
schoren: I'm not sure what you mean, but this `observer` patter is pretty standard for this kind of things. If I understand you correctly, `topics` would mean to have different queues, much like in Kafka. That is a lot more complex, but in essence it's just a complicated observer

server/go/subscription/manager_test.go
schoren: In go, unless specified, everything is passed by value. In this case, `manager1` and `manager2` are two potentially different pointers. Meaning that if the singleton is wrong, you would end up with two different pointers. 
So, this is passing `a pointer to a manager`, as the value. Not sure if that makes sense?

server/go/subscription/watcher.go
mathnogueira: Very good point.

server/go/subscription/manager.go
```diff
@@ -0,0 +1,61 @@
...
+	mutex         sync.Mutex
+}
+
+var managerInstance *Manager = nil
```
schoren: this var is now unused

schoren (approved):

server/go/subscription/manager.go
mathnogueira: Just fixed it :smile:

PR #377, opened 2022-04-27
New architecture

schoren:
This PR is a proposal of a new architecture for the go server. The main problem with the previous approach is that we were using openapi as a "framework" of sorts, writing our code around the autogenerated models and servers.
With this change, the autogenerated code is just a library, so we can write the code however we like, and only consume the generated models as a dependency when neccesary.

This has a few nice side effects:
- I was able to customize the package name, so it looks nicer both for the server package and the openapi package
- openapi is now just a lib, meaning we can totally change the way we interact with the server, including adding a CLI interface, gRPC, whatever
- we have less autogenerated garbage code, like unused dockerfiles, etc.
- having a more clear project layout makes it easier to identify problems, or find code that can be reused. For example: I've created an UUID/traceID/SpanID generator object that could be reused, but even I forgot that it was there. Now it lives in its own package at the root, so it's really visible.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


povilasv (approved):
Awesome idea!

You can use `.openapi-generator-ignore` and various options to generator, which might help with the generate code:

https://github.com/OpenAPITools/openapi-generator/blob/master/docs/customization.md

docker-compose.yaml
```diff
@@ -1,9 +1,27 @@
...
+    container_name: tracetest
+    image: kubeshop/tracetest
+    build: .
+    volumes:
+      - ${PWD}/server/config.yaml:/app/config.yaml
+    ports:
+      - 8080:8080
+      - 8081:8081
+    depends_on:
+      - postgres
```
mathnogueira, line 13: Only this doesn't ensure the database will be ready to accept connections. So `docker-compose up` might fail. To prevent that, you can add a [helthcheck routine to the database](https://github.com/kubeshop/pokeshop/blob/master/api/docker-compose.yml#L14) and then [update the depends_on](https://github.com/kubeshop/pokeshop/blob/master/api/docker-compose.yml#L78) configuration

server/testdb/testdb.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/kubeshop/tracetest/openapi"
+)
+
+var ErrNotFound = errors.New("record not found")
+
+type Tests interface {
```
mathnogueira, line 12: I don't think those `Tests`, `Results`, and `Assertions` names are good. They look like a definition of a model, not a repository. 

Could we add something else in their names to ensure it is explicit their function? Like `TestsRepository`.

server/tracedb/jaegerdb.go
```diff
@@ -1,25 +1,24 @@
...
 )
 
-type JaegerTraceDB struct {
+type jaegerTraceDB struct {
 	conn  *grpc.ClientConn
 	query pb.QueryServiceClient
 }
 
-func New(config *configgrpc.GRPCClientSettings) (tracedb.TraceDB, error) {
+func newJaegerDB(config *configgrpc.GRPCClientSettings) (*jaegerTraceDB, error) {
```
mathnogueira, line 21: any reason for returning a `*jaegerTraceDB` instead of `TraceDB`? Just like `TempoTraceDB` is doing.

server/tracedb/tracedb.go
```diff
@@ -0,0 +1,31 @@
...
+func New(c config.Config) (db TraceDB, err error) {
+	err = ErrInvalidTraceDBProvider
+	switch {
+	case c.JaegerConnectionConfig != nil:
+		db, err = newJaegerDB(c.JaegerConnectionConfig)
+	case c.TempoConnectionConfig != nil:
+		db, err = newTempoDB(c.TempoConnectionConfig)
+	}
+
+	return
```
mathnogueira, line 30: Nice

server/http/controller.go
```diff
@@ -0,0 +1,235 @@
...
+	runner  executor.Runner
+}
+
+func NewController(traceDB tracedb.TraceDB, testDB testdb.Database, runner executor.Runner) openapi.ApiApiServicer {
+	return &controller{
+		traceDB: traceDB,
+		testDB:  testDB,
+		runner:  runner,
+	}
+}
```
mathnogueira, line 27: I'm ok with this now, but I think we should split this controller in multiple controllers in the future.

mathnogueira (commented):
I've added some comments, but I really liked the change. Good job!

server/tracedb/jaegerdb.go
schoren: Nope, just forgot to update it

server/tracedb/tracedb.go
schoren: Still don't like this approach, we should use a more flexible pattern, but I didn't want to change too much

server/http/controller.go
schoren: totally agree, this is just a first step. This controller should be only a wrapper.

docker-compose.yaml
schoren: I didn't know you could set conditions for depends. thanks!

PR #392, opened 2022-05-02
add: use go-migrate to manage db migrations

mathnogueira:
This PR adds go-migrate to handle database changes instead of creating the database schema directly.

## Changes

- Adds database versioning
- Execute migrations automatically at the moment the server starts
- Uses alpine linux instead of ubuntu for the base image (image is less than half the original size)
image

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


web/. dockerignore
```diff
@@ -1,4 +1,3 @@
 node_modules
 build
-.dockerignore
```
mathnogueira, line 3: I've removed it because of [this](https://codefresh.io/docker-tutorial/not-ignore-dockerignore-2/)

xoscar (approved):
GJ @mathnogueira 🚀

Issue #400, opened 2022-05-03
Add dropdown actions to each test table row with the delete option


xoscar:
For this page
![Screen Shot 2022-05-03 at 7 51 08 a m](https://user-images.githubusercontent.com/11051031/166456079-c8bccf54-06b4-4ec9-86f3-c3837bd4e842.png)

We need to add a three dots option to the right side of each row with a delete option. 

Work with the backend team @mathnogueira & @schoren to see if the delete endpoint is available and if so integrate it when the user selects the option.

xoscar:
Design https://www.figma.com/file/0MPQK2Icv7FvqHJnlColpv/MVP?node-id=382%3A3657

kdhamric:
Couple questions:
 - does this include the backend work to delete a test?  
 - when you do delete a test on the backend, are you actually deleting the test and test results, or marking them as deleted?  I would recommend marking them as deleted, but not actually removing them (just not showing them any longer).

kdhamric:
I do not see the answers to my questions above?

Also - it looks like the delete action was added to the test runs, not the tests.

kdhamric:
https://www.loom.com/share/f67e6715aa83416a8bebc7e48334e64a

kdhamric:
How are we doing this:
when you do delete a test on the backend, are you actually deleting the test and test results, or marking them as deleted?

xoscar:
I think the current implementation is deleting the tests from the DB, we might need to alter the schema to support soft delete @mathnogueira

cescoferraro:
@kdhamric the current implementation on the backend deletes the test. I have talked with @schoren about this.
he sad for "soft" deletion will be implemented in the future.

PR #407, opened 2022-05-03
adds delete test actions

cescoferraro:
This PR...

## Changes

Add an actions column to the test table, this column contains a dropdown with the delete test action. 

## Fixes

- https://github.com/kubeshop/tracetest/issues/400

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/pages/Home/TestList.tsx
+++ b/web/src/pages/Home/TestList.tsx
@@ -7,13 +7,14 @@ import CustomTable from '../../components/CustomTable';
 import HomeAnalyticsService from '../../services/Analytics/HomeAnalytics.service';
 import NoResults from './NoResults';
 import {ITest} from '../../types/Test.types';
+import {useMenuDeleteCallback} from './useMenuDeleteCallback';
 
 const {onTestClick} = HomeAnalyticsService;
 
 const TestList = () => {
   const navigate = useNavigate();
   const eventRef = useRef<{previousPageX: number; currentPageX: number}>({previousPageX: 0, currentPageX: 0});
-  const {data: testList = [], isLoading, refetch} = useGetTestListQuery();
+  const {data: testList = [], isLoading} = useGetTestListQuery();
 
   const [deleteTestMutation] = useDeleteTestByIdMutation();
 
@@ -28,7 +29,7 @@ const TestList = () => {
       }
     }
   };
-
+  const onDelete = useMenuDeleteCallback(deleteTestMutation);
   return (
     <CustomTable
       scroll={{y: 'calc(100vh - 300px)'}}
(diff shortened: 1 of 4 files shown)


web/src/pages/Home/TestList.tsx
```diff
@@ -62,6 +65,37 @@ const TestList = () => {
...
+        key="actions"
+        align="right"
+        render={i => {
+          return (
+            <Dropdown
+              overlay={
+                <Menu>
+                  <Menu.Item
+                    onClick={async e => {
+                      e.domEvent.stopPropagation();
```
xoscar, line 79: What do you think about moving this to a function? like `onDelete`?

web/src/redux/apis/Test.api.ts
```diff
@@ -38,7 +38,9 @@ const TestAPI = createApi({
       query: id => `/tests/${id}`,
       providesTags: result => [{type: 'Test', id: result?.testId}],
     }),
-
+    deleteTestById: build.mutation<ITest, string>({
```
xoscar, line 41: I think we don't have to run the refetch, we can just invalidate the tag here that is going to automatically trigger the refetch of the entire list.

xoscar (changes requested):
Hey @cescoferraro gj! I left some comments and concerns. Let me know what you think 😄

web/src/pages/Home/TestList.tsx
cescoferraro: done!

web/src/redux/apis/Test.api.ts
cescoferraro: makes sense. missed this one. but done

web/src/pages/Home/useMenuDeleteCallback.tsx
```diff
@@ -0,0 +1,15 @@
+import {useCallback} from 'react';
+import {MenuInfo} from 'rc-menu/lib/interface';
+import {ITest} from '../../types/Test.types';
+
+export function useMenuDeleteCallback(
```
xoscar, line 5: I don't think we need a new file, just having this function above the return section should be enough 😄

xoscar (approved):
Hey @cescoferraro I left a small comment. I'm approving

web/src/pages/Home/useMenuDeleteCallback.tsx
cescoferraro: All the projects that I have been working lately follow this standard. 
Having a single components/function per file should be something to aim for.
Abstracting hooks to different files also helps with readability of a component render method. 
Also help with its file size. React components should not be bigger than 50 lines.
Should I merge this? whats the correct procedure here? @xoscar

cescoferraro: @xoscar I think we should chat and come up with a Frontend standard to follow from now on

xoscar: I think components should have the basic functionality being part of it, then it would be harder to follow up on what is going on. I agree with you that we have to keep components small tho. That's why most of the business logic is encapsulated in the services and models.

Issue #399, opened 2022-05-03
Use openAPI generated client in frontend


xoscar:

xoscar:
By reviewing https://redux-toolkit.js.org/rtk-query/usage/code-generation and trying to include it into the repo I found out it is not that helpful as we still need to do some manual changes to support some of the parsing and tag invalidations.

I'm going to look at a way to introduce autogenerated types tho.

xoscar:
Findings:
1. The main package https://github.com/OpenAPITools/openapi-generator doesn't have a tool to only generate only types, it only supports creating the whole client. 
2. As we are using redux toolkit and RTK query any of the clients supported match what we need.
3. I tried using other tools like https://github.com/drwpow/openapi-typescript but the issue is that the rest of the packages like this only support openapi specs to openappi specs and we are mixing openapi specs and swagger specs 😢  this doesn't allow the proper types to be generated.
4. The trace.yaml swagger file is autogenerated which makes it not viable to be manually updated or changed to openapi.

Conclusion:
After spending some time on this I think it is better to keep continue with the current setup we have and then in the future come back to this issue as it might only slow us down with not much benefit atm. I think the OpenAPI generators are still quite green and not ready for our use case.

PR #408, opened 2022-05-03
399 using types from the open api specs

xoscar:
This PR uses the Open API specs as a base to generate all of the base types within the web application.

## Changes

- Deleted a bunch of constants that are now replaced by generated types.
- Adding extra checks for optional parameters.

## Fixes

- https://github.com/kubeshop/tracetest/issues/399
- The Open API specs really got behind and I had to fix some of them.

## Checklist

- [x] tested locally
- [x] added new dependencies
- [x] updated the docs
- [ ] added a test


api/openapi.yaml
```diff
@@ -319,6 +319,10 @@ components:
   schemas:
     Test:
       type: object
+      required:
```
schoren, line 322: Why are you using required here? I tried to do this in the backend, and the validation is stupidly strict, like a bool `false` is parsed as `empty` and fails the validation in the backend. I would leave this as it was, unless there is a blocking need for it

api/openapi.yaml
```diff
@@ -446,30 +481,56 @@ components:
...
+              intValue,
+              boolValue,
+              doubleValue,
+              kvlistValue,
+            ]
         operator:
           type: string
           description: Example "EQUALS"
-          enum: [EQUALS, LESSTHAN, GREATERTHAN, NOTEQUALS]
+          enum: [EQUALS, NOTEQUALS, LESSTHAN, GREATERTHAN, GREATOREQUALS, LESSOREQUAL, CONTAINS]
```
schoren, line 533: NOTE: This is fine, but it will become obsolete once we move the assertion engine to the backend.

schoren (changes requested):

api/openapi.yaml
xoscar: Yep, I did this to match what we currently have in the FE

schoren:
the file `api/traces.yaml` is generated from the otel proto definitions, with the `make swagger` command. The backend relies heavily on that, so at this time it would take a lot of effor to rewrite that definition. I've pushed a commit reverting that file (as well as fixing a makefile error), so you can re work the `api/openapi.yaml` definitions in a way that is compatible with the backend.

Please, keep in mind when working with these openapi files, to always run `make swagger`, `make server-generate` and `make server-build` to keep everything in sync

xoscar:
Findings:
1. The main package https://github.com/OpenAPITools/openapi-generator doesn't have a tool to only generate only types, it only supports creating the whole client. 
2. As we are using redux toolkit and RTK query any of the clients supported match what we need.
3. I tried using other tools like https://github.com/drwpow/openapi-typescript but the issue is that the rest of the packages like this only support openapi specs to openappi specs and we are mixing openapi specs and swagger specs 😢  this doesn't allow the proper types to be generated.
4. The trace.yaml swagger file is autogenerated which makes it not viable to be manually updated or changed to openapi.

Conclusion:
After spending some time on this I think it is better to keep continue with the current setup we have and then in the future come back to this issue as it might only slow us down with not much benefit atm. I think the OpenAPI generators are still quite green and not ready for our use case.

PR #410, opened 2022-05-03
use selectors to filter spans in assertions

schoren:
This PR fixes the assertion engine selectors usage. It now can actually filter spans using the `selectors` package

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


mathnogueira (approved):
Awesome!

PR #430, opened 2022-05-05
fix: analytics client

mathnogueira:
This PR fixes insecure TLS connection

## Changes

-

## Fixes

- Analytics names (google doesn't accept dashes in the event name)

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/analytics/analytics.go
+++ b/server/analytics/analytics.go
@@ -2,7 +2,6 @@ package analytics
 
 import (
 	"bytes"
-	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


xoscar (changes requested):
Hey @mathnogueira, I had a talk with Ken and he wanted us to use dashes instead of underscores for the analytics events, is that something that is causing issues?

xoscar (approved):

kdhamric:
I think we shifted to dashes instead of underscores in the event names being passed to GA so we could distinguish our events from GA automatic events.

PR #438, opened 2022-05-05
run go tests on PR

mathnogueira:
Now instead of relying on another piece of software to start the database (docker-compose, github action service, etc). When you run the tests, we use gnomock to start the postgres container for us. Each test runs using a different database running on the same postgres instance.

This also removes the `build and test` phase from PRs. This way we don't deploy our changes when they are in review, just when they get merged into main.

Last but not least important, this enforces all backend tests to pass before you can merge your PR.

## Changes

- tests start a postgres container automatically
- PR  checks if all backend tests before allowing it to be merged

## Fixes

- don't deploy code that is under review

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test


server/go.mod
```diff
@@ -5,20 +5,19 @@ go 1.13
 replace k8s.io/client-go => k8s.io/client-go v0.18.0
 
 require (
```
mathnogueira, line 7: I ran `go mod tidy`

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Install Go
+        if: success()
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Run tests
+        run: cd server && go test ./...
```
schoren, line 26: Can we use the makefile rules here? `make server-test`. This way we can always have the commands used locally and in CI in sync

schoren (approved):
This is awesome. Great work!

PR #444, opened 2022-05-05
fix validation of result update

schoren:
This PR adds a validation in the backend so that the test result state can only be updated by the frontend if the state is `awaiting test results`. That is the only case where the FE should update the test state.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira (approved):
LGTM

PR #450, opened 2022-05-06
feat: run assertions on backend

mathnogueira:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/traces/otel_converter.go
+++ b/server/traces/otel_converter.go
@@ -1,15 +1,14 @@
-package assertions
+package traces
 
 import (
 	"fmt"
 	"math"
 
 	"github.com/kubeshop/tracetest/openapi"
-	"github.com/kubeshop/tracetest/traces"
 	"go.opentelemetry.io/otel/trace"
 )
 
-func convertOTelTraceIntoTraceTree(trace openapi.ApiV3SpansResponseChunk) (traces.Trace, error) {
+func FromOtel(trace openapi.ApiV3SpansResponseChunk) (Trace, error) {
 	flattenSpans := make([]openapi.V1Span, 0)
 	for _, resource := range trace.ResourceSpans {
 		for _, librarySpans := range resource.InstrumentationLibrarySpans {
@@ -19,11 +18,11 @@ func convertOTelTraceIntoTraceTree(trace openapi.ApiV3SpansResponseChunk) (trace
 		}
 	}
 
-	spansMap := make(map[string]*traces.Span, 0)
+	spansMap := make(map[string]*Span, 0)
 	for _, span := range flattenSpans {
 		newSpan, err := convertOtelSpanIntoSpan(span)
 		if err != nil {
-			return traces.Trace{}, err
+			return Trace{}, err
 		}
 		spansMap[span.SpanId] = newSpan
 	}
@@ -31,8 +30,8 @@ func convertOTelTraceIntoTraceTree(trace openapi.ApiV3SpansResponseChunk) (trace
 	return createTrace(flattenSpans, spansMap), nil
 }
 
-func convertOtelSpanIntoSpan(span openapi.V1Span) (*traces.Span, error) {
-	attributes := make(traces.Attributes, 0)
+func convertOtelSpanIntoSpan(span openapi.V1Span) (*Span, error) {
+	attributes := make(Attributes, 0)
 	for _, attribute := range span.Attributes {
 		attributes[attribute.Key] = getAttributeValue(attribute.Value)
 	}
@@ -42,11 +41,11 @@ func convertOtelSpanIntoSpan(span openapi.V1Span) (*traces.Span, error) {
 		return nil, err
 	}
 
-	return &traces.Span{
+	return &Span{
 		ID:         spanID,
 		Name:       span.Name,
 		Parent:     nil,
-		Children:   make([]*traces.Span, 0),
+		Children:   make([]*Span, 0),
 		Attributes: attributes,
 	}, nil
 }
@@ -81,7 +80,7 @@ func createSpanID(id string) (trace.SpanID, error) {
 	return spanId, nil
 }
 
-func createTrace(spans []openapi.V1Span, spansMap map[string]*traces.Span) traces.Trace {
+func createTrace(spans []openapi.V1Span, spansMap map[string]*Span) Trace {
 	var rootSpanID string = ""
 	for _, span := range spans {
 		if span.ParentSpanId == "" {
@@ -97,7 +96,7 @@ func createTrace(spans []openapi.V1Span, spansMap map[string]*traces.Span) trace
 
 	rootSpan := spansMap[rootSpanID]
 
-	return traces.Trace{
+	return Trace{
 		RootSpan: *rootSpan,
 	}
 }
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -358,6 +358,8 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/SelectorItem"
+        selector:
+          type: string
         spanAssertions:
           type: array
           items:
--- a/server/assertions/assertion_converter.go
+++ b/server/assertions/assertion_converter.go
@@ -13,6 +13,7 @@ func convertAssertionsIntoTestDefinition(assertions []openapi.Assertion) TestDef
 		newAssertions := make([]Assertion, 0, len(assertions))
 		for _, spanAssertion := range assertion.SpanAssertions {
 			newAssertion := Assertion{
+				ID:         assertion.AssertionId,
 				Attribute:  spanAssertion.PropertyName,
 				Comparator: getComparator(spanAssertion.Operator),
 				Value:      spanAssertion.ComparisonValue,
--- a/server/assertions/assertion_converter.go
+++ b/server/assertions/assertion_converter.go
@@ -1,0 +1,42 @@
+package assertions
+
+import (
+	"github.com/kubeshop/tracetest/assertions/comparator"
+	"github.com/kubeshop/tracetest/openapi"
+)
+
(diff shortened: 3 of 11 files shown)


server/executor/runner.go
```diff
@@ -65,6 +68,10 @@ func (r persistentRunner) handleDBError(err error) {
 }
 
 func (r persistentRunner) Start(workers int) {
+	r.tp.OnPollComplete(func(result openapi.TestRunResult) {
```
xoscar, line 71: 🚀  This means that on every trace poll end we are going to `try` to run the assertions right? in case the assertion contains the `Selector` entry

server/executor/trace_poller.go
```diff
@@ -201,3 +206,10 @@ func (tp tracePoller) requeue(job tracePollReq) {
 		tp.enqueueJob(job)
 	}()
 }
+
+func (tp tracePoller) OnPollComplete(callback func(openapi.TestRunResult)) {
+	go func() {
```
xoscar, line 211: I guess this is some async programming going on here right?
1. On execution we create a new goroutine.
2. Within the goroutine we wait until the `result` is assigned by the `completePoolChannel` channel.
3. We run the callback with the assigned result value.

Is that right? 😬

server/executor/trace_poller.go
```diff
@@ -150,6 +153,8 @@ func (tp tracePoller) processJob(job tracePollReq) {
 		Type:    "result_update",
 		Content: res,
 	})
+
+	tp.completePoolChannel <- res
```
xoscar, line 157: res (on poll complete) -> completePoolChannel -> result -> callback(result), right?

server/traces/otel_converter.go
```diff
@@ -0,0 +1,102 @@
...
+	"math"
+
+	"github.com/kubeshop/tracetest/openapi"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func FromOtel(trace openapi.ApiV3SpansResponseChunk) (Trace, error) {
+	flattenSpans := make([]openapi.V1Span, 0)
+	for _, resource := range trace.ResourceSpans {
+		for _, librarySpans := range resource.InstrumentationLibrarySpans {
```
xoscar, line 14: Remember that there are some attributes at the resource level as well. Things like `service.name` are there.

server/traces/otel_converter.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if value.DoubleValue != 0.0 {
+		isFloatingPoint := math.Abs(value.DoubleValue-math.Abs(value.DoubleValue)) > 0.0
+		if isFloatingPoint {
+			return fmt.Sprintf("%f", value.DoubleValue)
+		}
+
+		return fmt.Sprintf("%.0f", value.DoubleValue)
+	}
```
xoscar, line 69: Remember to cover the `kvlistValue` and `arrayValue` 😬

server/traces/otel_converter.go
```diff
@@ -0,0 +1,102 @@
...
+	spanId, err := trace.SpanIDFromHex(id)
+	if err != nil {
+		return trace.SpanID{}, fmt.Errorf("could not convert spanID")
+	}
+
+	return spanId, nil
+}
+
+func createTrace(spans []openapi.V1Span, spansMap map[string]*Span) Trace {
+	var rootSpanID string = ""
```
xoscar, line 84: 🌳

xoscar (commented):
Awesome job @mathnogueira!. I left some comments and concerns, let me know what you think 😄

server/executor/assertion_converter.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+
+			newAssertions = append(newAssertions, newAssertion)
+		}
+		testDefinition[assertions.SpanQuery(selector)] = newAssertions
+	}
+	return testDefinition
+}
+
+func getComparator(operatorName string) comparator.Comparator {
```
schoren, line 30: This should be done using a comparator registry. Like:
```
reg := comparators.DefaultRegistry() // this returns a new instance each time, so should be injected
reg.Get(">") // returns `comparator.Gt`
```

server/executor/assertion_runner.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Temporary patch to disable the assertion engine if frontend request is not prepared yet (old selector format)
+	if e.shouldIgnoreTest(test) {
+		return &testResult, nil
+	}
+
+	testDefinition := convertAssertionsIntoTestDefinition(test.Assertions)
```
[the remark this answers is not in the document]
schoren: Yes, I think we should convert everything before it enters the queue. Is that possible?

server/executor/assertion_runner.go
```diff
@@ -0,0 +1,153 @@
...
+	return false
+}
+
+func (e *defaultAssertionRunner) setResults(result *openapi.TestRunResult, testResult assertions.TestResult) {
+	result.State = TestRunStateFinished
+	result.CompletedAt = time.Now()
+	assertionResultArray := make([]openapi.AssertionResult, 0)
+	allTestsPassed := true
+
+	for _, assertionExecutionResult := range testResult {
```
schoren, line 121: same here. we should work with the `assertions` entities, and leave the convertion to `openapi` for the controller.

server/executor/runner.go
schoren: I don't think of the runner as an orchestrator. The whole process is a chain of queues. I would move this dependency to the `tracePoller`, so when that queue determines the trace is `polled`, or ready to be asserted, it can just pass it to the assertion runner, much in the same way this queue sends the work to the next queue

schoren (changes requested):

server/executor/assertion_runner.go
mathnogueira: TBH, I'm not sure that would be possible to do here. This is being executed at the end of the assertion execution that happens on the worker side. The controller cannot be executed here.

server/executor/integration_test.go
```diff
@@ -24,25 +24,87 @@ func TestExecutorIntegration(t *testing.T) {
...
 
 	time.Sleep(1 * time.Second)
 
+	testCases := []struct {
+		Name string
+		Fn   func(*testing.T, *app.App, *test.DemoApp)
+	}{
+		{
+			Name: "Test with new selector should be asserted by assertion engine",
+			Fn:   happyPathWithNewSelector,
```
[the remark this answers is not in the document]
mathnogueira: I thought about that, but that would increase the complexity of the setup because I would have to ensure only one stack is being called when `setup()` is called twice, otherwise, this means we will have to start everything twice and that would add a 30s overhead on the tests.

server/traces/otel_converter.go
mathnogueira: Good call. I think only the span name is at the resource top level. I'm almost sure service.name is inside the `attributes` array.

server/traces/otel_converter.go
mathnogueira: Supporting those mean that we have to rethink a couple of things. I'll rather do that in a different PR

server/executor/assertion_converter.go
```diff
@@ -0,0 +1,57 @@
...
+
+			newAssertions = append(newAssertions, newAssertion)
+		}
+		testDefinition[assertions.SpanQuery(selector)] = newAssertions
+	}
+	return testDefinition, nil
+}
+
+func getComparator(operatorName string) (comparator.Comparator, error) {
+	operationName := "="
```
schoren, line 37: The idea was to remove the switch. If you need to use the words instead of the simbols, feel free to change the comparators to use that instead.

server/executor/assertion_runner.go
schoren: Yes, I wasn't sure if this was doable

server/executor/integration_test.go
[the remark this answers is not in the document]
schoren: In that case, I would use direct function call instead of a loop simulating a table test.

schoren (changes requested):

server/executor/assertion_converter.go
schoren: As discussed offline, it makes sense to keep this switch now to keep compatibility with both the current FE implementation and the newer selectors format. We'll remove this later when we better integrate the FE with the new selectors

server/executor/runner_test.go
```diff
@@ -126,6 +127,10 @@ func runnerSetup(t *testing.T) runnerFixture {
 
 	mtp := new(mockTracePoller)
 	mtp.t = t
+
+	mar := new(mockAssertionRunner)
```
schoren, line 131: this can be removed

server/executor/runner_test.go
```diff
@@ -259,11 +264,40 @@ type mockTracePoller struct {
...
+func (m *mockTracePoller) OnPollComplete(callback func(result openapi.TestRunResult)) {
+	m.Called(callback)
+}
+
 func (m *mockTracePoller) expectPoll(test openapi.Test) *mock.Call {
 	return m.
 		On("Poll", test.TestId)
 }
+
+func (m *mockTracePoller) expectOnPollComplete() *mock.Call {
```
schoren, line 282: This can be removed

server/executor/runner_test.go
```diff
@@ -259,11 +264,40 @@ type mockTracePoller struct {
...
+	return m.On("OnPollComplete", mock.Anything)
+}
+
+type mockAssertionRunner struct {
```
schoren, line 286: THis can be moved to `trace_poller_test.go`

schoren (approved):

PR #457, opened 2022-05-09
💣  453 adding integration tests for the trace view

xoscar:
This PR adds unit tests for the trace view flow.

## Changes

- Adding `data-cy` tags for the trace view section.

## Fixes

- Re-adds the test results tab that was removed

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/cypress/integration/Trace/CreateAssertion.spec.ts
+++ b/web/cypress/integration/Trace/CreateAssertion.spec.ts
@@ -1,4 +1,4 @@
-import { createTest, deleteTest, testId } from "../utils/common";
+import {createTest, deleteTest, testId} from '../utils/common';
 
 describe('Create Assertion', () => {
   before(() => {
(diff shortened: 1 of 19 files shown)


web/cypress/integration/Trace/CreateAssertion.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it('should create a basic assertion', () => {
+    cy.get(`[data-cy=test-url-${testId}]`).first().click();
+    cy.location('href').should('match', /\/test\/.*/i);
+
+    cy.get(`[data-cy^=test-run-result-]`).first().click();
+    cy.location('href').should('match', /resultId=.*/i);
+
+    cy.wait(5000);
```
mathnogueira, line 19: Can't you replace this with something like:

```js
cy.get('[data-cy=add-assertion-button]', { timeout: 10000 }).should('be.visible');
```

PS: I didn't test it, but I'm almost sure that would work.

web/cypress/integration/Trace/CreateAssertion.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    cy.location('href').should('match', /\/test\/.*/i);
+
+    cy.get(`[data-cy^=test-run-result-]`).first().click();
+    cy.location('href').should('match', /resultId=.*/i);
+
+    cy.get('[data-cy=add-assertion-button]').click();
+    cy.get('[data-cy=create-assertion-form]', {timeout: 10000}).should('be.visible');
+
+    cy.get('[data-cy=item-selector-tag] + [role=img]').first().click();
+    cy.get('[data-cy=affected-spans-count]').should('have.text', 'Affects 7 spans');
```
mathnogueira, line 23: I think we should not set the expected text as `Affected 7 spans` because that might make the test brittle. What if we change our demo and now we have 8 spans? That would break the FE test.

I think it would be better to have something like:

```js
cy.get('[data-cy=affected-spans-count]').should('match', /Affects \d+ spans/);
```

kdhamric (approved):

PR #466, opened 2022-05-10
Fix ga settings

schoren:
This PR moves the GA settings parsing to the config package, and removes all `os.getenv` calls from the `analytics` package. It also fixes the gh action so it keeps analytics enabled every deploy.

Depends on https://github.com/kubeshop/helm-charts/pull/157

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

### DIFF
--- a/server/analytics/analytics.go
+++ b/server/analytics/analytics.go
@@ -8,139 +8,132 @@ import (
 	"net/http"
 	"os"
 	"runtime"
-	"strconv"
 
 	"github.com/denisbrodbeck/machineid"
+	"github.com/kubeshop/tracetest/config"
 )
 
-var (
-	appName          = "tracetest"
-	analyticsEnabled = false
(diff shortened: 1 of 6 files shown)


server/analytics/analytics.go
```diff
@@ -8,139 +8,132 @@ import (
...
+		return nil
+	}
+
+	// setup an actual client
+	hostname, err := os.Hostname()
 	if err != nil {
-		analyticsEnabled = false
-	} else {
-		analyticsEnabled = analyticsIsEnabled
+		return fmt.Errorf("could not get hostnmae: %w", err)
```
mathnogueira, line 33:
```suggestion
		return fmt.Errorf("could not get hostname: %w", err)
```

mathnogueira (approved):
LGTM. Just apply the change I added to fix a typo

PR #476, opened 2022-05-11
add basic test suites for hooks folder

cescoferraro:
This PR creates basic test suites for the hooks placed at the hooks folder

## Changes

- define typing for existing hooks
- add tests

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -39,7 +39,6 @@
   "scripts": {
     "start": "craco start",
     "build": "craco build",
-    "test:work": "craco test --watchAll=false --coverage ./src/hooks --collectCoverageFrom='src/hooks/**/*.(tsx|ts)'",
     "test:unit": "craco test --watchAll=false --coverage",
     "test:integration": "npm run cy:run",
     "test:lint": "eslint ./src/**/*.{ts,tsx}",
@@ -79,8 +78,7 @@
   "jest": {
     "transformIgnorePatterns": [
       "/node_modules/(?!antd|@ant-design|rc-.+?|@babel/runtime).+(js|jsx)$"
-    ],
-    "testMatch": [ "<rootDir>/src/hooks/**/*.test.tsx" ]
+    ]
   },
   "resolutions": {
     "styled-components": "^5"
--- a/web/src/hooks/useDAGChart.ts
+++ b/web/src/hooks/useDAGChart.ts
@@ -3,12 +3,12 @@ import {Dag} from 'd3-dag';
 import _ from 'lodash';
 import {TSpanMap} from '../components/Diagram/components/DAG';
 
-export const useDAGChart = (
-  spanMap: TSpanMap = {}
-): void | {
+interface DAGResponse {
   dag: Dag<{id: string; parentIds: string[]}, undefined>;
   layout: {width: number; height: number};
-} => {
+}
+
+export const useDAGChart = (spanMap: TSpanMap = {}): void | DAGResponse => {
   if (_.isEmpty(spanMap)) {
     return;
   }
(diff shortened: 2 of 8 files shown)


web/package.json
```diff
@@ -39,7 +39,8 @@
   "scripts": {
     "start": "craco start",
     "build": "craco build",
-    "test:unit": "craco test --watchAll=false",
+    "test:work": "craco test --watchAll=false --coverage ./src/hooks --collectCoverageFrom='src/hooks/**/*.(tsx|ts)'",
```
xoscar, line 42: Did this leak into the commit?

web/package.json
```diff
@@ -78,7 +79,8 @@
   "jest": {
     "transformIgnorePatterns": [
       "/node_modules/(?!antd|@ant-design|rc-.+?|@babel/runtime).+(js|jsx)$"
-    ]
+    ],
+    "testMatch": [ "<rootDir>/src/hooks/**/*.test.tsx" ]
```
xoscar, line 83: Shouldn't this match everything not only hooks?

web/src/hooks/useDAGChart.ts
```diff
@@ -1,8 +1,14 @@
...
+import {Dag} from 'd3-dag';
 import _ from 'lodash';
-import { TSpanMap } from '../components/Diagram/components/DAG';
+import {TSpanMap} from '../components/Diagram/components/DAG';
 
-export const useDAGChart = (spanMap: TSpanMap = {}) => {
+export const useDAGChart = (
+  spanMap: TSpanMap = {}
+): void | {
+  dag: Dag<{id: string; parentIds: string[]}, undefined>;
```
xoscar, line 9: Should we encapsulate this into an interface?

web/src/hooks/usePolling.ts
```diff
@@ -6,18 +6,26 @@ interface usePollingProps {
   isPolling: boolean;
 }
 
-const usePolling = ({callback, delay = 1000, isPolling}: usePollingProps) => {
+const usePolling = ({callback, delay = 1000, isPolling}: usePollingProps): void => {
   useEffect(() => {
     let interval: any = null;
 
+    console.log(1);
```
xoscar, line 13: I guess these logs aren't supposed to be here right?

web/src/hooks/usePolling.test.tsx
```diff
@@ -0,0 +1,10 @@
+import {renderHook} from '@testing-library/react-hooks';
+import usePolling from './usePolling';
+
+test('usePolling', async () => {
+  const callback = jest.fn();
+  const payload = {callback, isPolling: true, delay: 0};
+  await renderHook(() => usePolling(payload));
+
+  expect(callback).toBeCalledTimes(0);
```
xoscar, line 9: Shouldn't the tests be more extensive to cover the different scenarios, like when the `isPolling` could change, etc?

xoscar (changes requested):
Hey @cescoferraro I left some comments. Let me know what you think.

A separate question: Could we move the test files under a `__tests__` folder inside of the same `hooks` one? Thanks!

web/src/hooks/usePolling.test.tsx
cescoferraro: i am focusing on getting test suites present. 
testing setInterval is a bit tricky.
I am aiming on getting coverage up on those 3 folders.

web/src/hooks/useDAGChart.ts
cescoferraro: sure

web/package.json
cescoferraro: my bad

web/package.json
cescoferraro: my bad

xoscar (approved):
Looking good! 🚀

PR #489, opened 2022-05-13
feat: endpoint to rerun test assertions on result

mathnogueira:
This PR adds an endpoint `/tests/:testID/results/:resultID/rerun` to rerun the assertion engine on a result.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/assertion_runner.go
```diff
@@ -160,6 +160,23 @@ func (e *defaultAssertionRunner) setResults(result *openapi.TestRunResult, testR
 	result.AssertionResultState = allTestsPassed
 }
 
-func (e *defaultAssertionRunner) RunAssertions(request AssertionRequest) {
-	e.inputChannel <- request
+func (e *defaultAssertionRunner) RunAssertions(ctx context.Context, result openapi.TestRunResult) error {
```
schoren, line 163: This is adding an unnecesary responsibility to this function. I think the controller can easily do this setup and keep the assertion runner just running assertions

schoren (changes requested):

schoren (approved):
Awesome!

PR #490, opened 2022-05-13
Clean traces api format

schoren:
This PR removes the otel protobuf format mappings from the openapi spec, and proposes a new structure for tests, assertions, etc. This change has a really big impact both in the back and frontend, since it refactors most of the key models.

This is not functional, nor does it compile, but should be enough to discuss if the format is good enough.

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test


mathnogueira (commented):
For me, those changes make sense.

api/tests.yaml
```diff
@@ -0,0 +1,161 @@
...
+        format: uuid
+        readOnly: true
+      testId:
+        type: string
+        format: uuid
+        readOnly: true
+      traceId:
+        type: string
+        readOnly: true
+      spanId:
```
kdhamric, line 79: What is spanId in reference to a TestRun?

schoren: both TraceID and SpanID are generated by us and sent in the triggering transaction

mathnogueira (approved):
integration test passing is a big green light for this change :D

I reviewed everything and I don't have anything against any change, so I'll approve it.

xoscar (approved):
Great job @schoren! 🔥

PR #501, opened 2022-05-19
add newSelectorLogic method

cescoferraro:
This PR create a method to transform `selectorList: IItemSelector[]` to a selector string described here

## Changes

- nothing, just adds a new method

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


web/src/components/CreateAssertionModal/newSelectorLogic.tsx
```diff
@@ -0,0 +1,20 @@
+import {IItemSelector} from '../../types/Assertion.types';
+
+export function newSelectorLogic(selectorList: IItemSelector[]): string {
```
xoscar, line 3: Let's move this to the Assertion service

web/src/components/CreateAssertionModal/newSelectorLogic.tsx
```diff
@@ -0,0 +1,20 @@
...
+  let result = '';
+  for (let i = 0; i < selectorList.length; i += 1) {
+    const selector = selectorList[i];
+    result += `span[`;
+    result += selector.propertyName;
+    result += selector.operator ? ` ${selector.operator.toLowerCase()} ` : '=';
+    // add quotes if value is a string
+    if (selector.valueType === 'stringValue') result += `'`;
```
xoscar, line 11: I think we should have a more complex way of creating this chunk to support the other value types.

web/src/components/CreateAssertionModal/newSelectorLogic.tsx
xoscar: I think we can change this to be more verbose by doing something like:

```
const getFilters = (selectorList) => {
  return selectorList.map(({propertyName, operator, value, valueType}) => `${propertyName} ${operator || '='} ${getValue(value, valueType)}`);
}

const result `span[${getFilters(selectorList).join(' ')}]`;
```

xoscar (changes requested):
Hey @cescoferraro I left some comments and concerns. Let me know what you think

xoscar (approved):

PR #516, opened 2022-05-19
feat: cli tests pipeline

mathnogueira:
This PR adds a workflow to run tests on the CLI module when something there changes. It also skips the backend checks if nothing has changed.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


.github/workflows/run-server-tests.yml
```diff
@@ -5,8 +5,10 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - server/**
```
mathnogueira, line 9: Now the server tests only run when something inside the directory `server` changes. Otherwise this check will be skipped.

cli/Makefile
```diff
@@ -0,0 +1,2 @@
+test:
```
mathnogueira, line 1: I created a new makefile instead of using the root one because the CLI probably gonna be extracted to an individual repo later. So I want to keep it as isolated as possible.

schoren (approved):
LGTM!

Issue #449, opened 2022-05-06
[Testing] Adding pre-commit hooks that run tests and lint


xoscar:
As a developer:

Before I commit something to the web repo I should see the integration tests, unit tests, and lint rules executed and if those fail the commit should not proceed.

kdhamric:
@xoscar Can we close this?

PR #519, opened 2022-05-20
#449 Pre-commit hooks

cescoferraro:
This PR add pre-commit hooks, but it requires an extra `npm i` in the root of the project.
This IMHO will create issues because it will require backend guys to have a Node.js environment setup.

## Changes

- It add git-hooks

## Fixes

- #449 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


xoscar:
Let's bring this up with the rest of the team, to see if they would have any issue if we do this

PR #528, opened 2022-05-23
feat: add version to Test and TestRun API definitions

mathnogueira:
This PR adds versioning to both tests and references the test in the `test run` as well. So now we know which test definition was used to create a specific run.

## Changes

- add `version` field to test
- add `testVersion` field to run

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/api/tests.yaml
+++ b/api/tests.yaml
@@ -12,6 +12,9 @@ components:
           type: string
         description:
           type: string
+        version:
+          type: string
+          description: version number of the test
         serviceUnderTest:
           type: object
           properties:
@@ -71,6 +74,9 @@ components:
         spanId:
           type: string
           readOnly: true
+        testVersion:
+          type: string
+          description: Test version used when running this test run
         state:
           type: string
           enum: [CREATED, EXECUTING, AWAITING_TRACE, AWAITING_TEST_RESULTS, FINISHED, FAILED]
--- a/api/tests.yaml
+++ b/api/tests.yaml
@@ -13,7 +13,7 @@ components:
         description:
           type: string
         version:
-          type: string
+          type: integer
           description: version number of the test
         serviceUnderTest:
           type: object
@@ -75,7 +75,7 @@ components:
           type: string
           readOnly: true
         testVersion:
-          type: string
+          type: integer
           description: Test version used when running this test run
         state:
           type: string
--- a/server/http/controller.go
+++ b/server/http/controller.go
@@ -290,11 +290,11 @@ func (c *controller) UpdateTest(ctx context.Context, testID string, in openapi.T
 
 	if testHasChanged {
 		updated.Version = test.Version + 1
-	}
 
-	err = c.testDB.UpdateTest(ctx, updated)
-	if err != nil {
-		return handleDBError(err), err
+		_, err = c.testDB.CreateTestVersion(ctx, updated)
+		if err != nil {
+			return handleDBError(err), err
+		}
 	}
 
 	return openapi.Response(204, nil), nil
--- a/server/http/controller.go
+++ b/server/http/controller.go
@@ -283,16 +283,9 @@ func (c *controller) UpdateTest(ctx context.Context, testID string, in openapi.T
 	updated.ID = test.ID
 	updated.ReferenceRun = nil
 
-	updated, err = model.BumpTestVersionIfNeeded(test, updated)
+	_, err = c.testDB.UpdateTest(ctx, updated)
 	if err != nil {
-		return openapi.Response(http.StatusUnprocessableEntity, err.Error()), err
-	}
-
-	if updated.Version != test.Version {
-		_, err = c.testDB.CreateTestVersion(ctx, updated)
-		if err != nil {
-			return handleDBError(err), err
-		}
+		return handleDBError(err), err
 	}
 
 	return openapi.Response(204, nil), nil
--- a/server/http/controller.go
+++ b/server/http/controller.go
@@ -3,7 +3,6 @@ package http
 import (
 	"context"
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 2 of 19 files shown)


api/tests.yaml
```diff
@@ -12,6 +12,9 @@ components:
           type: string
         description:
           type: string
+        version:
+          type: integer
+          description: version number of the test
```
schoren, line 17: what do you think? 
```suggestion
          description: current version number of the test
```

server/http/controller.go
```diff
@@ -280,10 +281,50 @@ func (c *controller) UpdateTest(ctx context.Context, testID string, in openapi.T
 	updated.ID = test.ID
 	updated.ReferenceRun = nil
 
+	testHasChanged, err := c.testHasChanged(test, updated)
```
schoren, line 284: all this logic should live outside the controller. I'm thinking we could have it in `server/model/test_changes.go` and have somehting like:

```go
func ChangeTest(in, updated Test) (Test, error) {
  testHasChanged, err := testHasChanged(test, updated)
  if err != nil {
    return Test{}, err
  }

  if testHasChanged {
    updated.Version = test.Version + 1
  }

  return updated
}
```

and move all the `testHasChanged` etc methods there too

server/http/controller.go
```diff
@@ -280,10 +281,50 @@ func (c *controller) UpdateTest(ctx context.Context, testID string, in openapi.T
...
+	if err != nil {
+		return false, err
+	}
+
+	newFieldJSON, err := json.Marshal(newField)
+	if err != nil {
+		return false, err
+	}
+
+	return string(oldFieldJSON) != string(newFieldJSON), nil
```
schoren, line 329: nice approach, smart and simple

server/migrations/4_test_version.up.sql
```diff
@@ -0,0 +1,15 @@
+ALTER TABLE tests
+ADD COLUMN "version" int NOT NULL DEFAULT 1;
+
+ALTER TABLE tests
+DROP CONSTRAINT tests_pkey;
+
+ALTER TABLE tests
+ADD CONSTRAINT tests_pkey PRIMARY KEY (id, "version");
```
schoren, line 8: 👍

server/testdb/tests_test.go
```diff
@@ -81,3 +81,30 @@ func TestGetTests(t *testing.T) {
...
 	assert.Len(t, actual, 0)
 }
+
+func TestGetTestsWithMultipleVersions(t *testing.T) {
+	db, clean := getDB()
+	defer clean()
+
+	test1 := createTestWithName(t, db, "1")
+	test1.Name = "1 v2"
+	test1.Version = 2
```
schoren, line 91: `db.UpdateTest` should automatically increase the version, so we don't forget to do it manually in every place we update a test

schoren (changes requested):

server/testdb/tests_test.go
mathnogueira: The thing is that we use `UpdateTest` internally. So I decided to introduce another method: `CreateTestVersion` that creates a new Test but doesn't prepopulate its fields like `ID`, just like `CreateTest` does.

server/http/controller.go
mathnogueira: Great idea. I just pushed the change

server/testdb/tests_test.go
schoren: Still, [this part](https://github.com/kubeshop/tracetest/pull/528/files#diff-cdf13307c512afb9b589e95ccb04f91bc209833689b1563b5f363a9f7513e0feR289) needs to be encapsulated.

What if we had a `ForceUpdateTest` that just updates the test leaving the version untouched, and the `UpdateTest` that creates a new version?

mathnogueira: I think we can do that, I just don't like the `ForceUpdateTest` because it gives the impression that we have some rule that doesn't allow the test to be updated, and using that, we skip that rule.

What if we have `UpdateTest` and `UpdateTestVersion`? 
`UpdateTest` updates the test and creates a new version. `UpdateTestVersion` updates the details of a test version without creating a new one.

schoren: yea, that looks much better. I like it

mathnogueira: Just pushed the change

mathnogueira: Forgot to remove the `test.Version = 2`. I'm gonna push this change now.

server/testdb/tests_test.go
```diff
@@ -60,12 +60,28 @@ func TestDeleteTest(t *testing.T) {
...
 
 }
 
+func TestGetLatestTestVersion(t *testing.T) {
+	db, clean := getDB()
+	defer clean()
+
+	test := createTestWithName(t, db, "1")
+	test.Name = "1 v2"
+	test.Version = 2
```
schoren, line 75: This line can go now

server/testdb/tests_test.go
```diff
@@ -60,12 +60,28 @@ func TestDeleteTest(t *testing.T) {
...
+
+	_, err := db.UpdateTest(context.TODO(), test)
+	require.NoError(t, err)
+
+	latestTest, err := db.GetLatestTestVersion(context.TODO(), test.ID)
+	assert.NoError(t, err)
+	assert.Equal(t, "1 v2", latestTest.Name)
```
schoren, line 82: And you could add a `assert.Equal(t, 2, latestTest.Version)` too

server/testdb/tests_test.go
```diff
@@ -81,3 +97,30 @@ func TestGetTests(t *testing.T) {
...
+	test1 := createTestWithName(t, db, "1")
+	test1.Name = "1 v2"
+	test1.Version = 2
+
+	_, err := db.UpdateTest(context.TODO(), test1)
+	require.NoError(t, err)
+
+	test2 := createTestWithName(t, db, "2")
+	test2.Name = "2 v2"
+	test2.Version = 2
```
schoren: this can go

mathnogueira: Just removed it

schoren (approved):
LGTM! This is a really nice implementation!

PR #529, opened 2022-05-23
fix: use old proto image if amd64 arch

mathnogueira:
This PR enables people using AMD64 architecture to generate our protobuf code.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,8 +36,13 @@ init-submodule:
 	git submodule update
 
 PROTOC_VER=0.3.1
-# temporarily using custom image to make it work on m1 macs
-PROTOC_IMAGE=schoren/protobuf:$(PROTOC_VER)
+UNAME_P := $(shell uname -p)
+ifeq ($(UNAME_P),x86_64)
+	PROTOC_IMAGE=jaegertracing/protobuf:$(PROTOC_VER)
+endif
+ifneq ($(filter arm%,$(UNAME_P)),)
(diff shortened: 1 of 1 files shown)


mathnogueira:
@schoren can you check if that breaks your M1 environment?

schoren:
@mathnogueira I don't have internet at home until friday, so I'm using my laptop, which isn't m1. I can test it then. Why do you need to do this change?

mathnogueira:
> @mathnogueira I don't have internet at home until friday, so I'm using my laptop, which isn't m1. I can test it then. Why do you need to do this change?

No problem. We can approve that later. I added this PR because I'm not using an M1 laptop, so when I try to run: `make`, I get this error:

```
WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
standard_init_linux.go:228: exec user process caused: exec format error
make: *** [Makefile:77: proto] Error 1
```

Applying the changes of this PR, it doesn't fail.

schoren:
Mhh, I think I can make the docker image cross platform. Let me try to fix it that way

mathnogueira:
This fix works on the M1 chip as well. Thanks @cescoferraro 

image

schoren (approved):

Issue #513, opened 2022-05-19
Simple issues from implementation of new UX


kdhamric:
These are items we want to change in the new UI that do not require design / @olha23 input:

- We want to be able to click on 'Test Status: Finished' and have the bottom test area expand (we watch somebody do this... might as well help get them to the test if they do not see it at the bottom)
- Click an X to delete one filter does delete the filter... but it puts you in 'add a new filter' mode and prompts you for an attibute.  I think you should have to click in the filed again to start the 'add filter' process.
- When editing an assertion, the 'Add' button should say 'Save'
- If you add a second check line (ie add check), there is no way to delete that line.  Should have a trash can next to it.
- Probably should allow you to hit 'add check', not enter anything, then hit 'Add' or 'Save'
- 'Cancel' when adding an assertion closes the test area at the bottom - it should not.

cescoferraro:
- We want to be able to click on 'Test Status: Finished' and have the bottom test area expand (we watch somebody do this... might as well help get them to the test if they do not see it at the bottom)

- Click an X to delete one filter does delete the filter... but it puts you in 'add a new filter' mode and prompts you for an attibute. I think you should have to click in the filed again to start the 'add filter' process.

- When editing an assertion, the 'Add' button should say 'Save' ✅ 

- If you add a second check line (ie add check), there is no way to delete that line. Should have a trash can next to it. ✅ 

- Probably should allow you to hit 'add check', not enter anything, then hit 'Add' or 'Save'

- 'Cancel' when adding an assertion closes the test area at the bottom - it should not.

kdhamric:
on:
Click an X to delete one filter does delete the filter... but it puts you in 'add a new filter' mode and prompts you for an attibute. I think you should have to click in the filed again to start the 'add filter' process.
(Could not make this work ⛑️, we probably need a specific ticket for this. As it is not trivial to implement)

Communicated with you - you are going to create the issue.

cescoferraro:
Here it is

https://github.com/kubeshop/tracetest/issues/552

PR #551, opened 2022-05-25
513 simple ux issues

cescoferraro:
This PR solves a couple of issues found on #513 

## Changes

- We want to be able to click on 'Test Status: Finished' and have the bottom test area expand (we watch somebody do this... might as well help get them to the test if they do not see it at the bottom) ✅ 

- When editing an assertion, the 'Add' button should say 'Save' ✅ 

- If you add a second check line (ie add check), there is no way to delete that line. Should have a trash can next to it. ✅ 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -41,6 +41,7 @@
   },
   "scripts": {
     "start": "craco start",
+    "port-forward": "kubectl -n tracetest-beta port-forward svc/tracetest-beta 8080:8080",
     "build": "craco build",
     "test": "npm run test:unit",
     "test:integration": "npm run cy:run",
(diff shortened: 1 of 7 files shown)


web/package.json
```diff
@@ -41,6 +41,7 @@
   },
   "scripts": {
     "start": "craco start",
+    "port-forward": "kubectl -n tracetest-beta port-forward svc/tracetest-beta 8080:8080",
```
xoscar, line 44: great addition imo

web/src/components/TestHeader/TestHeader.tsx
```diff
@@ -20,7 +22,7 @@ const TestHeader: React.FC<TTestHeaderProps> = ({test: {name, serviceUnderTest},
         </S.TestUrl>
       </S.Content>
       {testState && (
-        <S.StateContainer data-cy="test-run-result-status">
+        <S.StateContainer onClick={() => setIsCollapsed(o => !o)} data-cy="test-run-result-status">
```
xoscar, line 25: This is a weird request lol

xoscar (approved):
LGTM!

web/package.json
cescoferraro: Thanks

PR #555, opened 2022-05-25
feat: rerun test run creates a new run and copy data from original one

mathnogueira:
This PR creates a new test run when a run is rerun instead of updating the same run. It also uses the latest test version to run the assertions.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


xoscar (approved):
EZ PZ 🍋

PR #556, opened 2022-05-25
add test fixtures to improve test setup

mathnogueira:
This PR introduces a new testing structure called `fixtures` based on Python's pytest.

Fixtures are a way of obtaining a specific value that can be reused by tests. Fixtures also cache their value, so if you call a fixture twice, it will only generate the value once and all subsequent calls will use the cache.

## Why did I do that? 

1. Our integration test was complex
2. I wanted to be able to write more integration tests to prevent our changes from breaking things (versioning did that a couple of times and integration tests would have allowed us to see that during development). However, adding more tests that relied on the pokemon demo, the more our test suite would take to run. Thus, by adding a mechanism that would automatically cache those values and only run things once, we reduce this overhead.

## What can be improved?

One idea is to enable fixtures to turn off the cache mechanism: for example, we might want to have a new `test run` everytime we need one instead of reusing the same one. It could be achieved by calling the `rerun` method to clone the test run and return it. By disabling the cache, we would ensure that all calls to that fixture would return a clone of the run instead of reusing it.

## Side effects
Because integration tests need to be in a separated package, code coverage is not reflecting its real value, but I prefer to have tests ensuring everything works than having a high code coverage metric anyway.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/integration/rerun_test_run_test.go
+++ b/server/integration/rerun_test_run_test.go
@@ -20,7 +20,7 @@ func TestRerun(t *testing.T) {
 	importPokemonTestRun, err := testfixtures.GetPokemonTestRun()
 	require.NoError(t, err)
 
-	newTestRun, err := rerunTestRun(importPokemonTest, importPokemonTestRun)
+	newTestRun := rerunTestRun(t, importPokemonTest, importPokemonTestRun)
 	assert.NoError(t, err)
 
 	assert.NotEqual(t, importPokemonTestRun.Id, newTestRun.Id)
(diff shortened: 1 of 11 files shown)


mathnogueira:
I've added another integration test to test the `rerun` endpoint as a way of testing if the fixture had the right behavior.

I also moved the integration tests to the same package for a reason: `go` runs tests of different packages in different processes (os.Getpid() returns different  pids). This caused problems with the fixture as the `fixtures` map would have different values.

server/integration/rerun_test_run_test.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func rerunTestRun(test *openapi.Test, testRun *openapi.TestRun) (*openapi.TestRun, error) {
+	url := fmt.Sprintf("%s/api/tests/%s/run/%s/rerun", endpointUrl, test.Id, testRun.Id)
+	response, err := http.Post(url, "application/json", nil)
+	if err != nil {
+		return nil, err
+	}
+
+	if response.StatusCode != 200 {
```
schoren, line 39: can we just use `require` here?

schoren (approved):

server/integration/rerun_test_run_test.go
mathnogueira: Changed it.

PR #576, opened 2022-05-26
fix: return test version on run object

mathnogueira:
This PR fetches the test version from the proper location and returns it on all `run` endpoints.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/testdb/runs.go
+++ b/server/testdb/runs.go
@@ -57,7 +57,7 @@ func (td *postgresDB) UpdateRun(ctx context.Context, r model.Run) error {
 }
 
 func (td *postgresDB) GetRun(ctx context.Context, id uuid.UUID) (model.Run, error) {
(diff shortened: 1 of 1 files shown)


server/testdb/runs.go
```diff
@@ -57,7 +57,7 @@ func (td *postgresDB) UpdateRun(ctx context.Context, r model.Run) error {
 }
 
 func (td *postgresDB) GetRun(ctx context.Context, id uuid.UUID) (model.Run, error) {
-	stmt, err := td.db.Prepare("SELECT run FROM runs WHERE id = $1")
+	stmt, err := td.db.Prepare("SELECT run, test_version FROM runs WHERE id = $1")
```
mathnogueira, line 60: test_version is not present in the json, but as a separate column in the table.

schoren (approved):
LGTM!

Issue #562, opened 2022-05-26
When creating or editing an assertion, validate that the selector is unique.


xoscar:
Each test definition should have a unique selector, if the selector exists in the draft list or the published one, throw a validation error

PR #578, opened 2022-05-26
🔀  562 when creating or editing an assertion validate that the selector is unique

xoscar:
This PR blocks the user from adding the same selector twice. 
The behavior is the following:
1. If the user selects a span in which a test definition already exists.
2. The app will concatenate the new set of checks the user might want to add.
3. And the assertion form will be opened in edit mode.

1. If the user selects a span or configures the selection in a way that the result already exists.
2. A FE validation error would be displayed telling the user that the selector already exists blocking him from submitting the form.

## Changes

- New find selector logic.

## Fixes

- https://github.com/kubeshop/tracetest/issues/562

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/src/components/SpanDetail/SpanDetail.tsx
+++ b/web/src/components/SpanDetail/SpanDetail.tsx
@@ -8,7 +8,7 @@ import Http from './components/Http';
 import SpanHeader from './SpanHeader';
 import * as S from './SpanDetail.styled';
 import {useAssertionForm} from '../AssertionForm/AssertionFormProvider';
-import {CompareOperator, PseudoSelector} from '../../constants/Operator.constants';
+import {CompareOperator} from '../../constants/Operator.constants';
 import OperatorService from '../../services/Operator.service';
 
 export interface TSpanDetailProps {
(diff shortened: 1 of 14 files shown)


web/src/components/SpanDetail/SpanDetail.tsx
```diff
@@ -41,29 +41,24 @@ const SpanDetail: React.FC<TSpanDetailProps> = ({span}) => {
 
   const onCreateAssertion = useCallback(
     ({value, key}: TSpanFlatAttribute) => {
+      const {selectorList, pseudoSelector} = SpanService.getSelectorInformation(span!);
```
xoscar, line 44: These are now coming from a service function

web/src/providers/TestDefinition/hooks/useDraftMode.ts
```diff
@@ -8,6 +8,12 @@ const useDraftMode = (isDefaultDraftMode = false) => {
     else window.onbeforeunload = null;
   }, [isDraftMode]);
 
+  useEffect(() => {
+    return () => {
```
xoscar, line 12: On unload we need to remove the block

cescoferraro (approved):
Great work @xoscar

kdhamric (approved):
Watch the video - looks like you implemented it well.  I am still not crazy about what happens towards the end where if a user alters the selector and ends up matching an existing one, they get an error message - I do not expect anybody will understand that.  All that being said, lets move forward and see if it is an issue in the future.  Quick getting this in!

Issue #535, opened 2022-05-24
Remove assertion tab from span detail part


olha23:
design in #499

PR #582, opened 2022-05-27
Add assertion check in attribute list

jorgeepc:
This PR adds the `assertion check` information next to each attribute in the span detail list.

## Changes

- Attribute list.

## Fixes

- #535 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom

https://www.loom.com/share/ca75b63c0d6d45d59642e01ba75a37ae

## Coverage

Screen Shot 2022-05-27 at 10 49 20


web/src/components/AttributeList/AttributeList.tsx
```diff
@@ -1,17 +1,30 @@
...
+  assertions?: IResultAssertions;
   attributeList: TSpanFlatAttribute[];
   onCreateAssertion(attribute: TSpanFlatAttribute): void;
 }
 
-const AttributeList: React.FC<IAttributeListProps> = ({attributeList, onCreateAssertion}) => {
+const AttributeList: React.FC<IProps> = ({assertions, attributeList, onCreateAssertion}) => {
+  const onCopy = (value: string) => {
+    navigator.clipboard.writeText(value);
+  };
```
jorgeepc, line 15: Moving the `onCopy` function here to avoid creating a function for each attribute row.

web/src/components/SpanDetail/components/Http/Http.tsx
```diff
@@ -1,31 +1,36 @@
...
+import {ISpanDetailsComponentProps} from 'components/SpanDetail/SpanDetail';
+import * as S from 'components/SpanDetail/SpanDetail.styled';
+import {TSpanFlatAttribute} from 'types/Span.types';
 
 const filterRequestList = (attributeList: TSpanFlatAttribute[]) =>
   attributeList?.filter(a => HttpRequestAttributeList.includes(a.key) || a.key.includes('http.request'));
 
 const filterResponseList = (attributeList: TSpanFlatAttribute[]) =>
-  attributeList?.filter(a => HttpResponseAttributeList.includes(a.key) || a.key.includes('http.request'));
+  attributeList?.filter(a => HttpResponseAttributeList.includes(a.key) || a.key.includes('http.response'));
```
jorgeepc, line 14: Fixing this issue. We were displaying `request` fields in the response tab.

web/src/components/TestHeader/__tests__/TestHeader.test.tsx
```diff
@@ -4,6 +4,8 @@ import TestMock from '../../../models/__mocks__/Test.mock';
 import TestHeader from '../TestHeader';
 
 test('SpanAttributesTable', () => {
-  const result = render(<TestHeader onBack={jest.fn()} test={TestMock.model()} testState={TestState.CREATED} />);
+  const result = render(
+    <TestHeader onBack={jest.fn()} test={TestMock.model()} testState={TestState.CREATED} testVersion={1} />
```
jorgeepc, line 8: Types were failing for this component. It was missing the `testVersion`.

web/src/selectors/Assertion.selectors.ts
```diff
@@ -17,9 +17,46 @@ const selectAffectedSpanList = createSelector(stateSelector, paramsSelector, (st
   return trace?.spans.filter(({id}) => spanIdList.includes(id)) || [];
 });
 
+const selectResultAssertionsBySpan = createSelector(
```
jorgeepc, line 20: New selector that transforms the `run` data into a hashed object by attribute (filtered by span), something like this:
```js
{
    "cloud.availability_zone": {
        "failed": [
            {
                "id": "span[tracetest.span.type = \"database\"  service.name = \"pokeshop\"]",
                "label": "database pokeshop "
            },
            {
                "id": "span[tracetest.span.type = \"database\"]",
                "label": "database "
            }
        ],
        "passed": [
            {
                "id": "span[name = \"pg.query:SELECT\"  tracetest.span.type = \"database\"  service.name = \"pokeshop\"  db.system = \"postgresql\"  db.name = \"pokeshop\"  db.user = \"ashketchum\"]",
                "label": "pg.query:SELECT database pokeshop postgresql pokeshop ashketchum "
            }
        ]
    }
}
```

web/src/components/AttributeList/AttributeList.tsx
```diff
@@ -1,17 +1,30 @@
...
+interface IProps {
+  assertions?: IResultAssertions;
   attributeList: TSpanFlatAttribute[];
   onCreateAssertion(attribute: TSpanFlatAttribute): void;
 }
 
-const AttributeList: React.FC<IAttributeListProps> = ({attributeList, onCreateAssertion}) => {
+const AttributeList: React.FC<IProps> = ({assertions, attributeList, onCreateAssertion}) => {
+  const onCopy = (value: string) => {
+    navigator.clipboard.writeText(value);
```
xoscar, line 14: Take a look at what I did here https://github.com/kubeshop/tracetest/pull/583/files#diff-198edd331eb1e96f1a2636afe0b86500dcece0a7df2abe028ee881404afac599R14 I think it might overlap

web/src/components/AttributeRow/AttributeCheck.tsx
```diff
@@ -0,0 +1,54 @@
...
+  type: 'error' | 'success';
+}
+
+const AttributeCheck = ({items, type}: IProps) => {
+  const {setIsCollapsed} = useAssertionForm();
+
+  const handleOnClick = (id: string) => {
+    setIsCollapsed(true);
+    const resultListElement = document.getElementById(`assertion-${id}`);
+    const offsetTop = resultListElement?.offsetTop ?? 0;
```
xoscar, line 18: What do you think about moving this to a hook? like useScrollTo and we can call it by passing an id 😄

web/src/components/AttributeRow/AttributeCheck.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <div onClick={() => handleOnClick(items[0].id)}>
+        <S.CustomBadge status={type} text={items.length} />
+      </div>
+    );
+  }
+
+  return (
+    <div>
+      <Dropdown overlay={menuLayout} placement="bottomLeft" trigger={['click']}>
+        <div>
```
xoscar, line 46: Is this extra div necessary?

web/src/components/AttributeList/AttributeList.tsx
```diff
@@ -1,17 +1,30 @@
...
+  };
+
   return (
     <S.AttributeList>
       {attributeList.map(attribute => (
-        <AttributeRow attribute={attribute} key={attribute.key} onCreateAssertion={onCreateAssertion} />
+        <AttributeRow
+          assertionsFailed={assertions?.[attribute.key]?.failed}
+          assertionsPassed={assertions?.[attribute.key]?.passed}
```
cescoferraro, line 22: I would just pass the assertion as a prop.

cescoferraro (approved):
Great Job @jorgeepc

web/src/components/AttributeRow/AttributeRow.tsx
```diff
@@ -1,31 +1,46 @@
...
   return (
     <S.AttributeRow onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
       <S.TextContainer>
         <S.Text type="secondary">{key}</S.Text>
       </S.TextContainer>
-      <AttributeValue value={value} />
+
+      <S.AttributeValueRow>
+        <AttributeValue value={value} />
+        {passedCount > 0 && <AttributeCheck items={assertionsPassed!} type="success" />}
```
xoscar, line 36: Isn't this the same as !passedCount?

web/src/components/AttributeValue/AttributeValue.tsx
```diff
@@ -1,22 +1,27 @@
...
   value: string;
 }
 
-const AttributeValue: React.FC<IAttributeValueProps> = ({value, ...props}) => {
+const AttributeValue: React.FC<IProps> = ({value, ...props}) => {
   const [isCollapsed, setIsCollapsed] = useState(false);
-  const parsedValue = useMemo(() => (isJson(value) ? <JSONPretty data={value} /> : value), [value]);
+  const isJsonValue = useMemo(() => isJson(value), [value]);
+
+  if (isJsonValue) {
```
xoscar, line 16: I like the improvement

web/src/components/SpanDetail/SpanDetail.tsx
```diff
@@ -1,25 +1,41 @@
...
-  testId?: string;
+export interface ISpanDetailsComponentProps {
+  assertions?: IResultAssertions;
+  onCreateAssertion(attribute: TSpanFlatAttribute): void;
   span?: TSpan;
-  resultId?: string;
 }
 
-export interface TSpanDetailsComponentProps {
+export interface IResultAssertions {
```
xoscar, line 23: Let's move this to the Assertion types, and instead of using [key: string] I think its better to use Record<>

web/src/components/SpanDetail/components/Http/Http.tsx
xoscar, line 14: good catch!

web/src/components/AttributeList/AttributeList.tsx
jorgeepc: Thanks @cescoferraro 
I don't want to pass the whole object to the row if it's not needed. Here I'm looking for the `attribute.key` in the assertions object. With this we avoid having the assertions in every `AttributeRow`.

web/src/selectors/Assertion.selectors.ts
```diff
@@ -17,9 +17,46 @@ const selectAffectedSpanList = createSelector(stateSelector, paramsSelector, (st
...
+  stateSelector,
+  (_: RootState, testId: string, runId: string, spanId: string) => ({testId, runId, spanId}),
+  (state, {testId, runId, spanId}) => {
+    const {data} = endpoints.getRunById.select({testId, runId})(state);
```
xoscar, line 24: We grab the test run only once, and then it is updated using the dry-run, but that is only saved in the TestDefinitionSlice under the assertionResults entry. We should already have a selector to grab it from the store by passing the state so no biggies.

web/src/selectors/Assertion.selectors.ts
```diff
@@ -17,9 +17,46 @@ const selectAffectedSpanList = createSelector(stateSelector, paramsSelector, (st
...
+          label: `${selector.selectorList.map(({value}) => value).join(' ')} ${
+            selector.pseudoSelector?.selector ?? ''
+          }`,
+          result: assertion.spanResults.find(spanResult => spanResult.spanId === spanId),
+        }))
+      )
+      // Filter if it has result for the spanId
+      .filter(assertion => Boolean(assertion?.result))
+      // Hash items by attribute
+      .reduce((prev: {[key: string]: any}, curr) => {
```
xoscar, line 43: I tend to avoid using any, is this necessary or can we type everything that is part of this flow?

xoscar (changes requested):
Hey @jorgeepc great job man, I left some comments and concerns. Let me know what you think 😄

web/src/components/AttributeList/AttributeList.tsx
jorgeepc: Yes, I was thinking the same. You can merge. I can solve the conflicts.

web/src/components/AttributeRow/AttributeCheck.tsx
jorgeepc: Necessary for the Dropdown. A Badge cannot be a direct child of Dropdown because it doesn't have an onClick prop.

web/src/components/AttributeRow/AttributeRow.tsx
jorgeepc: It's the same as `!!passedCount` but we have a `no-implicit-coercion` lint rule. I could do `Boolean(passedCount)`, but in that case I prefer to do `passedCount > 0`

xoscar: I don't like that rule at all haha

web/src/components/AttributeList/AttributeList.tsx
jorgeepc: Done

web/src/components/AttributeRow/AttributeCheck.tsx
jorgeepc: Done

web/src/components/SpanDetail/SpanDetail.tsx
jorgeepc: Done

web/src/selectors/Assertion.selectors.ts
jorgeepc: Done

web/src/selectors/Assertion.selectors.ts
jorgeepc: @xoscar I don't think we should read from `TestDefinitionSlice`. But let's connect in slack about this.

xoscar (approved):

PR #584, opened 2022-05-27
fix: test definition versioning

mathnogueira:
This PR fixes the issue of updating the test definition and the test is on the same 

## Changes
- Version test definition
- Added option to disable cache in fixtures
- Added field to definitions table

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/http/controller.go
+++ b/server/http/controller.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/kubeshop/tracetest/server/analytics"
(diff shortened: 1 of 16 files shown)


server/integration/add_assertion_test.go
```diff
@@ -0,0 +1,71 @@
...
+	require.NoError(t, err)
+
+	bodyBuffer := bytes.NewBuffer(bodyJson)
+	req, err := http.NewRequest(http.MethodPut, url, bodyBuffer)
+	require.NoError(t, err)
+
+	resp, err := http.DefaultClient.Do(req)
+	require.NoError(t, err)
+
+	if resp.StatusCode != 204 {
```
schoren, line 53: you could directly use `require.Equal(t, 204, resp.StatusCode)`

server/http/controller.go
```diff
@@ -273,7 +273,21 @@ func (c *controller) SetTestDefinition(ctx context.Context, testID string, def o
...
+	}
+
+	newTest.Definition = newDefinition
+
+	newTest, err = c.testDB.UpdateTest(ctx, newTest)
+	if err != nil {
+		return openapi.Response(http.StatusInternalServerError, err.Error()), err
+	}
+
+	err = c.testDB.SetDefiniton(ctx, newTest, c.model.Definition(def))
```
schoren, line 290: This is not needed, since the `testdb.UpdateTest` already calls it.

schoren (approved):
LGTM!

PR #585, opened 2022-05-27
563 add revert functionality

cescoferraro:
This PR add revert draft assertion functionality

## Changes

- add revert draft assertion functionality
- remove transient-props warning 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/src/components/AssertionCard/AssertionCard.tsx
+++ b/web/src/components/AssertionCard/AssertionCard.tsx
@@ -9,7 +9,6 @@ import * as S from './AssertionCard.styled';
 import {useRevertDefinitionCallback} from './useRevertDefinitionCallback';
 
 interface TAssertionCardProps {
-  index: number;
   assertionResult: TAssertionResultEntry;
 
   onSelectSpan(spanId: string): void;
@@ -20,8 +19,7 @@ interface TAssertionCardProps {
 }
 
 const AssertionCard: React.FC<TAssertionCardProps> = ({
-  index,
-  assertionResult: {selector, resultList, selectorList, pseudoSelector, spanCount},
+  assertionResult: {id, selector, resultList, selectorList, pseudoSelector, spanCount},
   assertionResult,
   onSelectSpan,
   onDelete,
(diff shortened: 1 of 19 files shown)


web/src/components/AssertionCard/AssertionCard.tsx
```diff
@@ -51,7 +54,16 @@ const AssertionCard: React.FC<TAssertionCardProps> = ({
           <S.SpanCountText>{spanCountText}</S.SpanCountText>
         </div>
         <S.ActionsContainer>
-          {isDraft && <S.StatusTag>draft</S.StatusTag>}
+          {isDraft && (
+            <>
+              <S.StatusTag>draft</S.StatusTag>
+              <Tooltip color="white" title="Revert Assertion">
+                <Button type="link" data-cy="assertion-action-revert" onClick={resetDefinition}>
```
xoscar, line 61: I think semantic wise it would be better to use the selector as the unique identifier.

web/src/components/TraceNode/TraceNode.styled.ts
```diff
@@ -66,7 +66,7 @@ export const TraceNotch = styled.div<{spanType: SemanticGroupNames}>`
   align-items: center;
 
   span {
-    color: ${({spanType}) => getTextColor(spanType)};
+    color: ${({$spanType}) => getTextColor($spanType)};
```
xoscar, line 69: Thanks for all of these changes, this way we can get rid of the console warnings!

web/src/redux/slices/TestDefinition.slice.ts
```diff
@@ -56,6 +56,11 @@ const testDefinitionSlice = createSlice({
         return def;
       });
     },
+    revertDefinition(state, {payload: {index}}: PayloadAction<{index: number}>) {
+      state.definitionList = state.definitionList
+        .map((d, idx) => (idx === index ? state.initialDefinitionList[index] || undefined : d))
```
xoscar, line 61: I think you could achieve the same behavior using reduce, something like:

```
const {selector} = payload;
state.definitionList.reduce((list, definition) => {
  const foundDef  = state.initialDefinitionList.find((def) => def.selector === selector);
  
  return foundDef ? list.concat([foundDef]) : list;
}, []);
```

xoscar (changes requested):
Hey @cescoferraro great job man, I left a small concern, let me know what you think

web/src/redux/slices/TestDefinition.slice.ts
cescoferraro: I have tried the selector approach, but the selector can change.
So what you suggested does not work

web/src/components/AssertionCard/AssertionCard.tsx
cescoferraro: selector can also change, so this does not work

cescoferraro:
https://www.loom.com/share/d7153aef43ae41d9b20c8ae8246520cd

PR #586, opened 2022-05-27
Fix: test time

mathnogueira:
This PR fixes `CompletedAt` value. Now it is populated at the end of the run or when it fails.

I also added a few more time variables so we can know know much time each step took.

## Changes

- CompletedAt is filled at the end of the execution
- New time fields added to TestRun

## Fixes

- Time difference between Created and Completed was always 1s

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/runner_test.go
```diff
@@ -53,8 +53,6 @@ func TestPersistentRunner(t *testing.T) {
 
 		run2 := f.mockDB.runs[test2.ID.String()]
 		require.NotNil(t, run2)
-
-		assert.True(t, run1.CompletedAt.UnixNano() > run2.CompletedAt.UnixNano(), "test1 did not complete after test2")
```
mathnogueira, line 57: Does this test still makes sense? I feel that having a `runner` test mocking everything doesn't provide great value right now that we have integration tests.

server/executor/runner.go
```diff
@@ -70,6 +70,7 @@ func (r persistentRunner) Start(workers int) {
 					fmt.Println("persistentRunner exit goroutine")
 					return
 				case job := <-r.executeQueue:
+					job.run.StartedAt = time.Now()
```
schoren, line 73: I would put this either at the `processExecQueue` func, or remove the field, since it's almost the same value as `ServiceTriggeredAt`

server/executor/trace_poller.go
```diff
@@ -201,7 +202,7 @@ func (tp tracePoller) donePollingTraces(job tracePollReq, trace traces.Trace) bo
 func (tp tracePoller) handleTraceDBError(job tracePollReq, err error) {
 	run := job.run
 	if errors.Is(err, tracedb.ErrTraceNotFound) {
-		if time.Since(run.CompletedAt) < tp.maxWaitTimeForTrace {
+		if time.Since(run.CreatedAt) < tp.maxWaitTimeForTrace {
```
schoren, line 205: This feels weird. This time now includes the time elapsed since the trigger started, so we need to either change the config name, or change the checked field. I'd incline for the latter, so we can have different timeouts for execution and trace polling

schoren (changes requested):

schoren (approved):
LGTM, I think you can also regenerate the cli client as part of this pr too

Issue #567, opened 2022-05-26
Make assertion selector control fading of non selected spans


kdhamric:
When creating/editing an assertion, we should fade out the non selected spans so the user can see the impact of their selector change.

Also, we want to be able to click on a particular assertion in a test result and have it go into 'selected' mode.  This will outlining the assertion block and fade out the unselected spans in the graph view and the timeline view.

PR #590, opened 2022-05-30
Make assertion selector control fading of non selected spans

jorgeepc:
This PR implements the following features:
- When `creating/editing` an assertion, we fade out the non selected spans so the user can see the impact of their selector change.
- When `clicking` on a particular assertion block in a test result, we go into 'selected' mode. This outlines the assertion block and fade out the unselected spans in the graph and the timeline view.

## Changes

- Add new `affectedSpans` redux state.
- Update `Graph View` and `Timeline View` to handle the affected spans.

## Fixes

- #567 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Loom

https://www.loom.com/share/9b4764b3705e4100bfac080a9821bd45

## Coverage

Screen Shot 2022-05-30 at 16 17 04


web/src/components/AssertionCard/AssertionCard.styled.ts
```diff
@@ -2,12 +2,13 @@ import {DeleteOutlined, EditOutlined} from '@ant-design/icons';
 import {Tag, Typography} from 'antd';
 import styled, {css} from 'styled-components';
 
-export const AssertionCard = styled.div`
+export const AssertionCard = styled.div<{isSelected: boolean}>`
```
cescoferraro, line 5: use transient props
https://styled-components.com/docs/api#transient-props

web/src/components/Diagram/components/TimelineChart.styled.ts
```diff
@@ -1,9 +1,9 @@
-import styled from 'styled-components';
+import styled, {css} from 'styled-components';
 
-export const Container = styled.div<{barHeight: number}>`
+export const Container = styled.div<{barHeight: number; showAffected: boolean}>`
```
cescoferraro, line 3: use transient props
https://styled-components.com/docs/api#transient-props

cescoferraro (changes requested):
Looks great! Added a few comments to avoid console warnings

web/src/components/AssertionCard/AssertionCard.styled.ts
jorgeepc: Done ✅

web/src/components/Diagram/components/TimelineChart.styled.ts
jorgeepc: Done ✅

web/src/models/AssertionResults.model.ts
```diff
@@ -10,7 +10,7 @@ const AssertionResults = ({allPassed = false, results = []}: TRawAssertionResult
     resultList: results.map(({selector = '', results: resultList = []}) => ({
       id: uniqueId(),
       selector,
-      spanCount: AssertionService.getSpanCount(resultList),
+      spanIds: AssertionService.getSpanIds(resultList),
```
xoscar, line 13: 👍

web/src/redux/slices/TestDefinition.slice.ts
```diff
@@ -74,6 +76,12 @@ const testDefinitionSlice = createSlice({
     setAssertionResults(state, {payload}: PayloadAction<TAssertionResults>) {
       state.assertionResults = payload;
     },
+    clearAffectedSpans(state) {
```
xoscar, line 79: Maybe a change for the future but, I was thinking of decoupling both the affected spans and the selected span into a separated slice. As you can see, the setSelectedSpan function is being drilled into multiple component levels. We can solve it by having a central place for both.

xoscar (approved):
GJ @jorgeepc! I left some non-blocking suggestions. Let me know what you think

web/src/redux/slices/TestDefinition.slice.ts
jorgeepc: Yeah, like the idea. I can work on that after the MVP priorities.

PR #604, opened 2022-06-01
Fix: ensure consistent ordering of test definitions

schoren:
This PR ensures consistent ordering of tests definition and results. The issue comes from go's map implementation, which we use as the internal representation. I've implemented an `OrderedMap` that allows us to keep using the map approach but maintaining the order.

I think this is the smallest change possible, since we are keeping the internal concepts untouched, it just updates the implementation.

NOTE: this is a **VERY** backwards incompatible change. All saved tests/results are unparseable now.

## Changes

- Change internal model `Definition` and `Result` representation from maps to `OrderedMaps`
- Update consumers implementations, including tests
- Add tests to ensure order is consistent

## Fixes

- Inconsistent definition and results order

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/assertions/assertions.go
+++ b/server/assertions/assertions.go
@@ -10,7 +10,7 @@ import (
 func Assert(defs model.Definition, trace traces.Trace) (model.Results, bool) {
 	testResult := model.Results{}
 	allPassed := true
-	for spanQuery, asserts := range defs {
+	defs.Map(func(spanQuery model.SpanQuery, asserts []model.Assertion) {
 		spans := selector(spanQuery).Filter(trace)
 		assertionResults := make([]model.AssertionResult, 0)
 		for _, assertion := range asserts {
@@ -21,7 +21,7 @@ func Assert(defs model.Definition, trace traces.Trace) (model.Results, bool) {
 			assertionResults = append(assertionResults, res)
 		}
 		testResult[spanQuery] = assertionResults
-	}
+	})
 
 	return testResult, allPassed
 }
--- a/server/assertions/assertions.go
+++ b/server/assertions/assertions.go
@@ -7,8 +7,8 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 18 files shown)


server/assertions/assertions.go
```diff
@@ -7,10 +7,10 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-func Assert(defs model.Definition, trace traces.Trace) (model.Results, bool) {
-	testResult := model.Results{}
+func Assert(defs model.OrderedMap[model.SpanQuery, []model.Assertion], trace traces.Trace) (model.OrderedMap[model.SpanQuery, []model.AssertionResult], bool) {
```
mathnogueira, line 10: I think it would be nice to have an alias here.

server/model/tests_test.go
```diff
@@ -0,0 +1,157 @@
...
+	})
+
+	generateDef := func() model.OrderedMap[model.SpanQuery, []model.Assertion] {
+		def := (model.Test{}).Definition
+
+		def, _ = def.Add(model.SpanQuery("1"), []model.Assertion{{"1", comparator.Eq, "1"}})
+		def, _ = def.Add(model.SpanQuery("2"), []model.Assertion{{"2", comparator.Eq, "2"}})
+
+		return def
+	}
```
mathnogueira, line 37: You are duplicating that function. Instead of creating an anonymous function, create it outside the test functions so you can reuse the same function in both tests.

mathnogueira (changes requested):
A new type would help in the code readability

server/model/tests_test.go
schoren: It's not exactly the same. This function is being used in the `Map` and `Get` tests. The `Add` test is not exactly the same, since it also checks the errors.

mathnogueira: You're right.

Issue #160, opened 2022-04-07
Use text files to maintain test definitions


mathnogueira:
## Context
One important aspect of automated tests is that they are part of the application. So, when the application changes, some tests might change as well. It is important that those changes can be reviewed by peers to ensure it's testing what needs to be tested. However, if we keep application code and test definitions apart, verifying those changes becomes a challenge.

Therefore, it is very important that the test definitions can be versioned in the same repository as the application. This allows our users to be able to:

1. Keep track of all changes in the test suite;
2. Review changes before they are applied to the project;

I think it would be extremely beneficial to the project if we stored all test definitions in a text format (probably `YAML`) so our users could have all benefits of using `git`.

## Proposed solution

Instead of using a database to store all definitions, we start using `yaml` files instead. At the current state of the application, we could represent all our resources using the following yaml schemas:

```yaml
# Test schema
nameOfTheResource:
    name: string
    execution:
        endpoint: string
        method: string
        headers:
            - name: string
              value: string
        body: string
    assertions:
        - name: string
          attribute: string
          comparison: string
          value: string
```

The problem is that the definition file might become huge if we store all tests and definitions in the same place. We can improve that experience by enabling the user to define multiple files and import them from the main file.

Let's imagine we organize our files like this:
```
my_project/
├─ tracetest.yaml
├─ tracetest/
│  ├─ tests/
│  │  ├─ create_user_test.yaml
│  │  ├─ update_user_test.yaml
│  │  ├─ get_user_test.yaml
│  │  ├─ delete_user_test.yaml
```

Considering that `tracetest.yaml` is the main definitions file. We could define it as:

```yml
# my_project/tracetest.yml
tracetest:
    config:
        # Tracetest would scan all files in this folder recursively and generate 
        # one big file by merging all files from that directory into one.
        tracetestDir: ./tracetest/ 
    tests:
        - createUserTest
        - getUserTest
        - updateUserTest
        - deleteUserTest



# my_project/tracetest/tests/create_user_test.yml
createUserTest:
    # Test definitions here


# my_project/tracetest/tests/update_user_test.yml
updateUserTest:
    # Test definitions here


# my_project/tracetest/tests/delete_user_test.yml
deleteUserTest:
    # Test definitions here


# my_project/tracetest/tests/get_user_test.yml
getUserTest:
    # Test definitions here
```

#159 proposes the creation of a new type of resource. Referencing existing resources are extremely important for that idea to work.

[...]

PR #609, opened 2022-06-01
feat: cli create test command

mathnogueira:
This PR adds the command `tracetest test run --definition file.yml`. Right now it just creates the test if it doesn't exist yet. I'll open more PRs to add the functionality for waiting until the result is finished and to update an existing test.

Other than the run command itself, it also introduces the text format specified in #160.

Loom video explaining it: https://www.loom.com/share/f738c2c1ef8e4500a358b76fa7ad1e63

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test


api/http.yaml
```diff
@@ -23,7 +23,6 @@ components:
             $ref: "#/components/schemas/HTTPHeader"
         body:
           type: string
-          format: byte
```
mathnogueira, line 26: Basically, this means the `body` should be in `base64` format. But this is not true. I removed this so prism could add the correct validations to the requests I'm sending.

cli/testdata/definitions/valid_http_test_definition.yml
```diff
@@ -0,0 +1,35 @@
...
+  - messaging.message.payload contains 52
+- selector: span[span.name = "consume message from queue"]:last
+  assertions:
+  - messaging.message.payload contains 52
+- selector: span[span.name = "consume message from queue"]:last span[span.name = "import pokemon from pokeapi"]
+  assertions:
+  - http.status = 200
+- selector: span[span.name = "consume message from queue"]:last span[span.name = "save pokemon on database"]
+  assertions:
+  - db.repository.operation = "create"
```
mathnogueira, line 34: There are some details in the syntax here. When defining an assertion, the value must be quoted if it is a string. Right now we support `int`, `double`, and `string` values.

api/http.yaml
schoren: good catch

cli/testdata/definitions/valid_http_test_definition.yml
```diff
@@ -0,0 +1,33 @@
+name: POST import pokemon
+description: Import a pokemon using its ID
```
schoren, line 2: do we have a `description` in the UI?

cli/file/definition.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+
+	"github.com/kubeshop/tracetest/cli/definition"
+	"gopkg.in/yaml.v2"
+)
+
+func LoadDefinition(file string) (definition.Test, error) {
+	fileBytes, err := os.ReadFile(file)
+	if err != nil {
+		return definition.Test{}, fmt.Errorf("could not read file %s: %w", file, err)
```
schoren, line 14: nit: a more specific message could help debugging, like `could not read test definition file...`

cli/conversion/parser/assertion_parser.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	var assertionParserObject assertionParserObject
+	err = parser.ParseString("", assertionQuery, &assertionParserObject)
+	if err != nil {
+		return Assertion{}, fmt.Errorf("could not parse assertion (%s): %w", assertionQuery, err)
+	}
+
+	value := assertionParserObject.Value
+	if value[0:1] == "\"" {
```
schoren, line 60: maybe this looks easier to read?
```suggestion
	if value[0:1] == `"` {
```

mathnogueira:
Applied all changes! Thanks for the review @schoren

Issue #602, opened 2022-05-31
Assertion area should open to a minimum height when adding / editing


kdhamric:
The bottom assertion area needs to have a minimum height set when you:
- add an assertion
- edit an assertion
So that the user can see the Cancel & Save buttons.  If the screen is too small to give that minimum height in the bottom area, we need scroll bars so the user can scroll down (currently they cannot)

https://www.loom.com/share/91b007fc79f7439b958072421ecba0e1

cescoferraro:
this was fixed by #611

PR #611, opened 2022-06-01
new drawer behavior

cescoferraro:
This PR implements isCollapsable as a state instead of a boolean

## Changes

- Drawer behavior

## Fixes

- #602 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/components/AssertionForm/AssertionFormProvider.tsx
+++ b/web/src/components/AssertionForm/AssertionFormProvider.tsx
@@ -1,13 +1,14 @@
 import {noop} from 'lodash';
-import {useState, createContext, useCallback, useMemo, useContext, Dispatch, SetStateAction} from 'react';
 
 import {useTestDefinition} from 'providers/TestDefinition/TestDefinition.provider';
 import {useTestRun} from 'providers/TestRun/TestRun.provider';
+import {createContext, Dispatch, SetStateAction, useCallback, useContext, useMemo, useState} from 'react';
 import {useAppDispatch, useAppSelector} from 'redux/hooks';
 import {clearAffectedSpans, setSelectedAssertion} from 'redux/slices/TestDefinition.slice';
-import SelectorService from 'services/Selector.service';
 import TestDefinitionSelectors from 'selectors/TestDefinition.selectors';
+import SelectorService from 'services/Selector.service';
 import {TTestDefinitionEntry} from 'types/TestDefinition.types';
+import {DrawerState} from '../ResizableDrawer/ResizableDrawer';
 import {IValues} from './AssertionForm';
 import AssertionFormConfirmModal from './AssertionFormConfirmModal';
 
@@ -18,8 +19,8 @@ interface IFormProps {
 }
 
 interface ICreateAssertionModalProviderContext {
-  isCollapsed: boolean;
-  setIsCollapsed: Dispatch<SetStateAction<boolean>>;
+  drawerState: DrawerState;
+  setDrawerState: Dispatch<SetStateAction<DrawerState>>;
   isOpen: boolean;
   open(props?: IFormProps): void;
   close(): void;
@@ -32,8 +33,8 @@ const initialFormProps = {
 };
 
 export const Context = createContext<ICreateAssertionModalProviderContext>({
-  isCollapsed: false,
-  setIsCollapsed: noop,
+  drawerState: DrawerState.INITIAL,
+  setDrawerState: noop,
   isOpen: false,
   open: noop,
   close: noop,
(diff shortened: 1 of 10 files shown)


web/src/components/AssertionForm/AssertionFormProvider.tsx
```diff
@@ -32,26 +33,28 @@ const initialFormProps = {
...
   onSubmit: noop,
 });
 
-export const useAssertionForm = () => useContext(Context);
+export const useAssertionForm = () => useContext<ICreateAssertionModalProviderContext>(Context);
 
 const AssertionFormProvider: React.FC<{testId: string}> = ({children}) => {
   const dispatch = useAppDispatch();
-  const [isCollapsed, setIsCollapsed] = useState(true);
+  const [drawerState, setDrawerState] = useState<DrawerState>('INITIAL');
```
xoscar, line 49: I think more and more this is becoming the trace drawer provider instead of the assertion form provider, let's think about that for the future

web/src/components/TraceDrawer/TraceDrawerHeader.tsx
```diff
@@ -85,7 +90,7 @@ const TraceDrawerHeader: React.FC<IProps> = ({
           Add Assertion
         </S.AddAssertionButton>
         <span style={{marginLeft: 16}}>
-          <S.Chevron $isCollapsed={height === visiblePortion} />
+          <S.Chevron $isCollapsed={useChevronDirectionMemo(height, max, min)} />
```
xoscar, line 93: Isn't it better to read if we have this assigned to a constant first?

```
const isCollapsed = useChevronDirectionMemo(height, max, min)
```

web/src/components/TraceDrawer/index.css
```diff
@@ -0,0 +1,4 @@
+.ant-drawer-body {
```
xoscar, line 1: I think you can achieve the same behavior if you add this to the styled component for the resizable drawer by doing something like:

```
const CustomDrawer = styled(Drawer)`
 .ant-drawer-body {
    display: flex;
    flex-direction: column;
}
`
```

web/src/components/TraceDrawer/useChevronDirectionMemo.tsx
```diff
@@ -0,0 +1,13 @@
+import {useMemo} from 'react';
+
+export function useChevronDirectionMemo(height: number | undefined, max: number | undefined, min: number | undefined) {
```
xoscar, line 3: You can give a default to the optional parameters by doing something like:
```
export function useChevronDirectionMemo(height: = 0, max: = 0, min:  =0) {
```

This way you are saying both that the type is `number` and if undefined the default value should be `0`

xoscar (approved):
GJ @cescoferraro I left some questions and concerns. Let me know what you think

web/src/components/AssertionForm/AssertionFormProvider.tsx
jorgeepc: Agree. We can improve this later.

jorgeepc, line 49: What do you think about having a constant or enum for the states?

web/src/components/ResizableDrawer/ResizableDrawer.tsx
```diff
@@ -1,17 +1,21 @@
...
-  open: boolean;
+  visiblePortion: number;
   children: JSX.Element[];
 }
 
-const ResizableDrawer: React.FC<IProps> = ({open, children, min, max}: IProps) => {
+export type DrawerState = 'OPEN' | 'CLOSE' | 'INITIAL' | 'FORM';
+
+const ResizableDrawer: React.FC<IProps> = ({children, visiblePortion}: IProps) => {
+  const open = useAssertionForm().drawerState;
```
jorgeepc, line 15: Let's rename this. 
```js
const {drawerState} = useAssertionForm();
```

jorgeepc (changes requested):

web/src/components/ResizableDrawer/useSetIsCollapsedCallback.tsx
```diff
@@ -0,0 +1,14 @@
...
+export function useSetIsCollapsedCallback() {
+  const {setDrawerState} = useAssertionForm();
+  return useCallback<React.MouseEventHandler<HTMLDivElement>>(
+    e => {
+      e.stopPropagation();
+      setDrawerState(c => (c === 'OPEN' ? 'CLOSE' : 'OPEN'));
+    },
+    [setDrawerState]
+  );
+}
```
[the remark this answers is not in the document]
cescoferraro: This cant be on AssertionFormProvider because not everytime this is used is under the Provider

web/src/components/TraceDrawer/useChevronDirectionMemo.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export function useChevronDirectionMemo(height = 0, max = 0, min = 0) {
+  return useMemo(() => {
+    const isMax = (height || 0) > (max || 0) - 10;
+    if (isMax) return true;
+    const isMin = (height || 0) < (min || 0) + 10;
+    if (isMin) {
+      return false;
+    }
+    return false;
```
jorgeepc, line 11: is this the same?
```suggestion
    const isMax = (height || 0) > (max || 0) - 10;
    if (isMax) return true;
    return false;
```

cescoferraro: yes

PR #612, opened 2022-06-01
Add execution time in test run response

schoren:
This PR adds execution time to the api response. It returns the diff in seconds from `createdAt` to either `completedAt` if the test completed, or `time.Now()` if the test is not completed.

To ensure that run dates and other info is correctly set, I've centralized the state updating into functions within the model. 

## Changes

- add `executionTime` field to openapi `RestRun`
- add `ExecutionTime()` func to `model.Run`
- move run state change logic to encapsulated functions

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/model/run.go
```diff
@@ -0,0 +1,74 @@
...
+
+var (
+	Now = func() time.Time {
+		return time.Now()
+	}
+)
+
+func (r Run) Copy(testVersion int) Run {
+	r.Results = nil
+	r.Trace = nil
```
mathnogueira, line 18: Trace should be kept. This is being called by the `rerun` command and today it doesn't (and shouldn't) retrigger the transaction to run the test, it assumes the trace didn't change and only re-execute the assertion engine using the latest test definition.

mathnogueira (changes requested):

PR #616, opened 2022-06-01
rename `vendors` to `external` to avoid confilcts with go tooling

schoren:
We were using `server/vendors` path to keep external submodules used for otel proto generation. The problem is that `vendor` is a special directory for go mod, and we needed to append `-mod=readonly` to any `go` commands run in the server dir. This is not convenient, and not a very `go`y way of doing thinkgs.

The main problem we have is that the `server/external/tempo-idl` dir has some go files that are used just to copy to the generated files, so they don't compile, but the `go test` command is considering them part of the code, making the `go test ./...` to fail. uising `vendor` dir solved that because that dir is ignored by default.

Since the failing file was only used as a template, I've renamed it to `.tmp` so it's not contemplated as a go file, so won't be used to test, compile, etc

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira (approved):
Just a small comment. I'll approve it anyway

PR #639, opened 2022-06-02
Fix contains comparator not matching

schoren:
This PR fixes a bug detected when using the `contains` operator. The error was we were doing the comparations in the incorrect order (`expected < actual` instead of `actual < expected`), but we were also passing the conditions in an inverted order, so the error cancelled out. When using `contains`, we use the go `strings.Contains` function, and here the error doesn't cancel out.

## Fixes

- Fix expected/actual comparation order 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


mathnogueira (approved):
Lgtm

PR #641, opened 2022-06-02
Fix: Incorrect update behaviour when updating definition

schoren:
When rerunning a test after updating its definition, the new copied run kept some field values from the original, like dates and version. This caused the update to match a version it shouldn't.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira (approved):
LGTM!

Issue #627, opened 2022-06-02
Add test run button to the test page


kdhamric:
We need a 'Run Test' button to show at the top right when viewing any test run.  Some rules:
- it should replace the test status field as soon as the test goes to 'finished' state
- it should disappear when you have pending changes from altering assertions and reappear once you either revert or publish

If you are not using the latest version of the test, we should show a modal that says:

<Modal title="Version Mismatch" visible={isOpen} onOk={onConfirm} onCancel={onCancel}>
      <p>
        You are viewing version {currentVersion} of this test, and the latest version is {latestVersion}.
      </p>
      <p>Running the test will use the latest version of the test</p>
    </Modal>
    
    And we need a 'Run Test' or 'Cancel' Button

https://www.figma.com/file/5E7yr3kIPk1B08He0w9juk/V1?node-id=821%3A13445

PR #649, opened 2022-06-02
Add test run button to the test page

jorgeepc:
This PR adds a test run button to the test page. 

## Changes

- New `Run Test` button in test page.
- Fix a `race condition` with the dry-run call.

## Fixes

- #627 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom

https://www.loom.com/share/de343b15523540d7b346dd9e34e1c125

## Coverage

Screen Shot 2022-06-02 at 15 50 05


web/src/providers/TestDefinition/TestDefinition.provider.tsx
```diff
@@ -64,14 +66,15 @@ const TestDefinitionProvider: React.FC<IProps> = ({children, testId, runId}) =>
...
+    useTestDefinitionCrud({
+      testId,
+      runId,
+    });
 
   useEffect(() => {
-    init(run.result);
-  }, [init, isInitialized, reset, run.result]);
+    if (run.state === 'FINISHED') init(run.result);
+  }, [init, run.result, run.state]);
```
jorgeepc, line 77: This was causing a race condition with the `dryRun` call. 
I think we should call the init function only when the test state is finished, otherwise it was setting the `isInitialized` flag with an empty `definitionList`.

web/src/providers/TestDefinition/hooks/useTestDefinitionCrud.ts
```diff
@@ -44,10 +46,20 @@ const useTestDefinitionCrud = ({runId, testId}: IProps) => {
     const {id} = await dispatch(TestDefinitionActions.publish({testId, runId})).unwrap();
     setIsDraftMode(false);
     dispatch(clearAffectedSpans());
+    dispatch(setSelectedAssertion(''));
 
     navigate(`/test/${testId}/run/${id}`);
   }, [dispatch, navigate, runId, setIsDraftMode, testId]);
 
+  const runTest = useCallback(async () => {
+    const {id} = await dispatch(TestRunGateway.runTest(testId)).unwrap();
```
xoscar, line 55: I think these four dispatches could become an async thunk action

xoscar (approved):
Hey @jorgeepc GJ, I left a non-blocking suggestion.

One more thing, should we add an integration test for this scenario?

web/cypress/integration/Trace/RunTest.spec.ts
```diff
@@ -0,0 +1,29 @@
+import {createTest, deleteTest, getResultId, testId} from '../utils/common';
+
+describe('Run Test', () => {
```
xoscar, line 3: ez pz 🔥

PR #651, opened 2022-06-02
fix: quote string values in assertions when exporting a test

mathnogueira:
This PR quote string values in assertions when exporting a test

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


schoren (approved):
LGTM!

PR #652, opened 2022-06-02
feat: add wait for results flag in test run cmd

mathnogueira:
This PR adds a `wait-for-result` flag to the `test run` command. When provided, the command will only print an output when the result is in a final state (FINISHED or FAILED)

This also changes the output of the command to print the whole test and test run objects.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/actions/run_test_action.go
+++ b/cli/actions/run_test_action.go
@@ -179,15 +179,15 @@ func (a runTestAction) waitForResult(ctx context.Context, testId string, testRun
 		for {
 			select {
 			case <-ticker.C:
-				isReady, readyTestRun, err := a.isTestReady(ctx, testId, testRunId)
+				readyTestRun, err := a.isTestReady(ctx, testId, testRunId)
 				if err != nil {
 					lastError = err
 					wg.Done()
 					return
 				}
 
-				if isReady {
-					testRun = readyTestRun
+				if readyTestRun != nil {
+					testRun = *readyTestRun
 					wg.Done()
 					return
 				}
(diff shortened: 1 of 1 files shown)


cli/actions/run_test_action.go
```diff
@@ -154,3 +168,51 @@ func (a runTestAction) runTest(ctx context.Context, testID string) (openapi.Test
...
+func (a runTestAction) waitForResult(ctx context.Context, testId string, testRunId string) (openapi.TestRun, error) {
+	var testRun openapi.TestRun
+	var lastError error
+	var wg sync.WaitGroup
+	wg.Add(1)
+	ticker := time.NewTicker(1 * time.Second) // TODO: make this configurable
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
```
schoren, line 181: great candidate for a future websocket client

cli/actions/run_test_action.go
```diff
@@ -154,3 +168,51 @@ func (a runTestAction) runTest(ctx context.Context, testID string) (openapi.Test
...
+	wg.Wait()
+
+	if lastError != nil {
+		return openapi.TestRun{}, lastError
+	}
+
+	return testRun, nil
+}
+
+func (a runTestAction) isTestReady(ctx context.Context, testId string, testRunId string) (bool, openapi.TestRun, error) {
```
schoren, line 206: nit: in this case, it might make more sense to return an `*openapi.TestRun` and return `nil` instead of zeroed `openapi.TestRun`

PR #658, opened 2022-06-03
feat: support env variables in definition file

mathnogueira:
This PR enables users to define env variables in the definition file. Those variables will be replaced when the file is read from disk and its values are retrieved from the machine the CLI is running on.

Variables are declared as: `value: ${MY_VARIABLE}`. If the variable doesn't exist in the machine, it will be replaced with an empty string.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/cli/file/definition_test.go
+++ b/cli/file/definition_test.go
@@ -1,6 +1,7 @@
 package file_test
 
 import (
+	"os"
 	"testing"
 
 	"github.com/kubeshop/tracetest/cli/definition"
@@ -15,6 +16,7 @@ func TestLoadDefinition(t *testing.T) {
 		File               string
 		ExpectedDefinition definition.Test
 		ShouldSucceed      bool
+		EnvVariables       map[string]string
 	}{
 		{
 			Name:          "Should_parse_valid_definition_file",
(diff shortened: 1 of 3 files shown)


cli/file/definition_test.go
```diff
@@ -131,10 +133,55 @@ func TestLoadDefinition(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+		},
 	}
 
 	for _, testCase := range testCases {
+		envVariables := testCase.EnvVariables
```
mathnogueira, line 173: This is needed otherwise the cleanup function fails.

schoren: It would be nice if you already added that. I try to always run tests in parallel, because it's faster and helps detecting bugs and race conditions.

I usually do:
```
for _, c := range cases {
  t.Run(c.name, func(t *testing.T){
    cl := c
    t.Parallel()
    // the test
  })
}
```

There are 2 main differences with your code:
1. the copy of the case is done within the `t.Run` func
2. it copies the entire case. Otherwise you can get mixed results

cli/file/definition_test.go
```diff
@@ -131,10 +133,55 @@ func TestLoadDefinition(t *testing.T) {
...
 		t.Run(testCase.Name, func(t *testing.T) {
+			for key, value := range envVariables {
+				os.Setenv(key, value)
```
schoren, line 176: This will prevent this test to work parallel. Each parallel case might override the other at the same time, since `os.Setenv` is "global", so either refactor the code to have an `EnvGetter` injected, or totally ignore the parallel tests, since this is a cli, and parallelization shouldn't really be a problem

cli/testdata/definitions/valid_http_test_definition_with_env_variables.yml
```diff
@@ -0,0 +1,19 @@
...
+    url: http://pokemon-demo.tracetest.io/pokemon/import
+    method: POST
+    headers:
+    - key: Content-Type
+      value: application/json
+    authentication:
+      type: apiKey
+      apiKey:
+        key: X-Key
+        value: ${POKEMON_APP_API_KEY}
```
schoren, line 15: Niceeeee

schoren (changes requested):
It looks great, other than the tests comments

cli/file/definition_test.go
mathnogueira: You are right. I decided to make it as simple as possible because probably we don't need parallel tests here as it is a very small tool

cli/file/definition_test.go
mathnogueira: Decided to ignore the parallel tests.

schoren (approved):

PR #659, opened 2022-06-03
feat: add deleteTestRun endpoint

schoren:
This PR adds an endpoint to delete TestRuns

## Changes

- update openapi definitions
- implement endpoint

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/testdb/runs.go
+++ b/server/testdb/runs.go
@@ -56,6 +56,21 @@ func (td *postgresDB) UpdateRun(ctx context.Context, r model.Run) error {
 	return nil
 }
 
+func (td *postgresDB) DeleteRun(ctx context.Context, r model.Run) error {
+	stmt, err := td.db.Prepare("DELETE FROM runs WHERE id = $1")
+	if err != nil {
+		return fmt.Errorf("prepare: %w", err)
+	}
+	defer stmt.Close()
+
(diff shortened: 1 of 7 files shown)


mathnogueira (approved):

server/testdb/runs.go
```diff
@@ -56,6 +56,21 @@ func (td *postgresDB) UpdateRun(ctx context.Context, r model.Run) error {
 	return nil
 }
 
+func (td *postgresDB) DeleteRun(ctx context.Context, r model.Run) error {
+	stmt, err := td.db.Prepare("DELETE FROM runs WHERE id = $1")
```
xoscar, line 60: Aren't there any foreign keys we should worry about?

jorgeepc (approved):

server/testdb/runs.go
schoren: No, no entity has dependency with runs. It's the "last in the chain". Also, since we reference it by runID, we don't have to care about test version

mathnogueira: No thing relies on runs. We would have to worry about that if we were deleting tests or definitions.

xoscar: Cool, thanks!

PR #661, opened 2022-06-03
feat: better CLI configuration setup

mathnogueira:
This PR adds a command `tracetest configure` that allows you to define the configuration of the CLI. Now you don't have to manage the `config.yml` file manually.

Other than that, it also supports a global configuration inside `~/.tracetest/config.yml`. This is a fallback configuration that is only used if the current directory doesn't contain any config file in it. You also can write the configuration in the global config file by executing `tracetest configure --global`.

```shell
$ tracetest configure

# (current value: <value>) is only displayed if there is a config.yml file in the directory and it is 
# not empty. It shows the current value of that variable
Enter your Tracetest server URL (current value: http://localhost:8080): 
```

```shell
$ tracetest configure --global

# same thing as above, but the file ~/.tracetest/config.yml must exist for the current
# value be shown to the user
Enter your Tracetest server URL (current value: http://localhost:1234): 
```

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/cli/.gitignore
+++ b/cli/.gitignore
@@ -1,2 +1,3 @@
 tracetest
 openapitools.json
+config.yml
--- a/cli/actions/configure_action.go
+++ b/cli/actions/configure_action.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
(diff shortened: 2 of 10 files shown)


schoren (approved):
Awesome feature!

Issue #668, opened 2022-06-06
Issues with state when in draft mode


kdhamric:
Saw a couple issues when we are in draft mode:
- If we delete an assertion that we have 'selected' (ie graph view has other spans greyed out), the system does not revert to showing all the spans as 'normal' (ie they remain greyed out).
- If we delete an assertion we are adding in edit mode, should we go out of edit mode (ie should we not be in edit mode if there are 0 pending changes?)

https://www.loom.com/share/b203c64fbbac4c94a8439fb0cb1c6596

PR #673, opened 2022-06-06
🐛  668 Issues with state when in draft mode

xoscar:
This PR fixes the issues we had with the edit mode state. things like:
1. Deleting all of the assertions when reverting a newly created one.
2. Not resetting the edit state after there aren't any pending changes.
3. Affected spans should only be present when the assertion exists.

## Changes

- Moving the `isDraftMode` flag to the slice.
- Updating the affected spans selector to try to find the selected assertion first.
- Adding a validation for pending changes to remove the draft status.

## Fixes

- https://github.com/kubeshop/tracetest/issues/668

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/src/providers/TestDefinition/hooks/useDraftMode.ts
+++ b/web/src/providers/TestDefinition/hooks/useDraftMode.ts
@@ -1,8 +1,6 @@
-import {useEffect, useState} from 'react';
+import {useEffect} from 'react';
 
-const useDraftMode = (isDefaultDraftMode = false) => {
-  const [isDraftMode, setIsDraftMode] = useState(isDefaultDraftMode);
-
+const useDraftMode = (isDraftMode: boolean) => {
   useEffect(() => {
     if (isDraftMode) window.onbeforeunload = () => true;
     else window.onbeforeunload = null;
@@ -13,8 +11,6 @@ const useDraftMode = (isDefaultDraftMode = false) => {
       window.onbeforeunload = null;
     };
   }, []);
-
-  return {isDraftMode, setIsDraftMode};
 };
 
 export default useDraftMode;
--- a/web/src/redux/slices/TestDefinition.slice.ts
+++ b/web/src/redux/slices/TestDefinition.slice.ts
@@ -1,21 +1,7 @@
-import {CaseReducer, createSlice, PayloadAction} from '@reduxjs/toolkit';
-
+import {createSlice} from '@reduxjs/toolkit';
 import {TAssertionResults} from 'types/Assertion.types';
-import {TSpan} from 'types/Span.types';
-import {TTestDefinitionEntry} from 'types/TestDefinition.types';
-import TestDefinitionActions, {TChange} from '../actions/TestDefinition.actions';
-
-interface ITestDefinitionState {
-  initialDefinitionList: TTestDefinitionEntry[];
-  definitionList: TTestDefinitionEntry[];
-  assertionResults?: TAssertionResults;
-  changeList: TChange[];
-  isLoading: boolean;
-  isInitialized: boolean;
-  affectedSpans: string[];
-  selectedAssertion: string;
-  selectedSpan?: TSpan;
-}
+import {ITestDefinitionState, TTestDefinitionEntry, TTestDefinitionSliceActions} from 'types/TestDefinition.types';
+import TestDefinitionActions from '../actions/TestDefinition.actions';
 
 export const initialState: ITestDefinitionState = {
   initialDefinitionList: [],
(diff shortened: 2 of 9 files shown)


web/src/redux/slices/TestDefinition.slice.ts
```diff
@@ -26,6 +12,7 @@ export const initialState: ITestDefinitionState = {
   affectedSpans: [],
   selectedAssertion: '',
   selectedSpan: undefined,
+  isDraftMode: false,
```
xoscar, line 15: Moving the draft mode flag to redux

web/src/redux/slices/TestDefinition.slice.ts
```diff
@@ -38,27 +25,7 @@ export const assertionResultsToDefinitionList = (assertionResults: TAssertionRes
   }));
 };
 
-const testDefinitionSlice = createSlice<
-  ITestDefinitionState,
-  {
-    reset: CaseReducer<ITestDefinitionState>;
-    initDefinitionList: CaseReducer<ITestDefinitionState, PayloadAction<{assertionResults: TAssertionResults}>>;
-    addDefinition: CaseReducer<ITestDefinitionState, PayloadAction<{definition: TTestDefinitionEntry}>>;
```
xoscar, line 46: moved this to the types file

web/src/redux/slices/TestDefinition.slice.ts
```diff
@@ -109,7 +69,23 @@ const testDefinitionSlice = createSlice<
         return def;
       });
     },
+    revertDefinition(state, {payload: {originalSelector, selector}}) {
+      const initialDefinition = state.initialDefinitionList.find(
+        definition => definition.originalSelector === originalSelector
+      );
+
+      state.definitionList = initialDefinition
```
xoscar, line 77: We had a bug in this action, described in this video:

https://www.loom.com/share/6f8294b2c00748cd82f946c21b634e81

web/src/selectors/TestDefinition.selectors.ts
```diff
@@ -59,10 +59,17 @@ const TestDefinitionSelectors = () => ({
   selectDefinitionBySelector: createSelector(selectDefinitionList, selectorSelector, (definitionList, selector) =>
     definitionList.find(def => def.selector === selector)
   ),
-  selectAffectedSpans: createSelector(stateSelector, ({affectedSpans}) => affectedSpans),
+  selectAffectedSpans: createSelector(stateSelector, ({affectedSpans, assertionResults, selectedAssertion}) => {
+    const foundAssertion = assertionResults?.resultList.find(({selector}) => selector === selectedAssertion);
```
xoscar, line 63: We are checking if the selected assertion exists first

web/src/providers/TestDefinition/hooks/useDraftMode.ts
```diff
@@ -1,8 +1,6 @@
-import {useEffect, useState} from 'react';
-
-const useDraftMode = (isDefaultDraftMode = false) => {
-  const [isDraftMode, setIsDraftMode] = useState(isDefaultDraftMode);
+import {useEffect} from 'react';
 
+const useDraftMode = (isDraftMode: boolean) => {
```
jorgeepc, line 3: Not for this PR. but we could refactor this into a generic hook.

web/src/selectors/TestDefinition.selectors.ts
jorgeepc: This adds the validation at the selector level but the state still has the `affectedSpans` and `selectedAssertion`. I'm wondering if we should clear the state in the corresponding event (revert assertion). 
I'm ok with this approach for now.

jorgeepc (approved):
Great work Oscar.

web/src/redux/slices/TestDefinition.slice.ts
```diff
@@ -109,7 +69,23 @@ const testDefinitionSlice = createSlice<
...
+        ? state.definitionList.map(definition =>
+            definition.originalSelector === originalSelector ? initialDefinition : definition
+          )
+        : state.definitionList.filter(definition => definition.selector !== selector);
```
jorgeepc, line 81: Minor: I think we could do `definition.originalSelector !== originalSelector` here and avoid passing the `selector` to `revertDefinition`. But when adding a new assertion we also need to create the `originalSelector` field.

web/src/providers/TestDefinition/hooks/useDraftMode.ts
xoscar: Let me move it to the `hooks` folder and make the proper adjustments

web/src/redux/slices/TestDefinition.slice.ts
xoscar: Yeah, I was thinking the same. Let's me try to see how that goes

web/src/selectors/TestDefinition.selectors.ts
xoscar: Yes, I think it would be more complex to run validations against the different CRUD operations and decide when to drop the `selectedAssertion`

Issue #552, opened 2022-05-25
Make MultiSelectInputTag not trigger new selector workflow when hitting close button


cescoferraro:
ralated to #513

`Click an X to delete one filter does delete the filter... but it puts you in 'add a new filter' mode and prompts you for an attibute. I think you should have to click in the filed again to start the 'add filter' process.
 `

PR #674, opened 2022-06-06
Fix event propagation when removing selector

jorgeepc:
This PR fixes an event propagation when removing a value from the `MultiSelectInput` component. 
With this change we want to handle the following use case: When you click the X to delete one filter it puts you in the `add a new filter` mode and prompts you for an attribute. We want to avoid the add a new filter mode in this situation.

## Changes

- Antd lib uses the [rc-select](https://github.com/react-component/select) component. This component has an `onMouseDown` [event that was firing the function to open the options panel](https://github.com/react-component/select/blob/master/src/Selector/MultipleSelector.tsx#L127). With this change we are stopping the event propagation when doing a mouseDown in the close icon.

## Fixes

- #552 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Now

https://user-images.githubusercontent.com/3879892/172238887-4b654a14-d512-4692-a257-9ae049c2b30d.mov

## Before

https://user-images.githubusercontent.com/3879892/172238959-835c058c-3458-4f9e-8d6e-cda9e64b1ab6.mov

### DIFF
--- a/web/src/components/MultiSelectInput/MultiSelectInputTag.tsx
+++ b/web/src/components/MultiSelectInput/MultiSelectInputTag.tsx
@@ -15,10 +15,10 @@ const MultiSelectInputTag: React.FC<IProps> = ({value, entryListCount, onDeselec
     <S.SelectedTag
       {...props}
       closable={isLast}
-      onClose={event => {
-        event.stopPropagation();
-        event.preventDefault();
-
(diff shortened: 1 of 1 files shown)


xoscar (approved):
ez pz!

PR #686, opened 2022-06-07
Testing tracetest with tracetest

mathnogueira:
This PR adds the necessary CI configuration to run integration tests using tracetest. There are a some changes in the CLI that are needed for that to work. But this sets the correct path to start testing tracetest with tracetest.

Things that will be addressed on other PRs:

- [ ] We are testing a static version of tracetest, I'll open another PR to create two tracetest instances, one that we know it works and another using the latest docker image.
- [ ] CLI doesn't fail if the test fails: if an error happens, CLI exit code is still zero. I have to fix that
- [ ] CLI doesn't fail if an assertion fails: if the CLI works and retrieves the test result, it doesn't exit with an error if any of the assertions failed.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/config.yaml.sample
```diff
@@ -1,6 +1,6 @@
 postgresConnString: "host=postgres user=postgres password=postgres port=5432 sslmode=disable"
 jaegerConnectionConfig:
-  endpoint: jaeger-query:16685
+  endpoint: localhost:16685
```
mathnogueira, line 3: Changed it because we don't have a `jaeger-query` service on our docker-compose file.

tracetesting/ci/tracetest-testing-docker-compose.yml
```diff
@@ -0,0 +1,124 @@
...
+      - 8081:8081
+    depends_on:
+      postgres:
+        condition: service_healthy
+      jaeger:
+        condition: service_healthy
+      pokeshop:
+        condition: service_started
+    networks:
+      - local
```
mathnogueira, line 104: To run realistic tests, we would have to setup two `tracetest` services. One based on `sha-b6666d6` (probably v0.5.2), and another built from the latest changes in the code to be tested. As we progress, we can start using newer tracetest versions to test the latest tracetest changes.

mathnogueira:
This can be closed because #736 achieves the same goal.

PR #691, opened 2022-06-08
use last dragged height when reopening the drawer

cescoferraro:
This PR makes the drawer open using  to the previous dragged height

## Changes

- adds another state to DrawerState called RESIZING
- keeps tracks of the lastClosedHeight
- when lastClosedHeight exists we use it instead of the default OPEN height

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cescoferraro:
https://www.loom.com/share/718eabc38883474aa83774722412519b

PR #700, opened 2022-06-10
feat: http server with custom prefix

greut:
This PR... is a work in progress. Any help is much appreciated.

TODO:

- [ ] Figure out how to inject `<base href="/">` into the page, and rewrite the JS path to be relative to that

## Changes

- Allows to configure the HTTP server, host, port, and prefix

## Fixes

- Expose `tracetest` on a particular prefix, the routing will be done by Traefik.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,4 +29,5 @@ COPY --from=build-go /go/src/tracetest-server ./
 COPY --from=build-go /go/src/migrations/ ./migrations/
 COPY --from=build-js /app/build /app/html
 EXPOSE 8080/tcp
+EXPOSE 8081/tcp
 ENTRYPOINT ["/app/tracetest-server"]
--- a/server/app/app.go
+++ b/server/app/app.go
@@ -41,7 +41,7 @@ func New(config config.Config, db model.Repository, tracedb tracedb.TraceDB, tra
 }
 
 func spaHandler(staticPath, indexPath string, tplVars map[string]string) http.HandlerFunc {
-	var fileMatcher = regexp.MustCompile(`\.[a-zA-Z]*$`)
+	fileMatcher := regexp.MustCompile(`\.[a-zA-Z]*$`)
 	return func(w http.ResponseWriter, r *http.Request) {
 		if !fileMatcher.MatchString(r.URL.Path) {
 			tpl, err := template.ParseFiles(filepath.Join(staticPath, indexPath))
@@ -106,7 +106,7 @@ func (a *App) Start() error {
 
 	router := openapi.NewRouter(customController)
 
-	router.PathPrefix("/").Handler(
+	router.PathPrefix(fmt.Sprintf("/%s", a.config.Http.Prefix)).Handler(
 		spaHandler(
 			"./html",
 			"index.html",
@@ -130,8 +130,13 @@ func (a *App) Start() error {
 		wsRouter.ListenAndServe(":8081")
 	}()
 
-	log.Printf("HTTP Server started")
-	log.Fatal(http.ListenAndServe(":8080", router))
+	hostPort := ":8080"
+	if a.config.Http.Port > 0 {
+		hostPort = fmt.Sprintf("%s:%d", a.config.Http.Host, a.config.Http.Port)
+	}
+
+	log.Printf("HTTP Server started on %s", hostPort)
+	log.Fatal(http.ListenAndServe(hostPort, router))
 
 	return nil
 }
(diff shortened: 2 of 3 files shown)


schoren:
Hey @greut, thank you very much for your contribution. We'd like to understand what you are trying to achieve with this. I get that you want tracetest to live under a customisable path.
Could you provide some examples on the kind of deployment you are looking for?

schoren:
Got it. Seems like a useful setting , plus I like having the HTTP settings configurable. I think it would be great to have an integration test covering this thogh.
Also, this setting will impact the fronted too.

We're working on figuring out how to test this setting so it doesn't break anything

schoren:
@cescoferraro @jorgeepc do you know how hard would it be to make the web UI work on a subdomain?

cescoferraro:
@schoren it depends on the approach we are taking. 
You need a way to run the frontend image setting the public path via an env variable?

mathnogueira:
Hey @greut! This definitely would be a cool feature for tracetest. About integration testing, I experimented in a different branch (#701) how we could write an integration test for that feature. I created this branch to test it, you might want to take a look to see how we are testing stuff.

About injecting things on the HTML, we use a template file for that, you could modify it to create a `<base href="{{ .ServerPathPrefix }}">` there. We are doing that with Google Analytics stuff [here](https://github.com/kubeshop/tracetest/blob/main/web/public/index.html#L29) and [here](https://github.com/kubeshop/tracetest/blob/597f239c52ed9db1d1bdd031a567e6e266718837/server/app/app.go#L114)

jorgeepc:
@greut @schoren @mathnogueira to serve the UI I think we have different options. Of course we need to solve it at runtime and not build time. We could try something like this:
- Convert `static assets` paths to relative paths in the `index.html` file by setting `"homepage": "."` in our `package.json`. And then probably change the way we [serve them from the server](https://github.com/kubeshop/tracetest/blob/597f239c52ed9db1d1bdd031a567e6e266718837/server/app/app.go#L59) and adjust our [web router](https://github.com/kubeshop/tracetest/blob/main/web/src/components/Navigation/Router.tsx#L8).

jorgeepc:
@greut @schoren @mathnogueira to serve the UI I think we have different options. Of course we need to solve it at runtime and not build time. We could try something like this:
- Convert `static assets` paths to relative paths in the `index.html` file by setting `"homepage": "."` in our `package.json`. And then probably change the way we [serve them from the server](https://github.com/kubeshop/tracetest/blob/597f239c52ed9db1d1bdd031a567e6e266718837/server/app/app.go#L59) and adjust our [web router](https://github.com/kubeshop/tracetest/blob/main/web/src/components/Navigation/Router.tsx#L8).

greut:
> About injecting things on the HTML, we use a template file for that, you could modify it to create a `<base href="{{ .ServerPathPrefix }}">` there. We are doing that with Google Analytics stuff [here](https://github.com/kubeshop/tracetest/blob/main/web/public/index.html#L29) and [here](https://github.com/kubeshop/tracetest/blob/597f239c52ed9db1d1bdd031a567e6e266718837/server/app/app.go#L114).

I saw that bit and didn't find where the css/js where injected into that page. `%PUBLIC_URL%` could be a candidate to be changed via the `tplVars`.

mathnogueira:
> I saw that bit and didn't find where the css/js where injected into that page.

I'm not 100% sure, but it seems that they are injected during build time. @jorgeepc can you confirm that?

> %PUBLIC_URL% could be a candidate to be changed via the tplVars.

My understanding is that %PUBLIC_URL% is replaced during the web app build, so adding it on the backend will not have any effect. You would have to build another docker image with a specific path prefix for that to work.

I think that the best approach would be using a `<base>` tag for that. You would need to change [this const](https://github.com/kubeshop/tracetest/blob/main/web/src/redux/apis/TraceTest.api.ts#L13) to start using `document.baseURI` as part of the PATH.

jorgeepc:
@mathnogueira @greut 
Yes, `js` and `css` are injected during build time. And `PUBLIC_URL` is also replaced during build time.

mathnogueira:
Hey @greut . I tried understand what would be necessary to enable this feature and figured this is a complex feature that requires changes in multiple parts of tracetest. So, I updated my PR to update everything we would need to do to enable this feature. You can take a look at the implementation on the PR #701. As soon as it gets merged, I'll notify you.

greut:
@mathnogueira it was my feelings as well, thanks a lot for the hard work!

greut:
@mathnogueira it's working perfectly! :tada:

mathnogueira:
Awesome to hear that @greut! Let us know if you miss a feature you'd like to see in future versions of Tracetest. We really want feedback from people that are using it, and it would be awesome if we could chat sometime about what you want to achieve with Tracetest and hear your feedback about it. Could we set a meeting up to hear your feedback about the project?

PR #736, opened 2022-06-15
feat: test tracetest list endpoint

schoren:
This PR adds a tracetest test to test tracetest list endpoint. This serves as an example of how to do a setup step before running a test

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


mathnogueira (approved):
Amazing!

Issue #735, opened 2022-06-14
CLI doesn't support quote escaping in assertions


schoren:
I'm trying to run this file:

```yaml
name: Tests List
description: List all available tests
trigger:
  type: http
  http_request:
    url: ${TARGET_URL}/api/tests
    method: GET
    headers:
    - key: Content-Type
      value: application/json
testDefinition:
- selector: span[name="GET /api/tests" type="http"]
  assertions:
  - tracetest.span.duration <= 500
  - tracetest.response.status = 200
  - tracetest.response.body contains "\"id\":\"${TEST_ID}\""
```

with this command: `TEST_ID="asd" tracetest test run -d test.yaml` and get the following error:

```
{"level":"error","ts":1655246741.2517319,"msg":"failed to run test","error":"could not run definition: could not create test from definition: could not convert test definition: could not parse assertion (tracetest.response.body contains \"\\\"id\\\":\\\"asd\\\"\"): 1:39: invalid input text \"\\\\\\\":\\\\\\\"asd\\\\\\\"\\\"\""}
```

It works correctly with this definition:
```yaml
name: Tests List
description: List all available tests
trigger:
  type: http
  http_request:
    url: ${TARGET_URL}/api/tests
    method: GET
    headers:
    - key: Content-Type
      value: application/json
testDefinition:
- selector: span[name="GET /api/tests" type="http"]
  assertions:
  - tracetest.span.duration <= 500
  - tracetest.response.status = 200
  - tracetest.response.body contains "${TEST_ID}"
```

It also doesn't work if I use single quote(`'`) instead of double quote (`"`)

PR #740, opened 2022-06-15
fix: assertion query quoting

mathnogueira:
This PR enables the usage of single quotes to escape double quotes in queries.

#### "Escaping" double quotes by wrapping them in single quotes
```yaml
assertions:
    - tracetest.response.body contains '"some value that must be quoted"'
```

#### "Escaping" single quotes by wrapping them in double quotes
```yaml
assertions:
    - tracetest.response.body contains "'some value that must be singled quoted'"
```

#### Escaping quotes
```yaml
assertions:
    # Because we are adding JSON syntax to this string, and it contains a colon (:), 
    # if we don't wrap that string in single quotes, YAML will parse this string 
    # as a key-value instead of a string, breaking the YAML parser.
    - 'tracetest.response.body contains "\"id\": 52"'
```

## Fixes

- #735 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/cli/conversion/parser/assertion_parser.go
+++ b/cli/conversion/parser/assertion_parser.go
@@ -2,7 +2,6 @@ package parser
 
 import (
 	"fmt"
-	"strconv"
 
 	"github.com/alecthomas/participle/v2"
 	"github.com/alecthomas/participle/v2/lexer"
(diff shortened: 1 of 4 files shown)


mathnogueira:
> Do we have docs for the text format? If so, it would be great to update it

I'll write it

PR #748, opened 2022-06-15
add id to test

schoren:
This PR adds the test id to tracetest_test_list so it doesn't duplicate on each run

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira (approved):
Ok to be like that for a few days until we find a better approach

PR #749, opened 2022-06-15
fix: better variable injection

mathnogueira:
This PR injects variables in a better way using reflection and allows the injection to happen after the definition is loaded. This was done so the test definition doesn't have its variables replaced with the values present on the executor machine.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


schoren (approved):
LGTM!

PR #757, opened 2022-06-16
Improve websockets notifications

schoren:
This PR ensures that all testRun updates are correctly notified to subscriptions by changing the notification approach. Instead of having each step "manually" notifying when necessary, I've implemented a composite "RunUpdater" that encapsulates all the update related processes. Now, when a testRun is updated during the testing process, it will automatically notify any subscribers.

For cleanness, this PR also made the `ResourceID` a method of the `TestRun`, which required to have the `TestID` available on the TestRun entity. It is mainly an internal field, so no mappings to openapi where made.

## Changes

-

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

### DIFF
--- a/docs/websocket.md
+++ b/docs/websocket.md
@@ -15,7 +15,7 @@ Once the connection is open, you can send a message with the format:
 ```json
 {
     "type": "subscribe",
-    "resource": "test/{testID}/result/{resultID}"
+    "resource": "test/{testID}/run/{runID}"
 }
 ```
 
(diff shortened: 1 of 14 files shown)


mathnogueira (approved):
Nice. I liked the idea of executing the publish message every time we update the run on the database. GJ!

jorgeepc (approved):
🔥

PR #762, opened 2022-06-17
feat: test definition on server

mathnogueira:
This PR adds the API spec for the `/tests/{testId}/version/{version}/definition.yaml` route. This doesn't implement it yet because I wanted to merge all the changes in the definition before proceeding. 

## Changes

- Test definition uses `httpRequest` instead of `http_request`
- Replace `body` block with a `string` instead so we use the same HTTP definition on the open API Spec and definition file

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


server/openapi/model_test_1.go
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * OpenAPI definition for TraceTest endpoint and resources
+ *
+ * API version: 0.2.1
+ * Generated by: OpenAPI Generator (https://openapi-generator.tech)
+ */
+
+package openapi
+
+type Test1 struct {
```
schoren, line 12: This is not looking great. I guess this is because it already has a `Test` struct defined, right?
Maybe we can name this `FileTest` or use some kind of prefix/suffix so our naming is more self explainatory

server/openapi/model_test_definition_1.go
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * OpenAPI definition for TraceTest endpoint and resources
+ *
+ * API version: 0.2.1
+ * Generated by: OpenAPI Generator (https://openapi-generator.tech)
+ */
+
+package openapi
+
+type TestDefinition1 struct {
```
schoren, line 12: Same

api/openapi.yaml
```diff
@@ -388,3 +388,30 @@ paths:
...
+      summary: Get the test definition as an YAML file
+      description: Get the test definition as an YAML file
+      operationId: getTestVersionDefinitionFile
+      responses:
+        200:
+          description: OK
+          content:
+            application/yaml:
+              schema:
+                $ref: "./definition_file.yaml#/components/schemas/Test"
```
schoren, line 417: Maybe openapi is not the tool to define this, sine we're not actually using to transmit data to/from the server, but as a persistence file.  Adding to this, the naming conflicts with other existing entities.

I would leave this out of openapi, and just keep the format with autogenerated docs from the godocs in a `representation/yaml` package or somethig in that line.

schoren (changes requested):

schoren (approved):
LGTM!

PR #768, opened 2022-06-17
feat: yaml definition endpoint

mathnogueira:
This PR implements the endpoint that returns the yaml definition of a test version. The `encoding/yaml` package was copied from the CLI and adapted. Soon, we will remove that code from the CLI and centralize the object transformation on the server.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


api/openapi.yaml
```diff
@@ -412,6 +412,6 @@ paths:
         200:
           description: OK
           content:
-            application/yaml:
+            application/json:
               schema:
-                type: string
\ No newline at end of file
+                $ref: "./tests.yaml#/components/schemas/DefinitionFile"
```
mathnogueira, line 417: Well, openAPI doesn't allow us to define a `yaml` endpoint, so I'm wrapping it in a JSON object

tracetesting/funcs.bash
```diff
@@ -26,7 +26,11 @@ test() {
     echo -n "OK"
   else
     echo "FAIL"
+    echo "$name.json:"
     cat results/responses/$name.json
+    echo "$name.xml:"
+    cat results/$name.xml
```
mathnogueira, line 32: Wanted to debug a problem with the e2e tests and figured this file could help with that.

api/openapi.yaml
schoren: What happens when you use yaml? Isn't there a workaround this? If not, I'd put this outside openapi, or change the endpoint, because having a `/definition.yaml` endpoint that actually returns a json containing a yaml string is super confusing

mathnogueira: Yeah, it doesn't make sense. The only workaround would be writing the handler ourselves. The problem is that our HTTP clients would not contain that method as well :(

xoscar (approved):
LGTM 😄

schoren (approved):

PR #773, opened 2022-06-20
expand bottom panel when editing an assertion

cescoferraro:
This PR brings back the behavior of opening the bottom panel when editing an assertion.

## Changes

- call OpenBottomPanel when editing as assertion

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


xoscar:
LGTM, please take a look at the failing tests (try running the jobs) and this should be good to go

Issue #705, opened 2022-06-13
Use structured selectors from backend in frontend


jorgeepc:
**Context:** Right now selectors are strings, the frontend needs to parse them back into structured objects to display them in the UI and also to be able to update and add new ones. We need a way for the backend to send the same structured selectors to the FE so the work isn’t duplicated.

**AC1** 
The app should work with the provided structured selectors.

PR #774, opened 2022-06-20
feat: return selectors as a structure as well as a query

mathnogueira:
This PR changes the way our backend returns selectors. Now, instead of returning only the query, we also return a structured format as well.

Related to #705 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


.github/workflows/pull-request.yaml
```diff
@@ -211,10 +211,16 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          path: 'tracetest-main'
+          ref: 'main'
       - name: Build CLI
         run: |
-          cd cli
+          cd tracetest-main/cli
           make build
+          cp tracetest ../../cli
```
mathnogueira, line 223: `tracetest-integration` doesn't support the new format yet. So I'm building the client from the main branch so we can send a compatible request to it

schoren (approved):

PR #778, opened 2022-06-21
feat: Allow assertions at the span collection level

schoren:
This PR adds the ability to assert on the collection level. For now it only adds support for a `spans_collection.count` attribute, which maps to the amount of spans selected by the selector query.

## Changes

-

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/assertions/assertions.go
+++ b/server/assertions/assertions.go
@@ -27,7 +27,6 @@ func Assert(defs model.OrderedMap[model.SpanQuery, []model.Assertion], trace tra
 }
 
 func assert(a model.Assertion, spans []traces.Span) model.AssertionResult {
-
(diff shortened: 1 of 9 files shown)


mathnogueira (commented):
Great! I'm just not sure if `spans_collection.count` is a good name. @kdhamric what do you think?

mathnogueira:
> actually thought of using pseudo elements, like `:count` but the yaml parser didn't like that.

You have to wrap them in quotes for YAML to recognize it as a string and not as an invalid YAML syntax

mathnogueira (approved):

PR #783, opened 2022-06-21
import and export endpoints

mathnogueira:
This PR adds two new endpoints:

```
GET /api/tests/{testID/run/{runID}/export
POST /api/tests/import
```

Those endpoints can be used by users to send us their tests and runs and we can import them to our local instance to debug problems and help fix them.

## Questions
* Should we export tests and runs like this or should we generate the file and force the user browser to download it?

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
LGTM!

PR #792, opened 2022-06-22
fix: enforce order when validating pseudo classes

mathnogueira:
This PR fixes our pseudo-classes. Today pseudo-classes are not considering the `span.StartTime` to define its order. So, its kind of random :confused: . `:last` might give you different results for the same span.

## Changes

- Order spans by start time before applying pseudo-classes

## Fixes

- Order of spans

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/assertions/selectors/selector_test.go
```diff
@@ -27,7 +37,8 @@ var pokeshopTrace = traces.Trace{
 			"tracetest.span.type": "http",
 			"http.status_code":    "201",
 		},
-		Name: "POST /import",
+		Name:      "POST /import",
+		StartTime: t(0),
```
mathnogueira, line 41: I'm not sure about this function to get the span time. I created it based on physics time variables: `t0`, `t1`, `t2`, etc.

server/assertions/selectors/search.go
```diff
@@ -19,14 +21,23 @@ func filterSpans(rootSpan traces.Span, spanSelector SpanSelector) []traces.Span
 	})
 
 	uniqueSpans := filterDuplicated(filteredSpans)
+	orderedSpans := orderSpansByStartTime(uniqueSpans)
```
schoren, line 24: wouldn't it work better if we ordered the spans when we save them?

mathnogueira: It's more complex, but yeah, I think it would make it better mainly because we would show the spans in the right order in our DAG view

PR #799, opened 2022-06-23
Allow create tests with existing ID

schoren:
This PR allows creating tests with existing ID

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


cli/actions/run_test_action.go
```diff
@@ -79,26 +80,27 @@ func (a runTestAction) runDefinition(ctx context.Context, definitionFile string,
...
-		definition.Id = *test.Id
-		err = file.SaveDefinition(definitionFile, definition)
-		if err != nil {
-			return runTestOutput{}, fmt.Errorf("could not save definition: %w", err)
-		}
-	} else {
+	test, exists, err := a.createTestFromDefinition(ctx, definition)
+	if err != nil {
+		return runTestOutput{}, fmt.Errorf("could not create test from definition: %w", err)
+	}
```
mathnogueira, line 88: Instead of changing how `createTestFromDefinition` works, why don't you call `a.client.ApiApi.GetTest(ctx, id)`. If it returns an empty test, you create a new one. If not, you call `updateTestFromDefinition`.

schoren: I think this approach is more flexible, because we are not duplicating the logic of verifying if a test exists. If at some point we decide to change that logic to check things besides the ID, this method wouldn't work anymore and we'd need to update both the client and the server

mathnogueira (approved):

PR #801, opened 2022-06-23
fix: duplicated ids on span results

mathnogueira:
This PR fixes the bug where all selected spans had the same ID

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/assertions/assertions.go
+++ b/server/assertions/assertions.go
@@ -57,10 +57,11 @@ func assertIndividualSpans(a model.Assertion, spans []traces.Span) model.Asserti
 	res := make([]model.SpanAssertionResult, len(spans))
 	allPassed := true
 	for i, span := range spans {
+		spanID := span.ID
 		res[i] = apply(
 			a,
 			span.Attributes.Get(a.Attribute.String()),
-			&span.ID,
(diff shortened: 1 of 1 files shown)


server/assertions/assertions.go
```diff
@@ -57,10 +57,11 @@ func assertIndividualSpans(a model.Assertion, spans []traces.Span) model.Asserti
 	res := make([]model.SpanAssertionResult, len(spans))
 	allPassed := true
 	for i, span := range spans {
+		spanID := span.ID
 		res[i] = apply(
 			a,
 			span.Attributes.Get(a.Attribute.String()),
-			&span.ID,
```
mathnogueira, line 63: As `span` is a single variable that changes its value every time we iterate through spans and we are passing its address, the persisted value was the ID of the last span in `spans`.

By storing the id in a local variable inside the loop, we fix this problem. This new variable is reallocated every time the loop executes.

xoscar (approved):

jorgeepc (approved):

schoren (approved):

PR #809, opened 2022-06-27
feat: set ctx in assertion runner

mathnogueira:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/assertion_runner.go
```diff
@@ -5,42 +5,43 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/kubeshop/tracetest/server/assertions"
 	"github.com/kubeshop/tracetest/server/model"
 )
 
 type AssertionRequest struct {
+	Ctx  context.Context
```
schoren, line 12: We shouldn't use ctx as a property. Instead, we should always pass it as a func param (https://go.dev/blog/context-and-structs). You could instead do something like:

```
import "go.opentelemetry.io/otel/propagation"

type AssertionRequest struct {
  carrier propagation.TextMapCarrier
  //...
}
```

then, in line 66:
```
ctx := otel.GetTextMapPropagator().Extract(context.Background(), assertionRequest.carrier)
err := e.runAssertionsAndUpdateResult(ctx, assertionRequest)
```

schoren (changes requested):

server/executor/assertion_runner.go
```diff
@@ -55,13 +58,14 @@ func (e *defaultAssertionRunner) Stop() {
...
 
-func (e *defaultAssertionRunner) startWorker(ctx context.Context) {
+func (e *defaultAssertionRunner) startWorker() {
 	for {
 		select {
 		case <-e.exitChannel:
 			fmt.Println("Exiting assertion executor worker")
 			return
 		case assertionRequest := <-e.inputChannel:
+			ctx := e.getCtxFromRequest(assertionRequest)
```
schoren, line 68: nit: I think it would read easier if we have something like `assertionRequest.Context()` and have the func attached to that struct instead.

schoren (approved):

PR #814, opened 2022-06-27
Force install helm repo if it exists

aabedraba:
Avoids creating an error of existing helm chart

This PR...

## Changes

-

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -30,7 +30,7 @@ install_cert_manager(){
   kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.0/cert-manager.yaml
   echo
   echo "--> waiting for cert-manager"
-  kubectl wait --for=condition=ready pod -l app=webhook --namespace cert-manager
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -114,7 +114,7 @@ echo "Installing Tracetest"
 echo "----------------------------"
 echo
 
-helm repo add kubeshop https://kubeshop.github.io/helm-charts
+helm repo add --force-update kubeshop https://kubeshop.github.io/helm-charts
 helm repo update
```
schoren, line 118: Is this line still required? Or would the `force-update` flag already updates in any case?

aabedraba: I'm not so sure, that's why I kept it... 
From the CLI docs "--force-update  -- replace (overwrite) the repo if it already exists"

schoren (approved):
LGTM!

PR #790, opened 2022-06-22
Feat: poller instrumentation

mathnogueira:
This PR adds instrumentation for the `trace poller` process. Every time we try to get a trace, a new span is created. This also changes the integration test to check for the new instrumentation 

## Changes

- Add instrumentation to the trace polling process

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

PR #815, opened 2022-06-27
fix polling

mathnogueira:
This PR fixes the polling process. This bug was introduced by #790 
 
## Fixes

- trace polling. Before this patch, polling was marked as `true` the first time it got executed.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/integration/run_and_assert_pokemon_app_test.go
+++ b/server/integration/run_and_assert_pokemon_app_test.go
@@ -29,4 +29,5 @@ func TestExecutorIntegration(t *testing.T) {
 	}
 
 	assert.Equal(t, 2, count)
+	assert.Equal(t, 10, len(testRun.Trace.Flat))
 }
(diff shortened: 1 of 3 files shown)


server/integration/run_and_assert_pokemon_app_test.go
```diff
@@ -29,4 +29,5 @@ func TestExecutorIntegration(t *testing.T) {
 	}
 
 	assert.Equal(t, 2, count)
+	assert.Greater(t, len(testRun.Trace.Flat), 8)
```
mathnogueira, line 32: I left it as 8 because we have some spans that sometimes show up and other times don't show up at all (`SET statements in the database`). So, if none of them show up in the trace, we have `9` spans. If they show up, we get `11`

jorgeepc (approved):

xoscar (approved):

PR #818, opened 2022-06-28
reduce number of db queries for getting tests

schoren:
This PR changes the way we fetch tests from DB. instead of having two separated methods to get tests and definitions, we get the definitions as a `JOIN` when getting tests.

## Changes

- Adds tracetest for number of queries on select
- Use join to reduce number of queries.

## Fixes

-  several fixes on scripts/docker compose for running local tracetests

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/docker-compose.testrunner.yaml
+++ b/docker-compose.testrunner.yaml
@@ -9,6 +9,6 @@ services:
       - TARGET_URL=http://tracetest:8080
       - TRACETEST_MAIN_ENDPOINT=tracetest:8080
       - TRACETEST_TARGET_ENDPOINT=tracetest:8080
-      - DEMO_APP_URL=http://demo:8001
+      - DEMO_APP_URL=http://demo
     depends_on:
       - tracetest
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -63,26 +63,11 @@ services:
       timeout: 5s
       retries: 60
 
-  demodb:
-    container_name: demodb
-    image: postgres
-    environment:
-      POSTGRES_PASSWORD: postgres
-      POSTGRES_USER:  postgres
-      POSTGRES_DB: demodb
-    ports:
-      - 5433:5432
-    healthcheck:
-      test: pg_isready -U "$$POSTGRES_USER" -d "$$POSTGRES_DB"
-      interval: 1s
-      timeout: 5s
-      retries: 60
-
   demo:
     image: kubeshop/demo-pokemon-api:latest
     environment:
       REDIS_URL: cache
-      DATABASE_URL: postgresql://postgres:postgres@demodb:5433/pokeshop?schema=public
+      DATABASE_URL: postgresql://postgres:postgres@postgres:5432/postgres?schema=public
       RABBITMQ_HOST: queue
       POKE_API_BASE_URL: https://pokeapi.co/api/v2
       JAEGER_HOST: jaeger
@@ -90,7 +75,7 @@ services:
     ports:
       - 8001:80
     depends_on:
-      demodb:
+      postgres:
         condition: service_healthy
       cache:
         condition: service_healthy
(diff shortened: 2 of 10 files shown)


docker-compose.testrunner.yaml
```diff
@@ -9,6 +9,6 @@ services:
       - TARGET_URL=http://tracetest:8080
       - TRACETEST_MAIN_ENDPOINT=tracetest:8080
       - TRACETEST_TARGET_ENDPOINT=tracetest:8080
-      - DEMO_APP_URL=http://demo:8001
+      - DEMO_APP_URL=http://demo
```
schoren, line 12: when running within docker-compose, we need to point to the container port, not the exposed port

docker-compose.yaml
```diff
@@ -63,34 +63,19 @@ services:
...
-      interval: 1s
-      timeout: 5s
-      retries: 60
-
   demo:
     image: kubeshop/demo-pokemon-api:latest
     environment:
       REDIS_URL: cache
-      DATABASE_URL: postgresql://postgres:postgres@demodb:5433/pokeshop?schema=public
+      DATABASE_URL: postgresql://postgres:postgres@postgres:5432/postgres?schema=public
```
schoren, line 70: Using a single DB for both demo and tracetest to reduce stack complexity

run-tracetesting-tests.sh
```diff
@@ -1,6 +1,27 @@
...
+done
+
+
+docker compose -f docker-compose.yaml up -d --build --remove-orphans
 docker compose -f docker-compose.yaml -f docker-compose.testrunner.yaml build
 docker compose -f docker-compose.yaml -f docker-compose.testrunner.yaml run testrunner
-docker compose -f docker-compose.yaml stop
+
+if [ "$STOP" == "yes" ]; then
+  docker compose -f docker-compose.yaml stop
```
schoren, line 26: When debugging tests, it's useful to leave the containers running so you can access the UI and see what happened. so we can now call `./run-tracetesting-tests.sh --no-stop`

server/.dockerignore
```diff
@@ -0,0 +1,4 @@
+config.yaml
```
schoren, line 1: speed up docker build by ignoring useless files

tracetesting/definitions/test_create_with_id_exists.yml
```diff
@@ -1,5 +1,5 @@
-id: d85c3336-5048-4652-b564-63dfe2c83b14
-name: Test Create
+id: 7761457a-0a4b-4ada-8eab-d26c8ba48f90
```
schoren, line 1: we had the same ID as the `Create Test` test, so we where overriding it each run

tracetesting/definitions/test_create_with_id_notexists.yml
```diff
@@ -1,5 +1,5 @@
-id: d85c3336-5048-4652-b564-63dfe2c83b14
-name: Test Create
+id: 4ab71eaa-7356-454a-b1c0-8064e2a7135d
```
schoren, line 1: same here

tracetesting/run.bash
```diff
@@ -23,15 +23,15 @@ echo "TRACETEST_MAIN_ENDPOINT: $TRACETEST_MAIN_ENDPOINT"
 echo "TRACETEST_TARGET_ENDPOINT: $TRACETEST_TARGET_ENDPOINT"
 echo "DEMO_APP_URL: $DEMO_APP_URL"
 
-tee config.main.yml << END
+cat << EOF > config.main.yml
```
schoren, line 26: `tee` outputs to stdout in addition of appending to the file. Using `cat` does not, so the ouput looks cleaner

PR #836, opened 2022-06-30
feat: convert time into the right time unit

mathnogueira:
# Problem
Today we show time attributes (especially `tracetest.span.duration`) as a number. That number represents the number of milliseconds that the took to execute the operation. However, there is no indication that is in `ms`, and worse, sometimes some operations take less than one millisecond and the duration appears as `0`. 

> :information_source: Note: Before that, we were displaying the duration as the number of nanoseconds to execute the operation. It was precise, but the number was too big to be easily understood.

# Solution
Convert the duration from nanoseconds to a nice human-readable value and show the time unit. So instead of showing `1400000000`, Tracetest is going to show it as `1.4s` instead. It converts from nanoseconds (`ns`) to microseconds (`μs`), milliseconds(`ms`), seconds(`s`), minutes(`m`), and hours(`h`). Any value bigger than 24 hours will be displayed as hours anyway.

> Note: we are only showing decimal digits for seconds and above. So if duration is `1200000ns` it would be converted to `1ms` instead of `1.2ms`.

## Description

This PR adds code to convert the span attributes, assertion values, and result values into the right time unit. This is done in the mapping layer between the controller and the actual application code.

The values from the trace are kept untouched. So only the representation layer has changed.

It is executing the conversion during the trace fetch so we just have to convert it once.

## Screenshots of this working on the DAG page

image

image

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -155,6 +155,8 @@ jobs:
           --set tracingBackend=jaeger \
           --set jaegerConnectionConfig.endpoint="jaeger-query.tracetest.svc.cluster.local:16685" \
           --set telemetry.jaeger.host="jaeger-agent.tracetest.svc.cluster.local" \
+          --set poolingConfig.maxWaitTimeForTrace="60s" \
+          --set poolingConfig.retryDelay="5s" \
           --set ingress.enabled=false \
           --wait --timeout 3m
 
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -156,7 +156,7 @@ jobs:
           --set jaegerConnectionConfig.endpoint="jaeger-query.tracetest.svc.cluster.local:16685" \
           --set telemetry.jaeger.host="jaeger-agent.tracetest.svc.cluster.local" \
           --set poolingConfig.maxWaitTimeForTrace="60s" \
-          --set poolingConfig.retryDelay="5s" \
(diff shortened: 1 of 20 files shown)


server/traces/time_converter.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"math"
+)
+
+const (
+	NANOSECOND_SCALE  = 0
+	MICROSECOND_SCALE = 1
+	MILLISECOND_SCALE = 2
```
mathnogueira, line 11: Those are not used, I left them here to make it easier to understand what those numbers represent

.github/workflows/pull-request.yaml
```diff
@@ -155,6 +155,8 @@ jobs:
           --set tracingBackend=jaeger \
           --set jaegerConnectionConfig.endpoint="jaeger-query.tracetest.svc.cluster.local:16685" \
           --set telemetry.jaeger.host="jaeger-agent.tracetest.svc.cluster.local" \
+          --set poolingConfig.maxWaitTimeForTrace="60s" \
+          --set poolingConfig.retryDelay="1s" \
```
mathnogueira, line 159: Integration test was timing out because of the maxWaitTimeForTrace

mathnogueira:
Should we keep using `μs` or should we rename it to `us` for simplicity?

xoscar (approved):

PR #856, opened 2022-07-04
add duration fields to AssertionForm

cescoferraro:
This PR adds specific duration fields for assertion checks

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/src/components/AssertionForm/AssertionFormCheck.tsx
+++ b/web/src/components/AssertionForm/AssertionFormCheck.tsx
@@ -1,0 +1,113 @@
+import {PlusOutlined} from '@ant-design/icons';
+import {Form, FormInstance, Input, Select} from 'antd';
+import {FormListFieldData} from 'antd/lib/form/FormList';
+import {capitalize} from 'lodash';
+import {useMemo} from 'react';
+import {durationRegExp} from '../../constants/Common.constants';
+import {CompareOperator} from '../../constants/Operator.constants';
+import CreateAssertionModalAnalyticsService from '../../services/Analytics/CreateAssertionModalAnalytics.service';
+import OperatorService from '../../services/Operator.service';
+import {TAssertion} from '../../types/Assertion.types';
+import {IValues} from './AssertionForm';
+import * as S from './AssertionForm.styled';
+import {DurationFields} from './DurationFields';
+
+const operatorList = Object.values(CompareOperator).map(value => ({
+  value: OperatorService.getOperatorSymbol(value),
+  label: capitalize(OperatorService.getOperatorName(value)),
+}));
+
+const getIsValid = ({attribute, comparator, expected}: TAssertion): boolean =>
+  Boolean(attribute && comparator && expected);
+
+interface IProps {
+  add(): void;
+  remove(name: number): void;
+  form: FormInstance<IValues>;
+  field: Pick<FormListFieldData, never>;
+  name: number;
+  attributeOptionList: JSX.Element[];
+  index: number;
+  length: number;
+  assertionList: TAssertion[];
+}
+
+export const AssertionFormCheck = ({
+  attributeOptionList,
+  field,
+  index,
+  length,
+  name,
+  assertionList,
+  form,
+  add,
+  remove,
+}: IProps) => {
+  const assertion = assertionList?.[index];
+  const match = useMemo(() => assertion?.expected?.match(durationRegExp), [assertion?.expected]);
+  return (
+    <S.Check>
+      <Form.Item
+        {...field}
+        name={[name, 'attribute']}
+        style={{margin: 0}}
+        rules={[{required: true, message: 'Attribute is required'}]}
+        data-cy="assertion-check-attribute"
+        id="assertion-check-attribute"
+      >
+        <S.Select style={{margin: 0}} placeholder="Attribute" showSearch>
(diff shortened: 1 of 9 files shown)


web/src/components/AssertionForm/AssertionFormCheck.tsx
```diff
@@ -0,0 +1,111 @@
...
+      >
+        <S.Select style={{margin: 0}} placeholder="Assertion Type">
+          {operatorList.map(({value, label}) => (
+            <Select.Option key={value} value={value}>
+              {label}
+            </Select.Option>
+          ))}
+        </S.Select>
+      </Form.Item>
+      {assertion?.attribute === TraceTestAttributes.TRACETEST_SPAN_DURATION ? (
```
xoscar, line 75: We should be checking the attributes using a regex that will tell us if we need to display this extra input or not like:
```
^([0-9]+(\\.[0-9]+)?)(ns|μs|ms|s|m|h)$
```

Instead of having to match for a specific field.

web/src/components/AssertionForm/AssertionFormCheck.tsx
```diff
@@ -0,0 +1,111 @@
...
+  value: OperatorService.getOperatorSymbol(value),
+  label: capitalize(OperatorService.getOperatorName(value)),
+}));
+
+const getIsValid = ({attribute, comparator, expected}: TAssertion): boolean =>
+  Boolean(attribute && comparator && expected);
+
+interface IProps {
+  add: any;
+  remove: any;
```
xoscar, line 24: Let's try to avoid using any

web/src/components/AssertionForm/AssertionFormCheck.tsx
```diff
@@ -0,0 +1,111 @@
...
+  tAssertions: TAssertion[];
+}
+
+export const AssertionFormCheck = ({
+  attributeOptionList,
+  field,
+  index,
+  length,
+  name,
+  tAssertions,
```
xoscar, line 40: What does `tAssertions` mean?

web/src/components/AssertionForm/DurationFields.tsx
```diff
@@ -0,0 +1,61 @@
...
+        type="number"
+      />
+      <S.Select
+        onChange={e => setDurationFieldValue(form, index, {value: state.value, duration: e as Duration})}
+        value={state.duration}
+        style={{margin: 0}}
+        placeholder="Assertion Type"
+      >
+        {[
+          {label: 'nanoseconds', value: 'ns'},
```
xoscar, line 47: Let's move this to a constant

web/src/components/AssertionForm/DurationFields.tsx
```diff
@@ -0,0 +1,61 @@
+import {FormInstance, Input, Select} from 'antd';
+import {TAssertion} from '../../types/Assertion.types';
+import {IValues} from './AssertionForm';
+import * as S from './AssertionForm.styled';
+
+type Duration = 'ms' | 's';
```
xoscar, line 6: It wont only support `ms` and `s`. It should support `ns|μs|ms|s|m|h`

xoscar (changes requested):
Hey @cescoferraro I left some comments and concerns. Let me know what you think!

web/src/components/AssertionForm/AssertionFormCheck.tsx
cescoferraro: i think doing this might the regex might fail depending on what the users types for value

web/src/components/AssertionForm/AssertionFormCheck.tsx
cescoferraro: done

web/src/components/AssertionForm/AssertionFormCheck.tsx
cescoferraro: done

web/src/components/AssertionForm/DurationFields.tsx
cescoferraro: created an enum and created this array based on that

web/src/components/AssertionForm/DurationFields.tsx
cescoferraro: made this a enum instead

web/src/components/AssertionForm/AssertionFormCheck.tsx
xoscar: We should be able to lock the value input for only numbers in case we find the value matches a duration. I want to avoid having specific custom logic just for one attribute, does it make sense?

xoscar (approved):

web/src/components/AssertionForm/AssertionFormCheck.tsx
```diff
@@ -0,0 +1,112 @@
...
+  index,
+  length,
+  name,
+  assertionList,
+  form,
+  add,
+  remove,
+}: IProps) => {
+  const assertion = assertionList?.[index];
+  const match = useMemo(() => assertion?.expected?.match(/(\d+)(ns|μs|ms|s|m|h)/), [assertion?.expected]);
```
xoscar, line 46: Let's move this regex to a constant

PR #877, opened 2022-07-06
fix: collector deployment

mathnogueira:
This PR fixes our collector deployment and its configuration.

## Fixes

- opentelemetry collector automatic deployment
- collector configuration


.github/workflows/pull-request.yaml
```diff
@@ -8,7 +8,7 @@ on:
     branches:
       - main
     paths:
-      - .github/workflows/**
+      - .github/workflows/pull-request.yaml
```
mathnogueira, line 11: This was being executed every time we added new workflows. It only makes sense if we only execute the flow if the flow itself has changed.

.github/workflows/update-dependency-deployments.yml
```diff
@@ -2,20 +2,19 @@ name: Update dependency deployments
...
-  apply:
+  deploy-collector:
+    name: Deploy OTEL Collector
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
       uses: actions/checkout@v3
 
-    - name: Install envsubst
-      run: apt-get install -y gettext-base
```
mathnogueira, line 17: Ubuntu already has this tool available

k8s/collector.yml
```diff
@@ -35,7 +35,7 @@ data:
       pipelines:
         traces:
           receivers: [otlp]
-          processors: [batch, probabilistic_sampler]
+          processors: [probabilistic_sampler, batch]
```
mathnogueira, line 38: batch must be the last processor, otherwise we might have some weird behavior

k8s/collector.yml
```diff
@@ -11,13 +11,13 @@ data:
...
 
     processors:
-      batch:
-
       # Data sources: traces
       probabilistic_sampler:
         hash_seed: 22
         sampling_percentage: 10
 
+      batch:
```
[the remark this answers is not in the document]
mathnogueira: Yeah, but I need to declare it in order for it to work. In this case, it's a batch processor with default config

.github/workflows/update-dependency-deployments.yml
```diff
@@ -28,8 +27,8 @@ jobs:
         credentials: ${{ secrets.GKE_SA_KEY }}
 
     - name: Apply
-      run: ./k8s/deploy-collector.sh
-
-
-
-
+      run: |
+        chmod +x ./k8s/deploy-collector.sh
```
schoren, line 31: You don't need this. Git already tracks permissions, so if you commit the file with 755 it will be executable

k8s/collector.yml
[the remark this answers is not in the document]
schoren: Shouldn't it be an empty object? Like `batch: {}`

schoren (approved):

PR #889, opened 2022-07-07
Fix frontend analytics bug

jorgeepc:
This PR fixes a frontend analytics bug.

## Changes

- Analytics service

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/src/services/Analytics/Analytics.service.ts
+++ b/web/src/services/Analytics/Analytics.service.ts
@@ -1,7 +1,7 @@
 import GA4React from 'ga-4-react';
 import {Categories} from '../../constants/Analytics.constants';
 
-const {analyticsEnabled = 'true', measurementId = ''} = window.ENV || {};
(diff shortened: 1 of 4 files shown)


web/src/services/Analytics/Analytics.service.ts
```diff
@@ -11,11 +11,9 @@ type TAnalyticsService = {
   event<A>(category: Categories, action: A, label: string): void;
 };
 
-const initializePromise = instance.initialize();
```
jorgeepc, line 14: We were initializing the analytics instance always. We already have the logic to initialize it (depending on the `analyticsEnabled` flag) in the `AnalyticsProvider`.

xoscar: I think I tried doing this to debug a problem with our integration tests. Great catch @jorgeepc

xoscar (approved):
LGTM

PR #903, opened 2022-07-13
Actual grpc trigger

schoren:
This PR adds gRPC triggering capabilites.

## Changes

-

## Fixes

-

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test


server/executor/trigger/grpc.go
```diff
@@ -0,0 +1,197 @@
...
+
+	options := grpcurl.FormatOptions{
+		EmitJSONDefaultFields: true,
+		IncludeTextSeparator:  true,
+		AllowUnknownFields:    true,
+	}
+
+	rf, _, err := grpcurl.RequestParserAndFormatter(grpcurl.FormatJSON, desc, strings.NewReader(tReq.Request), options)
+	if err != nil {
+		return response, fmt.Errorf("failed to construct request parser and formatter for %q", err)
```
mathnogueira, line 82: I think you are either using the wrong variable or the wrong format option:

Either it's `%q` but the variable is not `err` or it should be `%w` and the variable is `error` indeed.

mathnogueira (approved):
LGTM! Just two small things

server/executor/trigger/grpc.go
schoren: good catch!

mathnogueira:
@schoren feel free to merge it

Issue #905, opened 2022-07-13
Implement segment io for FE analytics


xoscar:
The company is moving to use a centralized method for analytics. For that we have decided to use https://segment.com/ as the primary vendor.

Currently, the FE application uses googles analytics to provide user data, we need to move away from that to use segment's framework instead.

Acceptance criteria:
**AC1**
As a developer, use Segments framework to send user data.
And remove Google analytics from the app.

**AC2**
As a user installing the tracetest app.
If I choose to disable analytics.
The UI should not trigger any request to the Segment API.

**AC3**
As a developer configuring the tracetest app.
I can choose in what environment to send analytics data (demo/installations).

PR #915, opened 2022-07-14
👀 905 Implement segment io for FE analytics

xoscar:
This PR updates the FE analytics framework to use segment IO instead of google analytics.

## Changes

- Adding the segment script.
- Updating the analytics service to use the segment functionality.

## Fixes

- https://github.com/kubeshop/tracetest/issues/905

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/f8b46946e8de4c47b294c5286a59d0b9


web/public/index.html
```diff
@@ -32,26 +29,31 @@
...
         var base = document.createElement('base');
-        base.href = window.ENV.serverPathPrefix
+        base.href = window.ENV.serverPathPrefix;
         document.getElementsByTagName('head')[0].appendChild(base);
-      };
+      }
 
       injectVariables();
+      
+      if (window.ENV.analyticsEnabled === 'true') {
```
xoscar, line 52: We are using the same keys as we did for the backend

xoscar: And we'll avoid loading the script if analytics are disabled

web/src/components/WithAnalytics/WithAnalytics.tsx
```diff
@@ -0,0 +1,16 @@
+import {useEffect} from 'react';
+import AnalyticsService from '../../services/Analytics/Analytics.service';
+
+const withAnalytics = <P extends object>(Component: React.ComponentType<P>, name: string) => {
```
xoscar, line 4: Page wrapper to load the analytics event on land

web/src/services/Analytics/Analytics.service.ts
```diff
@@ -1,23 +1,27 @@
...
-  const event = async <A>(category: Categories, action: A, label: string) => {
+const AnalyticsService = (): TAnalyticsService => ({
+  async event<A>(category: Categories, action: A, label: string) {
     if (!isEnabled) return;
-    instance.event(String(action), label, category);
-  };
-
-  return {event};
-};
+    analytics.track(String(action), {
```
xoscar, line 16: Moving out of GA to use Segment

web/public/index.html
jorgeepc: is it ok to have the `_writeKey` hardcoded?

web/src/pages/Home/HomeActions.tsx
```diff
@@ -8,17 +9,18 @@ const {onCreateTestClick} = HomeAnalyticsService;
...
-        data-cy="create-test-button"
-        type="primary"
-        href="/create-test"
-        onClick={() => {
-          onCreateTestClick();
-        }}
-      >
-        Create Test
-      </S.CreateTestButton>
+      <Link to="/test/create">
```
jorgeepc, line 12: 🔥

web/src/components/WithAnalytics/WithAnalytics.tsx
jorgeepc, line 4: I see the value of the HOC, what's your opinion about `HOC` vs `custom hooks`? This could also be a hook at the router component level (maybe 🤔 ) that triggers the event each time the path changes.

jorgeepc (approved):
Looks good man!

web/public/index.html
xoscar: It will come from the backend at some point. So for the moment, it will be disabled. But yes someone could open the console and look at it and that's fine

web/src/components/WithAnalytics/WithAnalytics.tsx
xoscar: I'm on the fence in this one, I think we cannot do it at the router component as it would need to be wrapped by a `<BrowserRouter>` or `<Routes>` component to use the hooks to get the path.

Another thing is that we'll need to create a mapping to parse the paths to page names.

The HOC approach seems more explicit to me, but I agree that it might be overkill. I'm not sure 🤷🏽‍♀️

xoscar:
Not resolving the conflicts now as we need to wait for the rest of the segment setup to be ready and added the `do not merge` label

schoren (approved):
LGTM

web/src/services/Analytics/Analytics.service.ts
```diff
@@ -1,23 +1,36 @@
...
+    });
+  },
+  page(name: string) {
+    if (!isEnabled) return;
+    analytics.page(name);
+  },
+  identify() {
+    if (!isEnabled) return;
+    analytics.identify({
+      serverId,
```
schoren, line 29: Can we make sure the casing is the same as in the server? We are using `serverID`

jorgeepc: What about `appVersion` and `env` ? Should them be `AppVersion` and `Env`? @schoren

schoren: no, those are fine. It's just the `D`

Issue #914, opened 2022-07-14
Long trace attributes name are chopped


schoren:
**Describe the bug**
Trace attribute name longer than 26 characters shows chopped to that length.

**Screenshots**

## Span attribute list
![Screen Shot 2022-07-14 at 12 08 20](https://user-images.githubusercontent.com/314548/179015494-1a7fd84f-57ce-4197-a9dc-2087566cb963.png)

## Add assertion dropdown
![Screen Shot 2022-07-14 at 12 11 02](https://user-images.githubusercontent.com/314548/179015825-f61d09c4-96a1-473e-8f40-1e311b8e7e82.png)

PR #936, opened 2022-07-19
Long trace attributes

cescoferraro:
This PR deals with attributes name being trucated

## Changes

- increase AttributeCheck grid column size
- Add tooltip to attribute table if key is longer than 26 characters

## Fixes

- #914 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/src/components/AttributeRow/AttributeTitle.tsx
+++ b/web/src/components/AttributeRow/AttributeTitle.tsx
@@ -1,0 +1,25 @@
+import {Tooltip} from 'antd';
+import Highlighted from '../Highlighted';
+import {Text, TextContainer} from './AttributeRow.styled';
+
+interface IProps {
+  title: string;
+  searchText: string;
+}
+
+export default ({searchText, title}: IProps): JSX.Element => {
+  const textContainer = (
+    <TextContainer>
+      <Text type="secondary">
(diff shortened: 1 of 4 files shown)


web/src/components/AttributeRow/AttributeTitle.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export default ({searchText, title}: IProps): JSX.Element => {
+  const textContainer = (
+    <TextContainer>
+      <Text type="secondary">
+        <Highlighted text={title} highlight={searchText} />
+      </Text>
+    </TextContainer>
+  );
+  return title.length > 26 ? (
```
xoscar, line 18: This seems to be very specific, isn't there a way to deal with this using CSS?

cescoferraro: this is because with out current grid column width & font size, when the title kets bigger than 26 character, it gets truncated. css wont work. maybe I can figure out a js solution to infer the column width, but not sure if its worth it

kdhamric:
What is the status if this and #914? What do we need to do to finish these?

xoscar (approved):
I still think the solution is very coupled but I guess that's good for the moment

PR #947, opened 2022-07-20
feat: send triggering transaction span to application exporter

mathnogueira:
This PR makes tracetest send the triggering transaction to the application tracing storage.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/poller_executor.go
```diff
@@ -113,7 +121,14 @@ func (pe DefaultPollerExecutor) donePollingTraces(job *PollingRequest, trace tra
...
 
-	if len(trace.Flat) > 0 && len(trace.Flat) == len(job.run.Trace.Flat) {
+	// We always expect to get the tracetest trigger span, so it has to have more than 1 span
+	if len(trace.Flat) > 1 && len(trace.Flat) == len(job.run.Trace.Flat) {
+		pe.repeatedTraceSizeCache[trace.ID.String()] = pe.repeatedTraceSizeCache[trace.ID.String()] + 1
+	} else {
+		pe.repeatedTraceSizeCache[trace.ID.String()] = 0
+	}
+
+	if pe.repeatedTraceSizeCache[trace.ID.String()] >= 3 {
```
mathnogueira, line 131: This way, we only return true if the number of spans doesn't change after 3 pollings

mathnogueira: Decided to remove it from this PR. This solution is too raw to be merged, I'll work on a better approach later.

server/executor/poller_executor.go
```diff
@@ -73,7 +73,9 @@ func NewPollerExecutor(
 
 func (pe DefaultPollerExecutor) ExecuteRequest(request *PollingRequest) (bool, model.Run, error) {
 	run := request.run
-	otelTrace, err := pe.traceDB.GetTraceByID(request.ctx, run.TraceID.String())
+	traceId := run.TraceID.String()
```
schoren, line 76: nit: go conventions suggest this should be named `traceID`

server/executor/trigger/grpc.go
```diff
@@ -49,15 +57,22 @@ func (te *grpcTriggerer) Trigger(_ context.Context, test model.Test, tid trace.T
 	if trigger.GRPC == nil {
 		return response, fmt.Errorf("no settings provided for GRPC triggerer")
 	}
+
+	ctx, span := te.tracer.Start(context.Background(), "Tracetest Trigger")
```
schoren, line 61: I think it would be better to have the span and ctx set outside the trigger, so each trigger doesn't have to copy/paste this code

server/executor/trigger/http.go
```diff
@@ -40,20 +49,25 @@ func (te *httpTriggerer) Trigger(_ context.Context, test model.Test, tid trace.T
 
 	client := http.Client{
 		Transport: otelhttp.NewTransport(http.DefaultTransport,
-			otelhttp.WithTracerProvider(te.traceProvider),
+			otelhttp.WithTracerProvider(noopTracerProvider()),
 			otelhttp.WithPropagators(propagators()),
 		),
 	}
 
+	ctx, span := te.tracer.Start(ctx, "Tracetest Trigger")
```
schoren, line 57: Same here

server/model/json.go
```diff
@@ -126,9 +134,15 @@ func (r Run) MarshalJSON() ([]byte, error) {
...
 		TestVersion:               r.TestVersion,
 		Trigger:                   r.Trigger,
-		TriggerResult:             r.TriggerResult,
 		Trace:                     r.Trace,
 		Results:                   r.Results,
+		TriggerResult: encodedTriggerResult{
+			Type:    r.TriggerResult.Type,
+			HTTP:    r.TriggerResult.HTTP,
+			GRPC:    r.TriggerResult.GRPC,
+			SpanID:  r.TriggerResult.SpanID.String(),
```
schoren, line 143: Why do we need to move the spanID/TraceID from the Run to the trigger?

schoren (changes requested):

server/model/json.go
mathnogueira: We either do this or we pass the pointer to the run so we can set the spanID and traceID to the run instance.

server/executor/trigger/triggerer.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 type Triggerer interface {
-	Trigger(context.Context, model.Test, trace.TraceID, trace.SpanID) (Response, error)
+	Trigger(context.Context, context.Context, model.Test, trace.TraceID, trace.SpanID) (Response, error)
```
schoren, line 14: It's really confusing to have 2 contexts here, so it would be great to name these params so we can know what is which

server/executor/trigger/http.go
```diff
@@ -10,23 +10,17 @@ import (
...
-
-type httpTriggerer struct {
-	traceProvider *sdktrace.TracerProvider
+	return &httpTriggerer{}
 }
 
-func (te *httpTriggerer) Trigger(_ context.Context, test model.Test, tid trace.TraceID, sid trace.SpanID) (Response, error) {
+type httpTriggerer struct{}
 
+func (te *httpTriggerer) Trigger(_ context.Context, ctx context.Context, test model.Test, tid trace.TraceID, sid trace.SpanID) (Response, error) {
```
schoren, line 23: this looks like the first context should be called `ctx` and the 2nd `triggerSpanCtx` or something like that

server/executor/trigger/grpc.go
```diff
@@ -24,17 +22,12 @@ import (
...
-
-type grpcTriggerer struct {
-	traceProvider *sdktrace.TracerProvider
+	return &grpcTriggerer{}
 }
 
-func (te *grpcTriggerer) Trigger(_ context.Context, test model.Test, tid trace.TraceID, sid trace.SpanID) (Response, error) {
+type grpcTriggerer struct{}
 
+func (te *grpcTriggerer) Trigger(_ context.Context, ctx context.Context, test model.Test, tid trace.TraceID, sid trace.SpanID) (Response, error) {
```
schoren, line 30: this looks like the first context should be called `ctx` and the 2nd `triggerSpanCtx` or something like that

server/executor/trigger/grpc.go
```diff
@@ -93,6 +88,8 @@ func (te *grpcTriggerer) Trigger(_ context.Context, test model.Test, tid trace.T
 		marshaller: marshaler,
 	}
 
+	propagators().Inject(ctx, NewGRPCHeaderCarrier(&tReq.Metadata))
```
schoren, line 91: why do you need a custom carrier? it's probablly easier to do something like:

```
// see https://pkg.go.dev/go.opentelemetry.io/otel/propagation#MapCarrier

propagators().Inject(ctx, propagators.MapCarrier(tReq.Metadata.Map()))
// where tReq.Metadata.Map() is a func that returns a map of key->values from the metadata

```

server/executor/trigger/grpc_carrier.go
```diff
@@ -0,0 +1,54 @@
+package trigger
+
+import (
+	"github.com/kubeshop/tracetest/server/model"
+	"go.opentelemetry.io/otel/propagation"
+)
+
+type GRPCHeaderCarrier struct {
```
schoren, line 8: I would remove this entire implementation and just use https://pkg.go.dev/go.opentelemetry.io/otel/propagation#MapCarrier

server/executor/runner.go
```diff
@@ -124,13 +127,19 @@ func (r persistentRunner) processExecQueue(job execReq) {
 		panic(err)
 	}
 
-	response, err := triggerer.Trigger(job.ctx, job.test, job.run.TraceID, job.run.SpanID)
+	triggerSpanCtx, span := r.triggerSpanTracer.Start(context.Background(), "Tracetest trigger")
```
schoren, line 130: If you move this to the `instrumentedTriggerer.Trigger` func, you wouldn't need to use 2 contexts in the `Trigger` interface. Given that no trigger type is actually using both contexts, I understand that this is passed just because the`instrumentedTriggerer` doesn't have the `triggerSpanTracer` dependency. I'd rather add a dependency to the `instrumentedTriggerer` than to have 2 contexts on an interface that aren't actually required

schoren (changes requested):

server/executor/runner.go
mathnogueira: Not only that, to do that we either have to:

* Add a pointer to the test run to the Trigger interface
* Change the trigger response object to have both traceID and spanID

Without either of those, we can't update the run object with the correct traceID and spanID.

schoren (approved):

PR #981, opened 2022-07-24
Add missing slash

agardnerIT:
Add missing backslash which fixes helm install command

## Changes

- Adds backslash to `helm upgrade --install` command

## Fixes

- Fixes installer

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
Good catch. Thanks for the contribution!

Issue #974, opened 2022-07-22
Need ability to select all the spans in a trace when using the advanced selector


kdhamric:
I want to be able to select every span in a trace.  The backend query language supports this by have an empty selector.

We need to allow the front end to accept a blank filter when in the advanced mode.  Currently, it does, but it does not allow you to 'Add' 
Use case: I want to set an assertion saying 'there should be 4 spans in this trace' so I know if the trace changes.  To do so, I would select all spans and use our lovely tracetest.selected_spans.count.  

Here is a recording: https://www.loom.com/share/91c041f2b60741e7b0612be3ed294487

PR #990, opened 2022-07-25
Fix selected assertion

jorgeepc:
With the changed introduce in #974, we were having an incorrect behavior in the `select assertion` logic when you add an empty selector (to select all spans). 

The problem is caused because we use the `selector` as an identifier of the assertion, but now (having it as empty) it was collapsing with the logic for a `non-selected` assertion.

## Changes

- Change the `selectedAssertion` state to be `undefined` if no assertion is selected.

## Fixes

- #974 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


xoscar (approved):
Great catch @jorgeepc I didn't realize that could happen

Issue #782, opened 2022-06-21
support expressions in assertions


mathnogueira:
# Problem
Today we only can execute assertions against static values. However, there are some use-cases for more complex assertions, such as asserting that an operation that is async runs in under 2 minutes. Or that your wallet balance is less than it was before the purchase. 

## Examples

```
myapp.user.balance < myapp.user.balance_before_purchase
```

```
span.endTime < tracetest.execution.startTime + 2m
```

## Design

To make it fast, I think this should not allow using attributes from other spans. So, you can only reference attributes of the span that is being validated at that moment.

mathnogueira:
This was introduced in v0.8. Closing this issue.

PR #1007, opened 2022-07-27
feat: attribute referencing and expression support

mathnogueira:
This PR allows users to specify assertions using variables from the selected span and to use basic arithmetic expressions with it. 

Examples:
```yaml
testDefinition:
  selector: span[name = "proxy"]
  assertions:
    - http.status_code = http.proxy.status_code # referencing an attribute from the span
    - tracetest.span.endtime < tracetest.span.startTime + 800ms # same as duration < 800ms
```

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


mathnogueira:
Related to #782

schoren:
I'm using https://github.com/pterm/pterm for CLI UI for the installer. I think we should use the same lib for consistency

server/assertions/operation.go
```diff
@@ -0,0 +1,156 @@
...
+func validateFieldType(field model.LiteralValue) error {
+	if field.Type != "number" && field.Type != "duration" {
+		return fmt.Errorf("operation is only allowed on numbers and duration fields")
+	}
+
+	return nil
+}
+
+func getOperationExecutor(operation string) (ExpressionOperation, error) {
+	switch operation {
```
schoren, line 102: This switch looks weird. I would either return a basic func that does the operation from here, or have a "registry" approach for extensibility. Otherwise, there's a lot of code overhead to end up with a small set of not-changing options

schoren (changes requested):
left a commetn

mathnogueira:
@schoren applied changes.

schoren (approved):

PR #1012, opened 2022-07-28
Fix setup script

schoren:
This PR fixes the setup script when calling it remotely by removing the dependency on the local `k8s/deploy-collector.sh` script.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira (approved):
LGTM

Issue #948, opened 2022-07-20
Add variables to test triggers


mrwatson:
**Is your feature request related to a problem? Please describe.**
Our potential use case involves sending API calls that have varying data in the request body (i.e., randomly generated names or IDs) because the call will fail if run twice with the same variables in the request unless the database is somehow cleaned up in between calls.

**Describe the solution you'd like**
Provide a means to include variables in the request that would be dynamically replaced at run-time. The variables could be specified by the user when the test is triggered or could be generated dynamically by the test itself (such as generating a UUID - Insomnia has a similar feature, for comparison).

**Describe alternatives you've considered**
Dynamically generating the entire test as a YAML file. This would work, but is significantly more effort.

**Additional context**
N/A

mrwatson:
Currently, I only see value for data generation on the triggering side. On the assertion side, it would be nice to be able to test against variables, though (see #950).

mathnogueira:
Yes, currently we only support data generation on the triggering side. I'm planning to start working on adding support to data generation probably on Monday.

mathnogueira:
About supporting variables, we have the goal of adding support for both variables and expressions: #782

kdhamric:
@mathnogueira one concept I believe we will want to add is 'adhoc' running of transactions. Think of it as manual testing - I am developing and need to exercise the api and see oth the response and the trace output. We will want to be able to save these and run them when needed. It will make sense to allow a variable to be passed in or set... so variables may end up helping with that scenario

mathnogueira:
Hey @mrwatson I just released `v0.6.4` and it introduces the concept of generator functions which are functions that generate random data in your test trigger section.

You can read more about it in the `generator functions` section of in [this doc page](https://kubeshop.github.io/tracetest/test-definition-file/).

mathnogueira:
@mrwatson if you feel this doesn't solve your problem. Feel free to reopen this issue.

mrwatson:
Okay, thanks guys. We'll take a look when we get back around to trying this again.

PR #1015, opened 2022-07-28
feat: generator functions

mathnogueira:
This PR adds the capability of using generator functions in your tests. 

Now you can use `{{ function(args) }}` in your test definition to generate the value on the fly.

TODO: Add documentation

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/executor/functions/function_registry.go
+++ b/server/executor/functions/function_registry.go
@@ -1,6 +1,9 @@
 package functions
 
-import "fmt"
+import (
+	"fmt"
+	"sync"
+)
 
 type Registry interface {
 	Add(string, Invoker, ArgConfig)
@@ -15,9 +18,13 @@ func newRegistry() Registry {
 
 type registry struct {
 	functions map[string]Function
+	mutex     sync.Mutex
 }
 
 func (r *registry) Add(name string, function Invoker, argsConfig ArgConfig) {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
 	r.functions[name] = Function{
 		name:      name,
 		invoker:   function,
--- a/server/executor/functions/function_registry.go
+++ b/server/executor/functions/function_registry.go
@@ -6,7 +6,7 @@ import (
 )
 
 type Registry interface {
-	Add(string, Invoker, ArgConfig)
+	Add(string, Invoker, ArgTypes)
 	Get(string) (Function, error)
 }
 
@@ -21,14 +21,14 @@ type registry struct {
 	mutex     sync.Mutex
 }
 
-func (r *registry) Add(name string, function Invoker, argsConfig ArgConfig) {
+func (r *registry) Add(name string, function Invoker, argsConfig ArgTypes) {
 	r.mutex.Lock()
 	defer r.mutex.Unlock()
 
 	r.functions[name] = Function{
-		name:      name,
-		invoker:   function,
-		argConfig: argsConfig,
+		name:     name,
+		invoker:  function,
+		argTypes: argsConfig,
 	}
 }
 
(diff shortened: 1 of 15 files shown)


mathnogueira:
This implements the feature request described in #948

server/executor/functions/function_registry.go
```diff
@@ -0,0 +1,34 @@
+package functions
+
+import "fmt"
+
+type FunctionRegistry interface {
```
schoren, line 5: You don't need to repeat the word `Function`, it's already in the package: https://michaelwhatcott.com/go-code-that-stutters/

server/executor/functions/functions.go
```diff
@@ -0,0 +1,59 @@
+package functions
+
+import "fmt"
+
+type FunctionInvoker func(args ...FunctionArg) string
```
schoren, line 5: Same for everything. It's a lot easier to read and write `functions.Arg` than `functions.FunctionArg`

server/executor/functions/functions.go
```diff
@@ -0,0 +1,59 @@
...
+type FunctionArg struct {
+	Value string
+	Type  string
+}
+
+type FunctionArgConfig struct {
```
schoren, line 11: This could be written as:
````
type ArgConfig struct {
	Count int
  Types   []string
}
```

server/executor/functions/parser.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not create parser: %w", err)
+	}
+
+	defaultParser = parser
+
+	return defaultParser, nil
+}
+
+func ParseFunction(input string) (Function, []FunctionArg, error) {
```
schoren, line 88:
```suggestion
func Parse(input string) (Function, []Arg, error) {
```

server/executor/replacer/replacer.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"fmt"
+
+	"github.com/kubeshop/tracetest/server/model"
+)
+
+var defaultInjector *Injector = nil
+
+func ReplaceTestPlaceholders(test model.Test) (model.Test, error) {
+	if defaultInjector == nil {
```
schoren, line 12: I'd move this to an `init` func, or directly to the var init, like:
```
var defaultInjector = NewInjector()
```

schoren (changes requested):

server/executor/replacer/replacer.go
```diff
@@ -0,0 +1,25 @@
+package replacer
+
+import (
+	"fmt"
+
+	"github.com/kubeshop/tracetest/server/model"
+)
+
+var defaultInjector *Injector = nil
```
schoren, line 9: Do we really need this to be a reference? It can be made a direct instance:

```suggestion
var defaultInjector Injector = NewInjector()
```

schoren (changes requested):
Looking great! there are a few details that needs attention, but the implementation is awesome

server/executor/functions/functions.go
```diff
@@ -0,0 +1,59 @@
...
+import "fmt"
+
+type Invoker func(args ...Arg) string
+type Arg struct {
+	Value string
+	Type  string
+}
+
+type ArgConfig struct {
+	Count int
```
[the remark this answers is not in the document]
mathnogueira: Yes it can :facepalm:

server/executor/functions/functions.go
```diff
@@ -0,0 +1,56 @@
...
+
+import "fmt"
+
+type Invoker func(args ...Arg) string
+type Arg struct {
+	Value string
+	Type  string
+}
+
+type ArgConfig struct {
```
schoren, line 11: Suggestion: since this is now a single field type, can it be converted to an alias?

```suggestion
type ArgTypes []string
```

schoren (approved):
Left a suggestion based on personal preference, so feel free to ignore it. otherwise LGTM!

PR #1016, opened 2022-07-28
Upgrade ansi-regex to 4.1.1

iblancasa:
Signed-off-by: Israel Blancas <iblancasa@gmail.com>

## Changes

- Upgrade `ansi-regex` dependency from `4.1.0` to `4.1.1`.

## Fixes

- [CVE-2021-3807](https://nvd.nist.gov/vuln/detail/CVE-2021-3807)

## Checklist

- [x] tested locally

# Description
The project is using ansi-regex 4.1.0. It seems to be a transitive dependency. That dependency has a vulnerability ([CVE-2021-3807](https://nvd.nist.gov/vuln/detail/CVE-2021-3807)). Upgrading it to 4.1.1, solves the problem

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -46798,9 +46798,9 @@
           }
         },
         "ansi-regex": {
-          "version": "4.1.0",
-          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-4.1.0.tgz",
-          "integrity": "sha512-1apePfXM1UOSqw0o9IiFAovVz9M5S1Dg+4TrDwfMewQ6p/rmMueb7tWZjQ1rx4Loy1ArBggoqGpfqqdI4rondg=="
+          "version": "4.1.1",
+          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-4.1.1.tgz",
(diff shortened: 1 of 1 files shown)


xoscar:
Hey @iblancasa thank you for contributing.

Do you mind providing more information about this change?

Thanks!

cescoferraro:
Hi @iblancasa, very nice to hear you are interested in the project. 🥇 

But this `ansi-regex` is a dependency of lots of dependencies we use right now.
Screen Shot 2022-07-31 at 11 14 37
So I think we need to wait for them to update theirs, since we cannot ditch them entirely.
I found this [#18713](https://github.com/storybookjs/storybook/issues/18713) , where @storybook is planning to solve this issue on the next release. But not sure about the other ones. Latest `eslint` version also does not includes this solution.
Good thing is that all those dependencies that use `ansi-regex` are only used at build time at Tracetest, so this should not affect our end users.

If there is anything else you see missing or would like to see on Tracetest, please do not hesitate in contacting us.
We are actively available on the [Discord channel](https://discord.com/invite/eBvEQRVyKX)

iblancasa:
Hi @cescoferraro, I see…
How about adding `ansi-regex` 4.1.1 to the dependencies in the meantime?

cescoferraro:
@iblancasa good idea. Although it would solve @types/jest version, cause the newly installed version would dedup ansi-regex@4.1.0
This would not solve the issue on @storybook/react that would still use version 2.1.1. Which is also compromised.
Screen Shot 2022-08-01 at 09 21 49
Screen Shot 2022-08-01 at 09 19 33

Issue #983, opened 2022-07-24
Timeline view should also have a '0 of X' total and ability to step through selected spans


kdhamric:
We should allow users to step through the selected spans in both the timeline and the system view.

PR #1018, opened 2022-07-29
add timeline step through

cescoferraro:
This PR implements #983 

## Changes

- Add Controls components to Timeline View
- Scroll timeline elements when stepping though affected spans
- add listener so user can step though the diagram using arrow keys

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/components/Diagram/components/DAG/Controls.tsx
+++ b/web/src/components/Diagram/components/DAG/Controls.tsx
@@ -1,27 +1,36 @@
 import {ApartmentOutlined, ExpandOutlined, ZoomInOutlined, ZoomOutOutlined} from '@ant-design/icons';
+import {noop} from 'lodash';
 import {useReactFlow} from 'react-flow-renderer';
-import * as S from './DAG.styled';
 import AffectedSpanControls from './AffectedSpanControls';
+import * as S from './DAG.styled';
+import {ControlsMode} from './DAG.styled';
 
 interface IProps {
-  isMiniMapActive: boolean;
-  onMiniMapToggle(): void;
+  isMiniMapActive?: boolean;
+  mode?: ControlsMode;
+  onMiniMapToggle?: () => void;
 }
 
-const Controls = ({isMiniMapActive, onMiniMapToggle}: IProps) => {
+const Controls = ({mode = 'dag', isMiniMapActive = false, onMiniMapToggle = noop}: IProps) => {
   const {fitView, zoomIn, zoomOut} = useReactFlow();
-
   return (
     <>
-      <S.SelectorControls>
+      <S.SelectorControls mode={mode}>
         <AffectedSpanControls />
       </S.SelectorControls>
-      <S.Controls>
-        <S.ZoomButton icon={<ZoomInOutlined />} onClick={() => zoomIn()} type="text" />
-        <S.ZoomButton icon={<ZoomOutOutlined />} onClick={() => zoomOut()} type="text" />
-        <S.ZoomButton icon={<ExpandOutlined />} onClick={() => fitView()} type="text" />
-        <S.ZoomButton icon={<ApartmentOutlined />} onClick={onMiniMapToggle} type="text" $isActive={isMiniMapActive} />
-      </S.Controls>
+      {mode === 'dag' && (
+        <S.Controls>
+          <S.ZoomButton icon={<ZoomInOutlined />} onClick={() => zoomIn()} type="text" />
+          <S.ZoomButton icon={<ZoomOutOutlined />} onClick={() => zoomOut()} type="text" />
+          <S.ZoomButton icon={<ExpandOutlined />} onClick={() => fitView()} type="text" />
+          <S.ZoomButton
+            icon={<ApartmentOutlined />}
+            onClick={onMiniMapToggle}
+            type="text"
+            $isActive={isMiniMapActive}
+          />
+        </S.Controls>
+      )}
     </>
   );
 };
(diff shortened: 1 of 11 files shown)


cescoferraro:
https://www.loom.com/share/14600aa495824aebb1d2a5d232d740f6

web/src/components/Diagram/components/DAG/Controls.tsx
```diff
@@ -5,23 +5,31 @@ import AffectedSpanControls from './AffectedSpanControls';
...
+      <S.SelectorControls $isTimeline={isTimeline}>
         <AffectedSpanControls />
       </S.SelectorControls>
-      <S.Controls>
-        <S.ZoomButton icon={<ZoomInOutlined />} onClick={() => zoomIn()} type="text" />
-        <S.ZoomButton icon={<ZoomOutOutlined />} onClick={() => zoomOut()} type="text" />
-        <S.ZoomButton icon={<ExpandOutlined />} onClick={() => fitView()} type="text" />
-        <S.ZoomButton icon={<ApartmentOutlined />} onClick={onMiniMapToggle} type="text" $isActive={isMiniMapActive} />
-      </S.Controls>
+      {!isTimeline && (
```
xoscar, line 20: This is a small thing but I think is preferable to evaluate positive cases like isDag instead of negative ones. We can have that flag be true as default

web/src/components/Diagram/components/DAG/useListenToArrowKeysDown.tsx
```diff
@@ -0,0 +1,35 @@
+import {useEffect} from 'react';
+
+export function useListenToArrowKeysDown(
+  affectedSpans: string[],
+  handleNextSpan: () => void,
+  handlePrevSpan: () => void
+): void {
+  useEffect(() => {
+    function downHandler({key}: KeyboardEvent) {
```
xoscar, line 9: This is a great improvement, but I'm a little concerned about going out of React to achieve this,  what if we instead use refs for at least the ones we can?

web/src/components/Diagram/components/Timeline/Timeline.tsx
```diff
@@ -1,13 +1,18 @@
...
+import Controls from '../DAG/Controls';
 
 import Visualization from './Visualization';
 
 const Timeline = (props: IDiagramComponentProps) => (
-  <ParentSize parentSizeStyles={{height: 'auto', overflowY: 'scroll'}}>
-    {({width}) => <Visualization {...props} width={width} />}
-  </ParentSize>
+  <>
+    <Controls isMiniMapActive={false} onMiniMapToggle={noop} isTimeline />
```
xoscar, line 11: What do you think about having default for the onMiniMapToggle and isMiniMapActive props as well

web/src/components/Diagram/components/Timeline/useScrollSpanNodeGroupIntoView.tsx
```diff
@@ -0,0 +1,25 @@
...
+import {useSpan} from '../../../../providers/Span/Span.provider';
+import useControls from '../DAG/hooks/useControls';
+
+interface IProps {
+  isSelected?: boolean;
+  id: string;
+  index: number;
+}
+
+export function useScrollSpanNodeGroupIntoView({index, isSelected, id}: IProps): void {
```
xoscar, line 11: Do you mind telling me what's the idea of this hook?

web/src/providers/Span/Span.provider.tsx
```diff
@@ -47,7 +47,7 @@ interface IProps {
   children: React.ReactNode;
 }
 
-export const useSpan = () => useContext(Context);
+export const useSpan = (): IContext => useContext(Context);
```
xoscar, line 50: I think the returned type is already deducted by the editor, no need to have it be explicit

xoscar (approved):
Hey @cescoferraro great job man, I left some comments and concerns. Let me know what you think

web/src/providers/Span/Span.provider.tsx
cescoferraro: i dont think so

web/src/components/Diagram/components/DAG/Controls.tsx
cescoferraro: ok! nice

web/src/components/Diagram/components/DAG/useListenToArrowKeysDown.tsx
cescoferraro: I think using the ref would be a performance issue. especially if the dag is really big.
that's I I have implemented outside react

web/src/components/Diagram/components/Timeline/Timeline.tsx
cescoferraro: ok

web/src/components/Diagram/components/Timeline/useScrollSpanNodeGroupIntoView.tsx
cescoferraro: it makes sure the timeline bar is visible when the user is stepping through the affected span

web/src/providers/Span/Span.provider.tsx
cescoferraro: we could all do useContext<IContext>(Context) but it boils down to the same thing

PR #1021, opened 2022-07-29
feat: lightstep support

mathnogueira:
This PR adds support to lightstep as a tracing backend

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -104,9 +104,9 @@ services:
       JAEGER_PORT: 6832
       NPM_RUN_COMMAND: api
     ports:
-      - "8081:8081"
+      - "8081:80"
     healthcheck:
-      test: ["CMD", "wget", "--spider", "localhost:8081"]
+      test: ["CMD", "wget", "--spider", "localhost/pokemon/healthcheck"]
       interval: 1s
       timeout: 3s
       retries: 60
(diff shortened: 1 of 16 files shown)


mathnogueira:
This will not be implemented for now

kdhamric:
Wanted to leave a longer description of why we stopped work on this - here was the discussion in slack:
Hey team, I have an update about lightstep. I managed to make it work, however, I found a limitation on their product. You can only get "Assembled traces", this means that you have to execute an operation to assemble the trace before you can actually get it from the API. The thing is, they have a pretty limiting rate-limit rate (60 calls per hour in the most expensive plan). This means that it will be almost impossible to run multiple tests using Lightstep as a backend. We talked with Austin Parker about this issue and he told us that they are going to implement a new endpoint that might solve our problem but it will also have rate-limits in it. So, after some discussion we figured out that the best of way to support lightstep is to actually use the OTel collector to send the traces directly to tracetest. So we will not implement lightstep support directly into tracetest.
Next steps are: take a look at the other tracing backends that we wanted to support and check if it's possible to support them with the existing API or if we should implement the collector solution from start.

We will enable Lightstep with the project to enable the capturing of traces by adding a second pipeline in the Otel Collector - #1071

PR #1026, opened 2022-08-03
add search for the test list at frontend home

cescoferraro:
This PR implements search feature on the home test list

## Changes

- add query parameter to Get test list
- adds frontend necessary code

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -26,6 +26,11 @@ paths:
           schema:
             type: integer
             default: 0
+        - in: query
+          name: query
+          description: "query to search tests"
+          schema:
+            type: string
       responses:
         200:
           description: successful operation
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -28,7 +28,7 @@ paths:
             default: 0
         - in: query
           name: query
-          description: "query to search tests"
+          description: "query to search tests, based on test name and description"
           schema:
             type: string
       responses:
--- a/server/openapi/api_api.go
+++ b/server/openapi/api_api.go
@@ -452,7 +452,8 @@ func (c *ApiApiController) GetTests(w http.ResponseWriter, r *http.Request) {
 		c.errorHandler(w, r, &ParsingError{Err: err}, nil)
 		return
 	}
-	result, err := c.service.GetTests(r.Context(), takeParam, skipParam, query.Get("query"))
+	queryParam := query.Get("query")
+	result, err := c.service.GetTests(r.Context(), takeParam, skipParam, queryParam)
 	// If an error occurred, encode the error with the status code
 	if err != nil {
 		c.errorHandler(w, r, err, &result)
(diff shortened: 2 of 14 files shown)


server/openapi/api_api.go
```diff
@@ -452,7 +452,7 @@ func (c *ApiApiController) GetTests(w http.ResponseWriter, r *http.Request) {
 		c.errorHandler(w, r, &ParsingError{Err: err}, nil)
 		return
 	}
-	result, err := c.service.GetTests(r.Context(), takeParam, skipParam)
+	result, err := c.service.GetTests(r.Context(), takeParam, skipParam, query.Get("query"))
```
schoren, line 455: this is an autogenerated file. Please use `make generate` from the server

api/openapi.yaml
```diff
@@ -26,6 +26,11 @@ paths:
           schema:
             type: integer
             default: 0
+        - in: query
+          name: query
+          description: "query to search tests"
```
jorgeepc, line 31: Minor: maybe it's a good idea to specify in which attributes the search is based on. It would be great to have the search using the `name`, `description` and `url` of the test.

web/src/hooks/useInfiniteScroll.ts
```diff
@@ -6,10 +6,20 @@ type TUseInfiniteScrollParams<P> = P & {
...
 };
 
+export interface InfiniteScrollModel<T> {
+  isLoading: boolean;
+  localPage: number;
+  loadMore: () => void;
+  hasMore: boolean;
+  isFetching: boolean;
+  refresh: () => void;
+  list: [T[], React.Dispatch<React.SetStateAction<T[]>>][0];
```
jorgeepc, line 16: is this the same as:
```ts
list: T[];
```

jorgeepc (commented):
Great job @cescoferraro Added a couple of comments.

api/openapi.yaml
cescoferraro: this is good idea!

web/src/hooks/useInfiniteScroll.ts
cescoferraro: right! I have auto-generated this with Intellij.
@schoren is working on a couple of improvements for the backend
then I can change this

api/openapi.yaml
schoren: I've added description, but URL is very trigger specific, so I think for this basic implementation it should be enough

web/src/hooks/useInfiniteScroll.ts
schoren: @cescoferraro ready

jorgeepc (approved):

schoren (approved):

PR #1034, opened 2022-08-07
fix bug of step though when advanced selector is focused

cescoferraro:
This PR fixes a bug

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


xoscar:
What specific bug does this fix? Do you mind adding more info? I also think the solution is very coupled to the advanced editor id.

cescoferraro:
when you focused on the advanced editor input and use the arrows keys, the feature I have added before is stepping though the affected spans both on the DAG & timeline view @xoscar

xoscar (approved):
LGTM

PR #1061, opened 2022-08-12
add workflow to test installer

schoren:
This PR adds a GH actions workflow to ensure basic functionality of the CLI installer. Right now it only tests linux on amd64, since that's the only arch we have runners configured, but can be extended to other OSs/architectures by creating self-hosted runners

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/.github/workflows/pull-request-installer.yaml
+++ b/.github/workflows/pull-request-installer.yaml
@@ -41,5 +41,10 @@ jobs:
               ;;
           esac
         run: |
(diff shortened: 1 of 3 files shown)


mathnogueira (approved):
Nice

jorgeepc (approved):

kdhamric (approved):

PR #1062, opened 2022-08-14
add query to test detail page

cescoferraro:
This PR adds search capabilities to the test detail page which shows a list of test runs for a given test

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -217,6 +217,12 @@ paths:
           schema:
             type: integer
             default: 0
+        - in: query
+          name: query
+          description: "indicates a keyword that will filter the results"
+          schema:
+            type: string
+            default: ''
       summary: "get the runs for a test"
       description: "get the runs from a particular test"
       operationId: getTestRuns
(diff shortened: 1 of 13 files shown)


cescoferraro:
@kdhamric @schoren this is working, but I am not sure what would the search be based upon.
The run entity does not have much information in it for us to query. We can always `inner-join` the test entity. But then it would match all of the runs always. What should the query look for in this case

kdhamric:
Yes - searching is not very helpful here. Filtering, however, is useful. Show me the test runs with or without failures. Select test runs by a date range.  

The spot where searching may be valuable is in searching against the fields captured by #1066 - being able to quickly find the test run associated with a particular PR, etc will be helpful.

I would, however, question putting much work into this capability yet, particularly if we need backend team involvement. Bigger fish to fry that will impact initial engagement with the product.

cescoferraro:
@kdhamric the backend part is done as well. if we need #1066 fields to be able to filter the results, we need to wait for it.
I can close this one now, but when they exist, we can just reopen this because the code is the same. Just need to change the SQL script

kdhamric:
Lets come back to it once the #1066 is done. Normally I would say to put what you have in there... but I literally cannot think of one thing that would be useful to search for.  Can it handle any kind of filtering - on date ranges or passed/failed tests?

cescoferraro:
@kdhamric we can add the filtering!

PR #1074, opened 2022-08-15
cli: test run improvements

mathnogueira:
This PR improves the way our CLI outputs the test result after running a test.

Example:
Screenshot from 2022-08-15 17-41-32

## TODO
- [x] add link to test run in output

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test


schoren:
I'm using https://github.com/pterm/pterm for CLI UI for the installer. I think we should use the same lib for consistency

schoren (approved):
looking great!

PR #1083, opened 2022-08-16
add create test cancel button

### LINKED ISSUE #980: Cannot easily cancel out of a create test flow
When creating a test, there is not a clear way to cancel out of the flow if you choose to abandon / stop during the process. If you go back to the tracetest main screen from deep in the creation wizard, you are returned to your original spot.  The video shows it best:

https://www.loom.com/share/4e6b91f6322041f08a661c5a80ca2dfc

Not sure the best way to handle this - probably would like @olha23 thoughts. If the team is fine with it 'as is', I am ok with that also.

cescoferraro:
This PR fixes #980 

## Changes

- Add cancel button to test action footer

## Fixes

- #980

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


web/cypress/integration/Home/CreateTest.spec.ts
```diff
@@ -11,6 +11,12 @@ describe('Create test', () => {
     cy.visit('http://localhost:3000/');
   });
 
+  it.only('should cancel a create test flow', () => {
```
xoscar, line 14: Let's remove the `only`

web/src/components/CreateTestSteps/CreateTestStepFooter.tsx
```diff
@@ -10,28 +11,42 @@ interface IProps {
...
         <Button data-cy="create-test-prev-button" type="primary" ghost onClick={onPrev}>
           Previous
         </Button>
-      )}
-      {stepNumber < stepList.length - 1 ? (
-        <Button data-cy="create-test-next-button" disabled={!isValid} onClick={onNext} type="primary">
-          Next
-        </Button>
       ) : (
+        <div />
```
xoscar, line 22: This empty div looks weird, is there another way we can achieve the same functionality without having it?

xoscar (changes requested):
Hey @cescoferraro I left two small comments, let me know what you think

web/src/components/CreateTestSteps/CreateTestStepFooter.tsx
cescoferraro: fixed

web/cypress/integration/Home/CreateTest.spec.ts
cescoferraro: sure

xoscar (approved):

jorgeepc (approved):

kdhamric:
Awesome. Verified it in the beta area - thanks!

PR #1084, opened 2022-08-16
add secondary filter options by tags to assertion-check-attribute

### LINKED ISSUE #1053: Add descriptions/tags to attributes
**Objective:** Have a description or tags associated to an attribute name to improve the search attribute experience. Example, you should be able to type seconds or milliseconds or ms, and see that we have a `tracetest.span.duration` attribute.

**AC1:**
As a user adding checks in the create assertion form,
I should be able to search for an attribute based on the `description` or `tags` associated with it.

**Note:**
The description or tags should be hardcoded in the frontend for now, and they should match the [OTel semantic convention.](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions)

cescoferraro:
This PR adds secondary filter options by tags to assertion-check-attribute

## Changes

- implements #1053 

## Fixes
Fixes #1053 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/src/constants/SpanAttribute.constants.ts
+++ b/web/src/constants/SpanAttribute.constants.ts
@@ -8,10 +8,10 @@ export const TraceTestAttributes = {
   TRACETEST_RESPONSE_STATUS: 'tracetest.response.status',
   TRACETEST_RESPONSE_BODY: 'tracetest.response.body',
   TRACETEST_RESPONSE_HEADERS: 'tracetest.response.headers',
-  TRACETEST_SELECTED_SPANS_COUNT: 'tracetest.selected_spans.count'
+  TRACETEST_SELECTED_SPANS_COUNT: 'tracetest.selected_spans.count',
 };
 
-export const Attributes = {
+export const Attributes: Record<string, string> = {
   ...SemanticAttributes,
   ...SemanticResourceAttributes,
   ...TraceTestAttributes,
--- a/web/cypress/integration/Trace/CreateAssertion.spec.ts
+++ b/web/cypress/integration/Trace/CreateAssertion.spec.ts
@@ -30,10 +30,10 @@ describe('Create Assertion', () => {
 
     cy.get('[data-cy=add-assertion-form-add-check]').click();
 
-    cy.get('[data-cy=assertion-check-attribute]').last().type('duration');
-    cy.get(`${getAttributeListId(2)} + div .ant-select-item`)
-      .first()
-      .click();
+      cy.get('[data-cy=assertion-check-attribute]').last().type('time');
+      cy.get(`${getAttributeListId(2)} + div .ant-select-item`)
+        .first()
+        .click();
 
     cy.selectOperator(2);
 
--- a/web/cypress/support/commands.ts
+++ b/web/cypress/support/commands.ts
@@ -152,7 +152,7 @@ Cypress.Commands.add('editTestByTestId', (testId: string) => {
 });
 
 Cypress.Commands.add('selectOperator', (index: number, text?: string) => {
-  cy.get('[data-cy=assertion-check-operator]').last().click();
+  cy.get('[data-cy=assertion-check-operator]').last().click({force: true});
   cy.get(`${getComparatorListId(index)} + div .ant-select-item`)
     .last()
     .click();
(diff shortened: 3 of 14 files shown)


web/src/constants/SpanAttribute.constants.ts
```diff
@@ -8,10 +8,14 @@ export const TraceTestAttributes = {
...
   TRACETEST_RESPONSE_BODY: 'tracetest.response.body',
   TRACETEST_RESPONSE_HEADERS: 'tracetest.response.headers',
-  TRACETEST_SELECTED_SPANS_COUNT: 'tracetest.selected_spans.count'
+  TRACETEST_SELECTED_SPANS_COUNT: 'tracetest.selected_spans.count',
 };
 
-export const Attributes = {
+export const AttributesTags: Record<string, string[]> = {
+  TRACETEST_SPAN_DURATION: ['ms', 'time'],
+};
```
jorgeepc, line 16: Looks good man. We need to create more tags and descriptions for attributes following the OTel semantic convention https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions

xoscar: Yes, the idea is to grab each of description text and add show it as part of the attribute input in the create assertion form.

cescoferraro: OK. Will add those I find on the docs. But will leave a placeholder for all the one we go so far.

web/src/constants/AttributesTags.ts
```diff
@@ -0,0 +1,275 @@
...
+  // NET_HOST_CARRIER_MCC: {tags:[],description:""},
+  // NET_HOST_CARRIER_MNC: {tags:[],description:""},
+  // NET_HOST_CARRIER_NAME: {tags:[],description:""},
+  // NET_HOST_CONNECTION_SUBTYPE: {tags:[],description:""},
+  // NET_HOST_CONNECTION_TYPE: {tags:[],description:""},
+  // NET_HOST_IP: {tags:[],description:""},
+  // NET_HOST_NAME: {tags:[],description:""},
+  // NET_HOST_PORT: {tags:[],description:""},
+  // NET_PEER_IP: {tags:[],description:""},
+  NET_PEER_NAME: {description: 'Name of the database host.', tags: ['example.com']},
```
cescoferraro, line 226: @xoscar is that what you need?

xoscar: Yeah, something like this, I'd suggest having this as json files. After that we need to add the description to the attribute dropdown.

cescoferraro: @xoscar where in the attribute dropdown?

web/src/constants/SpanAttribute.constants.ts
kdhamric: Can we leverage the yaml files that contain these definitions, for example: https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/http.yaml

xoscar:
@cescoferraro I see that we have something like this with your changes:
Screen Shot 2022-08-23 at 11 01 17

I think this is quite difficult to read and doesn't add too much value. Maybe we can ask @olha23 to give us a hand around what would be the best way to show it.

olha23:
@xoscar @cescoferraro adding the name attribute and the meaning for attribute in dropdown seams wrong for me. If you think that users do not know the meanings of attributes, we can add the link for doc page with all explanations at the top of the input

olha23:
@kdhamric i am sure that if we add the description as a part of what is searched - this will not solve the problem, it will only confuse the user even more because it will be very overwhelmed. The other way how we can help them - maybe think about implementing smart search. User types some characters inside the search box, the smart search function will find all the best matching attributes from our inner database. And like when user type "time" - instead 'No data' message show "Maybe you mean Duration". Or something like this 😂

cescoferraro:
@kdhamric  the PR is already searching it by three elements
- its key
- its description
- its tags.

Screen Shot 2022-08-24 at 11 47 16

The issue here is whether should we concatenate the description after the attribute key. Just like this screenshot Oscar posted before

Screen Shot 2022-08-23 at 11 01 17

kdhamric:
Perhaps we remove the 'dash description' from the dropdown list you are showing and just show the attribute name. On hover over an attribute name, show the attribute name - description just like you are. This may clean it up a bit. I would like to see that in an environment where we can play with it (beta?) to see how it feels.

cescoferraro:
@kdhamric  I removed the dash descriptor thing. But the tooltip just show the text written on the select option. It is something done automatically by antd. Let me try to keep the description on the tooltip. But anyhow, the PR is ready to be reviewed/approved  @xoscar @jorgeepc

xoscar (approved):
LGTM

PR #1087, opened 2022-08-17
feat: support opensearch as a tracing backend

mathnogueira:
This PR adds support for opensearch as a tracing backend.

## Checklist

- [x] tested locally
- [x] added new dependencies
- [x] updated the docs
- [ ] added a test


server/traces/traces.go
```diff
@@ -16,6 +16,50 @@ type Trace struct {
 	Flat     map[trace.SpanID]*Span `json:"-"`
 }
 
+func NewTrace(traceID string, spans ...Span) Trace {
```
mathnogueira, line 19: I'll add a card to our backlog so we use this method instead of building the trace manually when converting from the otel format.

mathnogueira:
I deleted the existing tracing backend tests because:

1. They were being skipped anyway
2. They didn't test anything, I rather think about another strategy for testing those (maybe running all tests using different data storages?)

server/executor/poller_executor.go
```diff
@@ -115,8 +114,7 @@ func (pe DefaultPollerExecutor) donePollingTraces(job *PollingRequest, trace tra
 		return false
 	}
 
-	// We always expect to get the tracetest trigger span, so it has to have more than 1 span
```
mathnogueira, line 118: This is only true if the user configures tracetest to send the triggering span to the same data store as its application. So, to prevent infinite loops, I reverted the `len() > 1` condition-

server/traces/traces.go
schoren, line 19: This can be named `New`, so the calls look like `traces.New(...)`

schoren: Also, it looks easier to pass a `[]Span` than to expand an array on every call. I wouldn't expect it to be called with a manual list of `Span`s

server/traces/traces.go
```diff
@@ -16,6 +16,50 @@ type Trace struct {
...
+	return trace
+}
+
+func augmentSpanData(span *Span) {
+	span.Attributes["name"] = span.Name
+	span.Attributes["tracetest.span.type"] = spanType(span.Attributes)
+	span.Attributes["tracetest.span.duration"] = getSpanDuration(*span)
+}
+
+func getSpanDuration(span Span) string {
```
schoren, line 58: I'd rename this to `spanDuration` so it's consistent with `spanType`

schoren (approved):
LGTM. Left a few nit comments

server/traces/traces.go
mathnogueira: We already have a `spanDuration` function right now here. I can rename it when I refactor the way we convert from OTEL to a trace.

mathnogueira: I renamed the older function. We will delete it anyway in a near future

Issue #1086, opened 2022-08-17
[SPIKE] Deep Linking


xoscar:
Investigate the different options to keep the app state in sync with the URL.

Acceptance criteria:

AC1:
As a developer
I should have an easy way to hoist the state of a specific slice to the URL.

AC2:
As a user going to a URL that contains a piece of the state
When I land on the website
The app should inject the parameters into the state.

AC3:
As a user looking through the app
When interacting with the application 
I should see the URL changing depending on key actions.

Issue #1052, opened 2022-08-09
Deep link to test result


jorgeepc:
**Objective:** We need a way to deep link to failed assertions in a given test run.

**AC1:**
When we show in the `CLI` that an assertion failed, 
we need to be able to go to a url and that assertion should be opened/highlighted.

PR #1091, opened 2022-08-18
🔗 1086 adding deep-linking feature

xoscar:
This PR adds the base logic to support deep linking across the application. It includes a single-way data flow where components update the URL query by using an action. Which is then listened by a redux middleware which triggers the side effect actions based on the URL change.

## Changes

- Component no longer update the selected span or assertion directly.

## Fixes

- https://github.com/kubeshop/tracetest/issues/1052
- https://github.com/kubeshop/tracetest/issues/1086

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test

https://www.loom.com/share/121648cb912248be9bd4f67b1ed31c0a


cescoferraro:
@xoscar  Code looks good at first glance. Thank you! One thing I noticed on your demo, is that we are passing the selector to the url. I get that we need this because the assertion's id is it's selector. But can we mask this, let's say by hashing it someway, so users wont notice that?

xoscar:
> @xoscar Code looks good at first glance. Thank you! One thing I noticed on your demo, is that we are passing the selector to the url. I get that we need this because the assertion's id is it's selector. But can we mask this, let's say by hashing it someway, so users wont notice that?

Great idea, I'll work on that!

xoscar:
Hey @jorgeepc & @cescoferraro this one should be ready for the final review

cescoferraro (approved):
LGTM

Issue #976, opened 2022-07-22
Issue with test definition file showing blank definitions for other trigger types


kdhamric:
A test triggered by a gRPC event is still showing a 'httpRequest' area in the trigger definition.  It is blank, but it probably should not be there.  

Conversely, a test triggered by a http trigger shows a blank grpc trigger info.

It will be cleaner if we do not show the 'blank' definitions for the unused trigger type

A grpc triggered test
`description: Import a Pokemon
id: 79ad313c-7504-45a3-96f7-c27fb2933cbe
name: GRPC - Pokemon - Import
trigger:
    grpc:
        address: demo-pokemon-api.demo.svc.cluster.local:8082
        authentication:
            apiKey:
                in: ""
                key: ""
                value: ""
            basic:
                password: ""
                user: ""
            bearer:
                token: ""
        method: pokeshop.Pokeshop.importPokemon
        protobufFile: |
            syntax = "proto3";

            option java_multiple_files = true;
            option java_outer_classname = "PokeshopProto";
            option objc_class_prefix = "PKS";

            package pokeshop;

            service Pokeshop {
              rpc getPokemonList (GetPokemonRequest) returns (GetPokemonListResponse) {}
              rpc createPokemon (Pokemon) returns (Pokemon) {}
              rpc importPokemon (ImportPokemonRequest) returns (ImportPokemonRequest) {}
            }

            message ImportPokemonRequest {
              int32 id = 1;
            }

            message GetPokemonRequest {
              optional int32 skip = 1;
              optional int32 take = 2;
            }

            message GetPokemonListResponse {
              repeated Pokemon items = 1;
              int32 totalCount = 2;
            }

            message Pokemon {
              optional int32 id = 1;
              string name = 2;
              string type = 3;
              bool isFeatured = 4;
              optional string imageUrl = 5;
            }
        request: '{"id":52}'
    httpRequest:
        authentication:
            apiKey:
                in: ""
                key: ""
                value: ""
            basic:
                password: ""
                user: ""
            bearer:
                token: ""
        method: ""
        url: ""
    type: grpc`
    
A http triggered test    
    `description: Get a Pokemon
id: e848f98a-75e1-4ea6-bcf0-473dc86065a4
name: Pokemon - List
trigger:
    grpc:
        address: ""
        authentication:
            apiKey:
                in: ""
                key: ""
                value: ""
            basic:
                password: ""
                user: ""
            bearer:
                token: ""
        method: ""
        protobufFile: ""
[...]

PR #1097, opened 2022-08-19
fix: replace id in test file instead of rewriting whole file

mathnogueira:
This PR fixes the problems related to empty fields being added to the test file after running a test. It only adds the id property to the file if it doesn't exist and replaces it if it exists.

Related to issue #976 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
LGTM!

PR #1100, opened 2022-08-19
add example of how to connect tracetest and opensearch

mathnogueira:
This PR adds an example of how one can setup tracetest to connect to an opensearch instance and use it as a tracing backend.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

### DIFF
--- a/examples/tracetest-opensearch/README.md
+++ b/examples/tracetest-opensearch/README.md
@@ -6,7 +6,7 @@ This repository objective is to show how you can configure your tracetest instan
 
 1. [Install the tracetest CLI](https://github.com/kubeshop/tracetest/blob/main/docs/installing.md#cli-installation)
 2. Run `tracetest configure` on a terminal and type: `http://localhost:8080` to make your CLI send all requests to that address
-3. Run the project by using docker-compose: `docker-compose up`
+3. Run the project by using docker-compose: `docker-compose up` (Linux) or `docker compose up` (Mac)
 4. Test if it works by running: `tracetest test run -d tests/list-tests.yaml`. This would trigger a test that will send and retrieve spans from the opensearch instance that is running on your machine.
 
 ## Project structure
--- a/examples/tracetest-opensearch/README.md
+++ b/examples/tracetest-opensearch/README.md
@@ -9,6 +9,8 @@ This repository objective is to show how you can configure your tracetest instan
 3. Run the project by using docker-compose: `docker-compose up` (Linux) or `docker compose up` (Mac)
 4. Test if it works by running: `tracetest test run -d tests/list-tests.yaml`. This would trigger a test that will send and retrieve spans from the opensearch instance that is running on your machine.
 
+> :warning: Note: The OpenSearch configuration used for this example is not meant to be used in production.
+
 ## Project structure
 
 - `opensearch` is a folder that contains all configuration files used to configure the local opensearch instance;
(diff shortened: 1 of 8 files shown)


examples/tracetest-opensearch/docker-compose.yml
```diff
@@ -0,0 +1,82 @@
...
+        ports:
+            - 8080:8080
+        depends_on:
+            postgres:
+                condition: service_healthy
+            otel-collector:
+                condition: service_started
+
+    postgres:
+        container_name: postgres
```
schoren, line 20: This setting is generating issues with a preexisting container: 
Screen Shot 2022-08-23 at 12 31 36

examples/tracetest-opensearch/docker-compose.yml
```diff
@@ -0,0 +1,82 @@
...
+            - "--config"
+            - "/otel-local-config.yaml"
+        volumes:
+            - ./collector.config.yaml:/otel-local-config.yaml
+        depends_on:
+            - data-prepper
+
+    data-prepper:
+        restart: unless-stopped
+        container_name: data-prepper
```
schoren, line 49: This one is required, for some reason, for the connections to work. Sounds like a bug, and I'd look into removing this setting, because it can conflict with existing containers

examples/tracetest-opensearch/README.md
```diff
@@ -0,0 +1,16 @@
+# Tracetest + Opensearch
+
+This repository objective is to show how you can configure your tracetest instance to connect to an opensearch instance and use it as its tracing backend.
+
+## Steps
+
+1. [Install the tracetest CLI](https://github.com/kubeshop/tracetest/blob/main/docs/installing.md#cli-installation)
+2. Run `tracetest configure` on a terminal and type: `http://localhost:8080` to make your CLI send all requests to that address
+3. Run the project by using docker-compose: `docker-compose up`
```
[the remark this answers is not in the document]
mathnogueira: This only applies to Mac. Docker on Linux doesn't ship with compose. I'll add two instructions. One for mac and another for linux.

Screenshot from 2022-08-23 12-45-06

examples/tracetest-opensearch/docker-compose.yml
mathnogueira: Removed all `container_name` attributes from the docker-compose file

examples/tracetest-opensearch/docker-compose.yml
mathnogueira: Are you sure? It worked fine on my machine after I removed the container_name from all services

examples/tracetest-opensearch/README.md
```diff
@@ -0,0 +1,16 @@
+# Tracetest + Opensearch
+
+This repository objective is to show how you can configure your tracetest instance to connect to an opensearch instance and use it as its tracing backend.
+
+## Steps
+
+1. [Install the tracetest CLI](https://github.com/kubeshop/tracetest/blob/main/docs/installing.md#cli-installation)
+2. Run `tracetest configure` on a terminal and type: `http://localhost:8080` to make your CLI send all requests to that address
```
kdhamric, line 8: Do we need to tell the user what the expectations are for location of OpenSearch and that they may need to look at X, Y, and Z config file to ensure it matches their environment? It looked like there were some hardcoded paths such as http://opensearch:9200 that could be different in different environments

kdhamric (approved):

examples/tracetest-opensearch/README.md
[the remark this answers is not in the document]
schoren: no, it's not specific for mac. it's the new way of using it

examples/tracetest-opensearch/docker-compose.yml
schoren: Yea, I `docker compose down` and `up` without the names and it worked now. Probably had some leftover config

PR #1101, opened 2022-08-22
feat: rename testDefinition to spec

mathnogueira:
This PR updates the `testDefinition` term and replaces it with `spec` instead. 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


xoscar:
I'm looking at this, I'll be updating the FE stuff today 😄

xoscar:
@mathnogueira this one should be ready from the FE side. I see that the run trace testing with tracetest step is failing, let me know if you need me to take a look

PR #1103, opened 2022-08-23
fix 'Does not operator' cypress issue

cescoferraro:
This PR fixes cypress issues triggered by the new Operator - Does not contain
Cypress tests relied on the order of the select item for the assertion operator.
Adding the Not contain operator changed the index of the expected result, causing tests to fail.
Still not sure why the PR that introduced this bug did not fail.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


web/cypress/support/commands.ts
```diff
@@ -162,3 +176,22 @@ Cypress.Commands.add('createTest', () => {
...
+  cy.get(`[data-cy=trace-node-database]`, {timeout: 25000}).first().click({force: true});
+  cy.get('[data-cy=add-assertion-button]').click({force: true});
+  cy.get('[data-cy=assertion-form]', {timeout: 10000}).should('be.visible');
+  cy.get('[data-cy=assertion-check-attribute]').type('db');
+  const attributeListId = getAttributeListId(index);
+  cy.get(`${attributeListId} + div .ant-select-item`).first().click({force: true});
+  cy.get('[data-cy=assertion-check-operator]').click({force: true});
+
+  // const comparatorListId = getComparatorListId(index);
+  // cy.get(`${comparatorListId} + div .ant-select-item`).last().click({force: true});
```
xoscar, line 190: should we remove this?

xoscar (approved):
LGTM

PR #1104, opened 2022-08-23
updating docker compose to ease local development

xoscar:
This PR eases the local development and usage for the tracetest app.

## Changes

- Includes a way for containers to talk to the host machine.
- Removes the requirement to run the pokeshop services all of the time
- Exposes the rest of the jaeger ports.

## Fixes

- Running tracetest for local development.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


local-config/config.yaml
```diff
@@ -0,0 +1,10 @@
+http:
+  prefix: someprefix
```
schoren, line 2: this looks like it shouldn't be here

local-config/config.yaml
```diff
@@ -0,0 +1,10 @@
...
+maxWaitTimeForTrace: 10s
+postgresConnString: "host=127.0.0.1 user=tracetest password=not-secure-database-password  port=5432 sslmode=disable"
+jaegerConnectionConfig:
+  endpoint: localhost:16685
+  tls:
+    insecure: true
+googleAnalytics:
+  enabled: false
```
mathnogueira, line 10: This config file is in an old format. Take a look at the new format: https://github.com/kubeshop/tracetest/blob/main/local-config/config.tests.yaml

local-config/config.yaml
```diff
@@ -0,0 +1,10 @@
+http:
+  prefix: someprefix
+maxWaitTimeForTrace: 10s
+postgresConnString: "host=127.0.0.1 user=tracetest password=not-secure-database-password  port=5432 sslmode=disable"
+jaegerConnectionConfig:
+  endpoint: localhost:16685
```
schoren, line 6: unless really neccesary, I'd try to use the docker internal network as much as possible. I imagine tracetest should be able to talk with jaeger using the `jaeger:16685` url

local-config/config.yaml
schoren, line 10: I think we should have GA enabled by default for new people

docker-compose.yaml
```diff
@@ -57,14 +55,18 @@ services:
...
-      - "16685:16685"
+      - 5775:5775/udp
+      - 6831:6831/udp
+      - 6832:6832/udp
+      - 5778:5778
+      - 4318:4318
+      - 16685:16685
+      - 16686:16686
+      - 14268:14268
+      - 9411:9411
```
schoren, line 69: is it necessary to expose ALL jaeger ports?

local-config/config.yaml
kdhamric: We probably need to rename it now that we are using Segment. Perhaps 'collectTelemetry:' instead?

local-config/config.yaml
xoscar: We are not even using this file so I just deleted it 😅

local-config/config.yaml
xoscar: We are not even using this file so I just deleted it 😅

docker-compose.yaml
xoscar: Updated 😄

schoren (approved):
LGTM!

PR #1112, opened 2022-08-25
feat: add server install cli cmd

schoren:
This PR adds a `server install` cli command. At the moment it can only handle docker compose deployments, but it can set up linux systems and do some basic config over the stack (demo enabled or not)

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/cmd/version.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/spf13/cobra"
+)
+
+var versionCmd = &cobra.Command{
+	Use:    "version",
+	Short:  "cli versions",
+	Long:   "Display this CLI tool version",
+	PreRun: setupCommand,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println(config.Version)
```
mathnogueira, line 16: We really needed that :heart:

mathnogueira (approved):
Awesome work

PR #1136, opened 2022-09-02
Run Cypress test inside a container

### LINKED ISSUE #1117: [Frontend] Improve Cypress configuration
Currently, our cypress tests only work if you run them with a particular setup that is:
1. The backend code must be running inside the k8 cluster.
2. In the same cluster, the pokeshop must also be running.
3. The FE types and specs need to match the current k8 version of the backend.

This is blocking us from running E2E test with a pure local environment, having the backend/pokeshop running on a local container, and validating that the FE won't break based on a specific non-released backend change.

The output of this ticket should be:
1. Having a config setup where we can run the E2E test straight from docker compose.
2. E2E test should be executed locally to test any backend changes before releasing them.

cescoferraro:
This PR adds a docker-compose configuration for us to be able to run cypress tests locally inside a container to test the local environment before committing new features

## Changes

- updates cypress to version 10
- adds docker-compose related configuration
-  add a REACT_APP_POKE_API environment variable to set the pokemon API URL

## Fixes

- fixes #1117 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


Dockerfile
```diff
@@ -5,7 +5,10 @@ COPY ./web/package.json ./
 COPY ./web/package-lock.json ./
 RUN npm ci --silent
 COPY ./web ./
-RUN npm run build
+
+ARG REACT_APP_POKE_API
```
xoscar, line 9: I think this needs to change from ARG to ENV as ARGS are only available during the build phase and with env we can change it during execution. If we need it before building. Then I think we need to use a similar approach that we have for the analytics config where the BE prints the config to the FE from the `index.html` file

xoscar (changes requested):
Hey @cescoferraro great work dude! I added a comment about configuration but the rest of the changes look amazing!

docker-compose.yaml
```diff
@@ -214,3 +215,20 @@ services:
       OPENSEARCH_PASSWORD: admin
     depends_on:
       - opensearch
+  cypress:
```
schoren, line 218: Please, add this to the `local-config/docker-compose.testrunner.yaml` file, so we keep "production" containers separated from "test" containers. You can use https://github.com/kubeshop/tracetest/blob/main/run-tracetesting-tests.sh as an example

Dockerfile
schoren, line 9: this should be an `ENV`, not an `ARG`

xoscar (approved):
GJ my man! 🔥

server/Makefile
```diff
@@ -1,8 +1,10 @@
 VERSION ?= "dev"
 TRACETEST_ENV ?= "dev"
+POKE_API?= "dev"
```
schoren, line 3: a good default for this would be `demo-api`:
```suggestion
POKE_API?= "demo-api:8082"
```

Also, it would be awesome to inject the same but for the grpc container

server/Makefile
```diff
@@ -1,8 +1,10 @@
...
 GO_LDFLAGS := $(shell echo \
 		-X "'github.com/kubeshop/tracetest/server/app.Version=$(VERSION)'" \
 		-X "'github.com/kubeshop/tracetest/server/app.Env=$(TRACETEST_ENV)'" \
+		-X "'github.com/kubeshop/tracetest/server/app.PokeApi=$(POKE_API)'" \
```
schoren, line 7: Don't forget to update the pipelines:

- [ ] `.github/workflows/deploy-main.yml#L65`
- [ ] `.github/workflows/pull-request.yml#L126`
- [ ] `.github/workflows/release-version.yml#L68`
- [ ] `.github/workflows/release-version.yml#L199`

server/app/app.go
```diff
@@ -30,6 +30,7 @@ import (
 var (
 	Version = "dev"
 	Env     = "dev"
+	PokeApi = "dev"
```
schoren, line 33:
```suggestion
	PokeAPI = "dev"
```

local-config/docker-compose.testrunner.yaml
```diff
@@ -18,3 +18,21 @@ services:
         condition: service_healthy
       demo-rpc:
         condition: service_healthy
+      - tracetest
```
schoren, line 21: this line looks like an merge error, it's already defined above to use the health check

server/Makefile
cescoferraro: I dont think we need to add this to the deploys. Frontend code already have a default value, so if the PokeAPI is a empty string it should use the default value. We only need to change the poke API URL when running within the docker compose configuration

server/app/app.go
cescoferraro: done

local-config/docker-compose.testrunner.yaml
cescoferraro: you are right. done

run-tracetesting-tests.sh
```diff
@@ -27,7 +27,7 @@ if [ "$RESTART" == "yes" ]; then
 fi
 docker compose -f docker-compose.yaml up -d --build --remove-orphans
 docker compose -f docker-compose.yaml -f local-config/docker-compose.testrunner.yaml build
-docker compose -f docker-compose.yaml -f local-config/docker-compose.testrunner.yaml run testrunner
+docker compose -f docker-compose.yaml -f local-config/docker-compose.testrunner.yaml run testrunner cypress
```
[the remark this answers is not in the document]
cescoferraro: I am not really good with bash. could not figure out a way to pass arguments while keeping the flag.
so I copied the whole file. maybe we could create a ticket for that

run-tracetesting-tests.sh
```diff
@@ -27,7 +27,7 @@ if [ "$RESTART" == "yes" ]; then
 fi
 docker compose -f docker-compose.yaml up -d --build --remove-orphans
 docker compose -f docker-compose.yaml -f local-config/docker-compose.testrunner.yaml build
-docker compose -f docker-compose.yaml -f local-config/docker-compose.testrunner.yaml run testrunner
+docker compose -f docker-compose.yaml -f local-config/docker-compose.testrunner.yaml run cypress
```
schoren, line 30:
```suggestion
docker compose -f docker-compose.yaml -f local-config/docker-compose.testrunner.yaml run tracetest
```

server/Makefile
schoren: It's not that its necessary, but it's a way to explicitly set the build args everywhere, so it's explicit how to change things

run-tracetesting-tests.sh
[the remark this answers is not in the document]
schoren: that's fine for now, if it becomes tedious we can merge the scripts later

run-tracetesting-tests.sh
cescoferraro: i have 2 copies of this file. this one runs the testrunner. the other one run the cypress test

schoren: Yes, this one should run tracetest, not cypress

schoren (approved):
Looks great now!

PR #1141, opened 2022-09-02
Update FE to use new term Test Specs

jorgeepc:
This PR updates the FE files to use the new term `Test Specs`.

## Changes

- Changed everything that was using Test Definition to Test Specs.

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


web/src/components/TestSpecForm/AssertionCheckList.tsx
```diff
@@ -37,7 +34,7 @@ const AssertionCheckList: React.FC<IProps> = ({form, fields, add, remove, attrib
               attributeList={attributeList}
               name={name}
               index={index}
-              assertionList={assertionList}
```
xoscar, line 40: Why don't you like the list naming for arrays? 😄

xoscar (approved):
GJ @jorgeepc LGTM!

web/src/components/TestSpecForm/AssertionCheckList.tsx
jorgeepc: Personal preference. I don't think you need to explain with the name the type of variable. However I'm ok with that.

PR #1167, opened 2022-09-07
Add clientID to analytics

schoren:
This PR updates the server and CLI so the client can pass its machineID to the server and use that info to relate events from cli/server

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira (approved):
Awesome

PR #1170, opened 2022-09-07
feat: assertions as string

mathnogueira:
This PR changes the API to receive assertions as strings instead of objects

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira:
@cescoferraro you can use this PR to update the frontend as well

mathnogueira:
We will probably not need this. Talked with @cescoferraro and found a better way to handle this.

PR #1188, opened 2022-09-09
setup tests for examples when PRs are open

mathnogueira:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -129,6 +129,27 @@ jobs:
         rm -rf /tmp/.buildx-cache
         mv /tmp/.buildx-cache-new /tmp/.buildx-cache
 
+  test-examples:
+    needs:  [build-docker]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        example_dir: [tracetest-opensearch, tracetest-tempo]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build example
+        run: |
+          cd examples/${{ matrix.example_dir }}
+          sed -i 's/kubeshop\/tracetest:.*/kubeshop\/tracetest:${{ github.event.pull_request.number }}/g' docker-compose.yml
+          docker-compose up -d
+      - name: Build CLI
+        run: |
+          cd cli
+          go build -o ./tracetest main.go
+      - name: Run example test
+        run: |
+          ./cli/tracetest test run examples/${{ matrix.example_dir }}/tests/list-tests.yaml
+
   deploy:
     needs:  [unit-test-frontend, unit-test-backend, unit-test-cli, build-docker]
     name: Deploy


server/tracedb/tempodb.go
```diff
@@ -55,6 +55,10 @@ func (ttd *tempoTraceDB) GetTraceByID(ctx context.Context, traceID string) (trac
 		return traces.Trace{}, fmt.Errorf("tempo err: %w", err)
 	}
 
+	if resp.Trace == nil {
+		return traces.Trace{}, ErrTraceNotFound
+	}
```
mathnogueira, line 60: This issue was found while running the tests :smile:

.github/workflows/pull-request.yaml
```diff
@@ -129,6 +129,35 @@ jobs:
...
+          docker-compose up -d
+      - name: Build CLI
+        run: |
+          cd cli
+          go build -o ./tracetest main.go
+          echo "http://localhost:8080" | ./tracetest configure -g
+      - name: Run example test
+        run: |
+          timeout 60 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' localhost:8080)" != "200" ]]; do sleep 5; done' || false
+          ./cli/tracetest test run -d examples/${{ matrix.example_dir }}/tests/list-tests.yaml --wait-for-result
```
schoren, line 159: is it possible to run all the dogfooding test for each example? that way we enforce full compatibility for all backends

examples/tracetest-signalfx/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 services:
 
     tracetest:
-        image: kubeshop/tracetest:v0.6.5
+        image: kubeshop/tracetest:v0.6.8
```
schoren, line 5: If you use a version here, you won't be testing the actual code being built. you need to do a `build` or use the `sha-*` tag for the commit. You already depend on the `build docker` step, so that makes more sense

schoren (changes requested):

examples/tracetest-signalfx/docker-compose.yml
mathnogueira: I'm replacing those values before. I just bumped the values just to make sure if a user uses those examples, they are using a more stable version.

schoren: You can use env vars. It would be a better apprach, because it's easier to read, very obvious you're replacing the value, and we don't need to maintain the examples:

https://docs.docker.com/compose/environment-variables/

.github/workflows/pull-request.yaml
mathnogueira: We can but I rather work on that after we get the chaining tests right. I don't feel the bash solution is very stable right now and this might make our pipeline more brittle

schoren: Since the dogfood step runs isolated from the frontend tests, I didn't see it randomly failing anymore, so I'd feel confident in doing that

schoren:
I tink you also need to add `examples/` to the `paths` on `.github/workflows/pull-request.yaml`

examples/tracetest-signalfx/docker-compose.yml
mathnogueira: tried to use `compose run` but hit some problems. The easiest way was to generate a `.env` file and use it instead.

.github/workflows/pull-request.yaml
mathnogueira: to make it work well, I think we would still use docker-compose and use it as both "integration" and "target"  instances. Is that right? So we don't have to maintain opensearch and tempo in our cluster.

examples/tracetest-signalfx/docker-compose.yml
schoren: Sorry for the confusion, my suggestion wasn't to use `run` instead of `up`. You don't need a `.env` file, you can set the env, or prepend the value to the cmd, like `TAG=v0.1.2 docker compose up`

.github/workflows/pull-request.yaml
mathnogueira: In order for that to work, we will have to change all our examples to make them use the pokemon api. I'd do that but in another PR.

mathnogueira: @schoren is that ok? can this be merged?

examples/tracetest-signalfx/docker-compose.yml
schoren, line 5:
```suggestion
         image: kubeshop/tracetest:${TAG:-latest}
```

schoren (approved):
There's a hardcoded tag on a docker compose file, left a suggestion to fix. other than that it's good to go!

PR #1199, opened 2022-09-12
CLI installer: kubernetes

schoren:
This PR adds support for k8s installer

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/installer/kubernetes.go
+++ b/cli/installer/kubernetes.go
@@ -42,67 +42,16 @@ func kubernetesInstaller(config configuration, ui UI) {
 }
 
 func installTracetest(conf configuration, ui UI) {
-	execCmd(
-		helmCmd(conf, "repo add --force-update kubeshop https://kubeshop.github.io/helm-charts"),
-		ui,
-	)
-	execCmd(
-		helmCmd(conf, "repo update"),
-		ui,
-	)
+	setupHelmRepo(conf, ui)
 
-	cmd := []string{
-		"upgrade --install tracetest kubeshop/tracetest",
-		"--namespace " + conf.String("k8s.namespace") + " --create-namespace",
-	}
+	installTracetestChart(conf, ui)
+	fixTracetestConfiguration(conf, ui)
 
-	if conf.Bool("k8s.expose") {
-		cmd = append(cmd, []string{
-			"--set ingress.enabled=true",
-			"--set 'ingress.hosts[0].host=" + conf.String("k8s.ingress-host") +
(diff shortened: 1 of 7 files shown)


cli/installer/kubernetes.go
```diff
@@ -21,7 +25,258 @@ var kubernetes = installer{
...
+		)
+	}
+
+	execCmd(helmCmd(conf, cmd...), ui)
+
+	psql := "host=tracetest-postgresql user=tracetest password=not-secure-database-password  port=5432 sslmode=disable"
+	c := getTracetestConfigFileContents(psql, ui, conf)
+	ttc := createTmpFile("tracetest-config", string(c), ui)
+	defer os.Remove(ttc.Name())
+	execCmd(
```
mathnogueira, line 70: I saw you have a `execCmdIgnoreErrors` function. And I see you also use `execCmd` without assigning the returned error. Do you really need that `execCmdIgnoreErrors` function?

cli/installer/kubernetes.go
```diff
@@ -21,7 +25,258 @@ var kubernetes = installer{
...
+cd $tmpdir/helm-chart
+helm dependency update
+
+helm upgrade --install demo . \
+  --namespace demo --create-namespace \
+  -f values.yaml \
+  --set image.tag=latest \
+  --set image.pullPolicy=Always \
+  --set postgres.auth.username=ashketchum,postgres.auth.password=squirtle123,postgres.auth.database=pokeshop \
+  --set rabbitmq.auth.username=guest,rabbitmq.auth.password=guest,rabbitmq.auth.erlangCookie=secretcookie \
```
mathnogueira, line 153: I think you forgot to set the jaeger config. I cannot get traces back.

mathnogueira (changes requested):

cli/installer/kubernetes.go
schoren: `execCmd` internally calls `_execCmd` and exists on error. `execCmdIgnoreErrors` calls `_execCmd` and ignores errors. I don't want to directly call `_execCmd`

cli/installer/kubernetes.go
schoren: good catch. It's fixed now

mathnogueira: Screenshot from 2022-09-13 14-02-31
The Host is correct, but the port is set as the `namespace` instead.

schoren: Fixed!

mathnogueira (approved):
Great, everything works now! GJ

Issue #1181, opened 2022-09-08
Frontend should validate the trigger response before parsing it to JSON


xoscar:
Responses from triggers can be HTML, XML, or other mime-types, we need to validate the response first before rendering it as JSON.

PR #1205, opened 2022-09-13
1181 adding code block and response validation for the test run trigger

xoscar:
This PR adds FE validations to check if the trigger response is not a JSON, it tries to match it with the supported mime-types coming from the headers info.

## Changes

- Updates the response body section to match the code editor.

## Fixes

- #1181 

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira (approved):
LGTM

PR #1210, opened 2022-09-13
disabled edit form submit button when running test

cescoferraro:
This PR disable the edit form submit button while the test is being run

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/components/EditTest/EditTest.tsx
+++ b/web/src/components/EditTest/EditTest.tsx
@@ -1,10 +1,12 @@
 import {Button, Form} from 'antd';
+import EditTestForm from 'components/EditTestForm';
+import {TriggerTypeToPlugin} from 'constants/Plugins.constants';
+import useValidateTestDraft from 'hooks/useValidateTestDraft';
+import {useTest} from 'providers/Test/Test.provider';
 import {useCallback} from 'react';
 import {TDraftTest, TTest} from 'types/Test.types';
-import EditTestForm from 'components/EditTestForm';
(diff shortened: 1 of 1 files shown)


web/src/components/EditTest/EditTest.tsx
```diff
@@ -37,8 +41,8 @@ const EditTest = ({test}: IProps) => {
           </Button>
           <Button
             data-cy="edit-test-submit"
-            loading={isEditLoading}
-            disabled={!isValid}
+            loading={!stateIsFinished || isEditLoading}
```
xoscar, line 44: After the save button is done the new version of the test should be available, for me the loading state has to match the form save request, not the trace process. I think the way it was matched that.

cescoferraro: I have done this without changing the loading prop. But the button flicks the loading very fast because the api responds very fast. It looks weird. And second, because the button label reads “Save & Run” so it makes sense for me having it loading while it is running. But it is pretty easy to just revert that line. What should we do? Maybe @olha23 can chime in. I will record a loom video showing the difference

cescoferraro: I have reverted back the loading prop to what it was so we can approve this.
We can discuss this behaviour on the parking lot tommorow @xoscar

xoscar (approved):
LGTM

PR #1211, opened 2022-09-13
disable onboarding

cescoferraro:
This PR disables the current onboarding flow

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/cypress/support/commands.ts
+++ b/web/cypress/support/commands.ts
@@ -119,7 +119,7 @@ Cypress.Commands.add('cancelOnBoarding', () => {
   const parsedValue = value ? JSON.parse(value) : undefined;
 
   if (!parsedValue || parsedValue.trace === false) {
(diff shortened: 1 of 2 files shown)


web/cypress/support/commands.ts
```diff
@@ -119,7 +119,7 @@ Cypress.Commands.add('cancelOnBoarding', () => {
   const parsedValue = value ? JSON.parse(value) : undefined;
 
   if (!parsedValue || parsedValue.trace === false) {
-    cy.get('[data-cy=no-thanks]').click();
+    // cy.get('[data-cy=no-thanks]').click();
```
xoscar, line 122: I'm glad we have this so we only need to remove it from one place. GJ!

xoscar (approved):
GJ @cescoferraro! Let's keep this Pr in our back pocket for when we need to add it again

PR #1168, opened 2022-09-07
basic setup for the collector POC

mathnogueira:
This PR adds the PoC for using the collector to get spans directly from the collector. It only adapts the code for the HTTP trigger.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

PR #1212, opened 2022-09-14
feat: otlp endpoint + working collector PoC

mathnogueira:
This PR implements the collector PoC

This includes the code from #1168 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/runner.go
```diff
@@ -135,10 +134,15 @@ func (r persistentRunner) processExecQueue(job execReq) {
...
 	job.test = newTest
 
-	response, err := triggerer.Trigger(job.ctx, job.test)
+	traceID := model.IDGen.TraceID()
+	run.TraceID = traceID
+	r.handleDBError(r.updater.Update(job.ctx, run))
+
+	response, err := triggerer.Trigger(job.ctx, job.test, &trigger.TriggerOptions{
+		TraceID: traceID,
+	})
```
mathnogueira, line 143: I had to change that because the traceID would only be saved **after** the initial span was sent. This could cause race conditions (and I experienced a problem with it). So, by saving the run first and then triggering the transaction (injecting the same traceID saved into the run) we ensure that at the moment spans get sent to the otlp endpoint, we know the traceID and can add the spans to the right test run.

examples/collector-poc/collector.config.yaml
```diff
@@ -0,0 +1,34 @@
...
+      http:
+
+
+processors:
+  tail_sampling:
+    decision_wait: 5s
+    policies:
+      - name: tracetest-spans
+        type: trace_state
+        trace_state: { key: tracetest, values: ["true"] }
```
xoscar, line 14: 🔥

examples/collector-poc/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+            POSTGRES_USER: postgres
+        ports:
+            - 5432:5432
+        healthcheck:
+            test: pg_isready -U "$$POSTGRES_USER" -d "$$POSTGRES_DB"
+            interval: 1s
+            timeout: 5s
+            retries: 60
+
+    otel-collector:
```
xoscar, line 31: No Jeager instance 🪨

server/app/app.go
```diff
@@ -173,6 +174,9 @@ func (a *App) Start() error {
 		return err
 	}
 
+	// Start otlp endpoint
+	go func() { otlp.StartServer(21321, a.db) }()
```
xoscar, line 178: This is a goroutine right?

server/otlp/server.go
```diff
@@ -0,0 +1,110 @@
+package otlp
```
xoscar, line 1: 🚀

xoscar (approved):
Great job @mathnogueira! do you mind presenting it during tomorrow's planning session for the team?

server/app/app.go
mathnogueira: Yes it is. As this is a blocking operation, we need to start it in another thread to be able to run the HTTP server and this server at the same time

server/otlp/server.go
```diff
@@ -0,0 +1,110 @@
...
+	v1 "go.opentelemetry.io/proto/otlp/trace/v1"
+	"google.golang.org/grpc"
+)
+
+type Server struct {
+	pb.UnimplementedTraceServiceServer
+	db model.Repository
+}
+
+func StartServer(port int, db model.Repository) error {
```
schoren, line 22:
```suggestion
func StartServer(port int, db model. RunRepository) error {
```

server/otlp/server.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (s Server) Export(ctx context.Context, request *pb.ExportTraceServiceRequest) (*pb.ExportTraceServiceResponse, error) {
+	if len(request.ResourceSpans) == 0 {
+		return &pb.ExportTraceServiceResponse{}, nil
+	}
+
+	spansByTrace := s.getSpansByTrace(request)
+
+	fmt.Println(spansByTrace)
```
schoren, line 48: can we remove this?

server/tracedb/tracedb_test.go
```diff
@@ -68,7 +68,7 @@ func TestCreateClient(t *testing.T) {
 			cl := c
 			t.Parallel()
 
-			actual, err := tracedb.New(cl.config)
+			actual, err := tracedb.New(cl.config, nil)
```
schoren, line 71: let's fix this `New` method (in the future) so it doesn't require all the deps regardless of the final type being created

schoren (approved):
few nits, but looks awesome!

PR #1250, opened 2022-09-20
Create new onboarding experience

cescoferraro:
This PR creates the new onboarding expirience by this design https://www.figma.com/file/TMShLwGmcTwX7uRodvyRoE/Onboarding

## Changes

- Add New design

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


xoscar (changes requested):
Hey @cescoferraro I found a weird visual bug



Also, when clicking outside the onboarding popup it closes and ends the onboarding process.
And when you click prev the positions are wrong.

cescoferraro:
@jorgeepc @xoscar the issue reported by oscar in the last comment is because our different modes (Trigger/Trace/Test) are placed on different URLs. 
So when the onboarding goes from one URL to another, the next step component is not rendered. So the tour does not know where to place the dialog.
To solve the issue I have added 2 other steps, so we can showcase the Tabs Header while we navigate to the Test Mode so the glitch does not work on the proposed onboarding.

xoscar:
@cescoferraro I'm still seeing the weird behavior on some of the steps where it starts to shake. It also still has the issue that clicking elsewhere on the screen would close the onboarding.

cescoferraro:
@xoscar but closing when clicking elsewhere on the screen is the default behavior.
The shaking thing is hard to prevent when switching URLs. I think I need to give it a try with another library.
There is no way of making react-tour delay the navigating before presenting the new step.

web/package.json
```diff
@@ -48,12 +48,14 @@
...
     "react-flow-renderer": "10.3.7",
     "react-intersection-observer": "9.1.0",
+    "react-joyride": "2.5.3",
     "react-json-view": "1.21.3",
     "react-redux": "7.2.6",
     "react-reflex": "4.0.9",
     "react-router-dom": "6.2.1",
     "react-scripts": "5.0.1",
     "react-syntax-highlighter": "15.5.0",
+    "react-use": "17.4.0",
```
xoscar, line 58: Should we remove the react tour dependency if we are not using it anymore?

xoscar (approved):
Pre approving, GJ @cescoferraro

PR #1257, opened 2022-09-20
fix: prefer apt over brew

mathnogueira:
This PR fixes docker-compose installation on Linux when brew is available.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/installer/util.go
+++ b/cli/installer/util.go
@@ -63,10 +63,10 @@ func detectPkgManager() pkgManager {
 	}
 
 	switch true {
+	case commandExists("apt"):
+		lastDetectedPkgManager = apt
 	case commandExists("brew"):
 		lastDetectedPkgManager = homebrew
-	case commandExists("apt-get"):
-		lastDetectedPkgManager = apt
 	case commandExists("dnf"):
 		lastDetectedPkgManager = dnf
 	case commandExists("yum"):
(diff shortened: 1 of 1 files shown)


cli/installer/util.go
```diff
@@ -63,10 +63,10 @@ func detectPkgManager() pkgManager {
 	}
 
 	switch true {
+	case commandExists("apt"):
```
mathnogueira, line 66: apt-get is deprecated

schoren: it's not deprecated. it's intended for scripting, while `apt` is intended for manual user interaction. https://askubuntu.com/a/829866

https://manpages.ubuntu.com/manpages/xenial/man8/apt.8.html
https://linux.die.net/man/8/apt-get

schoren (changes requested):
please revert back to using apt-get. the order looks fine

cli/installer/util.go
```diff
@@ -63,10 +63,10 @@ func detectPkgManager() pkgManager {
 	}
 
 	switch true {
-	case commandExists("brew"):
-		lastDetectedPkgManager = homebrew
 	case commandExists("apt-get"):
 		lastDetectedPkgManager = apt
+	case commandExists("brew"):
```
schoren, line 68: I assume the same problems wuold happen on any linux with brew, so maybe it's better to move it below all linux package managers?

PR #1271, opened 2022-09-20
fix: opensearch install

mathnogueira:
This PR fixes a couple of problems I faced while trying to install a tracetest server to connect to an OpenSearch instance.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/installer/docker_compose.go
+++ b/cli/installer/docker_compose.go
@@ -166,7 +166,7 @@ func getDockerComposeFileContents(ui UI, config configuration) []byte {
 		ui.Exit(fmt.Sprintf("cannot configure tracetest container: %s", err.Error()))
 	}
 
-	if config.Bool("tracetest.collector.install") {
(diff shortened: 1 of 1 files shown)


cli/installer/docker_compose.go
```diff
@@ -166,7 +166,7 @@ func getDockerComposeFileContents(ui UI, config configuration) []byte {
 		ui.Exit(fmt.Sprintf("cannot configure tracetest container: %s", err.Error()))
 	}
 
-	if config.Bool("tracetest.collector.install") {
+	if config.Bool("tracetest.collector.install") && config.Bool("tracetest.backend.install") {
 		if err := fixOtelCollectorContainer(config, project); err != nil {
```
mathnogueira, line 170: This relies on Jaeger. So it doesn't have to be executed if jaeger is not being installed

mathnogueira:
#1272 probably fixed the issue already.

PR #1279, opened 2022-09-21
Fix autocomplete for expected value in Test Spec form

### LINKED ISSUE #1244: When editing duration and changing the value you can get duplicate units of time
When setting a test spec using the duration, it is easy to end up with the unit of time 'twice'. Here is an example of this occurring:
https://www.loom.com/share/0a5efb11c9eb4e53971128c7af1721eb

jorgeepc:
This PR fixes the autocomplete feature of the `expected` value in the Test Spec form.

## Changes

- Autocomplete logic for expected field

## Fixes

- fixes #1244

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cescoferraro (approved):
Nice! I did not realize there is that syntaxTree method!
Great job @jorgeepc Thanks you for the lesson

PR #1280, opened 2022-09-21
fix: opensearch install

mathnogueira:
This PR fixes the flow for installing Tracetest with an existing tracing backend

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/installer/docker_compose.go
+++ b/cli/installer/docker_compose.go
@@ -327,8 +327,6 @@ func fixOtelCollectorContainer(config configuration, project *types.Project) err
 	}
 
 	ocs.Volumes[0].Source = path.Join(config.String("output.dir"), otelCollectorConfigFilename)
-	je := config.String("tracetest.backend.endpoint.collector")
-	ocs.Environment["JAEGER_ENDPOINT"] = &je
 
 	return nil
 }


mathnogueira:
Even with the latest changes from 0.7.1, the CLI still panics

cli/installer/docker_compose.go
```diff
@@ -166,7 +166,7 @@ func getDockerComposeFileContents(ui UI, config configuration) []byte {
 		ui.Exit(fmt.Sprintf("cannot configure tracetest container: %s", err.Error()))
 	}
 
-	if config.Bool("tracetest.collector.install") {
+	if config.Bool("tracetest.collector.install") && config.Bool("tracetest.backend.install") {
```
schoren, line 169: what issue is this trying to fix? Looks like it will ask the user if they want to install the collector (asked after the backend install), but then it won't based on a previous question answer.

Can you explain the goal of this change?

mathnogueira: `fixOtelCollectorContainer` relies on `tracetest.backend.endpoint.collector` to inject the JAEGER endpoint to the collector env object. 

I added a change to check for jaeger instead of `tracetest.backend.install`. But the original problem is that `tracetest.backend.endpoint.collector` doesn't exist for all backend storages. In the case of OpenSearch, because this configuration was not available, the CLI would panic due to [this](https://github.com/kubeshop/tracetest/blob/c4d1c1f0763e179eba45ef16ffed80ae3df0a9c9/cli/installer/installer.go#L101)

mathnogueira: As fixOtelCollectorContainer is only necessary for jaeger installs, checking for the backend type is enough to prevent this issue from happening

cli/installer/docker_compose.go
```diff
@@ -166,7 +166,8 @@ func getDockerComposeFileContents(ui UI, config configuration) []byte {
 		ui.Exit(fmt.Sprintf("cannot configure tracetest container: %s", err.Error()))
 	}
 
-	if config.Bool("tracetest.collector.install") {
+	isJaegerBackend := config.String("tracetest.backend.type") == "jaeger"
+	if config.Bool("tracetest.collector.install") && isJaegerBackend {
```
schoren, line 170: still, this is not good. You need to fix the user input. Or even better, fix it so the collector works with any backend.
You can fix the fixOtelCollectorContainer. It will receive a project that is a parsed version of the docker-compose file. That file is fetch from the example for the selected backend. You can add a switch in that function so it correcty fixes the correct service.

mathnogueira: tbh, I don't think we need the `fixOtelCollectorContainer` function anymore. The example doesn't rely on the JAEGER_ENDPOINT env variable.

The old docker-compose file needed it.

schoren: makes sense. If you test it and it works with at least jaeger and open search, we should be good to go

mathnogueira: Actually we don't need the jaeger env variable. We still need the volume configuration. Just added a new commit for that.

schoren (approved):

PR #1294, opened 2022-09-22
Change default port to 11633

schoren:
This PR changes the default port to 11633

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -223,7 +223,7 @@ jobs:
         run: |
           kubectl port-forward \
             --namespace tracetest-pr-${{ github.event.pull_request.number }} \
-            svc/tracetest-pr-${{ github.event.pull_request.number }} 3000:8080 &
+            svc/tracetest-pr-${{ github.event.pull_request.number }} 3000:11633 &
 
           checkPort() {
               PORT=$1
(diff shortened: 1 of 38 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -287,25 +282,20 @@ jobs:
         run: |
           kubectl port-forward \
             --namespace tracetest-integration \
-            svc/tracetest-integration 8080 &
+            svc/tracetest-integration 11633:8080 &
```
mathnogueira, line 285: Won't this cause a problem with the CLI? You are configuring the CLI to send commands to `localhost:11633`, but here tracetest will be available on port `8080`.

mathnogueira, line 285: You will have to update this right after the PR is merged, right?

schoren: Yes. exactly. `integration` is still running on `8080`, so I need to talk to it, but leave it as ready as possible for later.
So this part is temporary until this PR is merged, and then I have to make a new PR to correctly point to integrations's `11633` port

mathnogueira (approved):
LGTM

jfermi (approved):

PR #1297, opened 2022-09-26
feat: windows installer using docker compose

mathnogueira:
This PR adds support to install Tracetest using docker-compose on Windows.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/installer/cmd.go
```diff
@@ -0,0 +1,139 @@
+package installer
```
mathnogueira, line 1: Just moved all command related functions here instead of utils

cli/installer/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return strings.TrimSpace(out)
+}
+
+func commandExists(cmd string) bool {
+	// windows requires the env to be reloaded before the command is checked. So if the program just got installed
+	// it will be found.
+	if isWindows() {
+		refreshEnvVariables()
+	}
```
mathnogueira, line 127: This is a change in an existing function.

cli/installer/win/cmd.go
```diff
@@ -0,0 +1,44 @@
+package win
+
+import (
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+func ExecCmd(cmd string, suppressOutput bool) error {
```
mathnogueira, line 9: added a `suppressOutput` option here so the CLI doesn't show "Refreshing environment variables" every time we run `refreshEnvVariables`. There's no option to make this command quiet. So I added this here

cli/installer/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+func getCmdOutputClean(cmd string) string {
+	out := getCmdOutput(cmd)
+
+	return strings.TrimSpace(out)
+}
+
+func commandExists(cmd string) bool {
+	// windows requires the env to be reloaded before the command is checked. So if the program just got installed
+	// it will be found.
+	if isWindows() {
```
schoren, line 125: instead of this `if`, you could use build tags to have an empty function for `posix` but an actual function for `win`. See https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

cli/installer/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+		refreshEnvVariables()
+	}
+
+	_, err := exec.LookPath(cmd)
+	return err == nil
+}
+
+func commandSuccess(probeCmd string) bool {
+	if isWindows() {
+		return win.CommandSuccess(probeCmd)
```
schoren, line 135: same here, you can use build tags and remove this if

cli/installer/win/cmd.go
schoren: You can "mute" the cmd: https://stackoverflow.com/a/6461021

cli/installer/cmd.go
mathnogueira: Awesome!

cli/installer/cmd_unix.go
```diff
@@ -0,0 +1,47 @@
+//go:build !windows
+// +build !windows
```
schoren, line 2: Is this neccesary? I think this form is deprecated in favor of the `//go:build`

cli/installer/cmd_windows.go
```diff
@@ -0,0 +1,53 @@
+//go:build windows
+// +build windows
```
schoren, line 2: same here

schoren (approved):
LGTM! left 2 nits on the build tags, but feel free to merge

PR #1303, opened 2022-09-28
add support for kubernetes (minikube doesn't work very well)

mathnogueira:
This PR fixes all functions that were specific to POSIX systems and enables windows to install Tracetest on a K8s cluster. The only problem is that minikube doesn't work well on Windows. I still have to figure out why it cannot mount any volume (probably some docker configuration that we have to document)

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira:
What I see when I try to run tracetest on minikube:
image

xoscar (approved):

PR #1308, opened 2022-09-29
require `attr:` prefix when referencing attributes from span

mathnogueira:
This PR makes the prefix `attr:` required when referencing an attribute using expressions. **This requires frontend changes as well.**

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/encoding/yaml/conversion/parser/assertion_parser_test.go
```diff
@@ -364,7 +364,6 @@ func TestParseAssertion(t *testing.T) {
 			require.NoError(t, err)
 			assert.Equal(t, testCase.ExpectedOutput.Attribute, output.Attribute)
 			assert.Equal(t, testCase.ExpectedOutput.Operator, output.Operator)
-			assert.Equal(t, testCase.ExpectedOutput.Value, output.Value)
```
mathnogueira, line 367: This was used when we had a `string` Value. Now, this is irrelevant and we should use assertExpression instead.

schoren (approved):
LGTM!

web/src/components/TestSpecForm/useExpectedInputLanguage.tsx
```diff
@@ -25,7 +25,9 @@ export function useExpectedInputLanguage() {
             const identifierText = state.doc.sliceString(parentNode.from, parentNode.to);
             const isN = isNumber(identifierText);
 
-            const attributeOptions = Object.values(Attributes).map(s => ({label: s, apply: `${s} `}));
+            const attributeOptions = Object.values(Attributes)
+              .map(s => `attr:${s}`)
+              .map(s => ({label: s, apply: `${s} `}));
```
jorgeepc, line 30: Do we need the double map? 🤔 
```ts
.map(s => ({label: `attr:${s}`, apply: `attr:${s} `}));
```

cescoferraro: I dont think we need it. I just think it is more readable and dryer.
I can change if you want.

cescoferraro: Just committed the fix @jorgeepc

jorgeepc (approved):

PR #1310, opened 2022-09-29
Search not working correctly when using pagination

### LINKED ISSUE #1063: Search not working correctly when using pagination
**Describe the bug**
The frontend search in the Home page is not working correctly when you already requested the second page of Tests.

**To Reproduce**
1. Go to the Home page.
2. Scroll down to request the second page of Tests. 
3. Scroll up and add a search term.
4. You will see that the list of Tests is not changing.

**Expected behavior**
Frontend search should work correctly even if you triggered the pagination.

**Screenshots**
https://www.loom.com/share/e9359c55324441b9a3585f8e8bef639f

**Browser (please complete the following information):**
 - Browser: [All]

**Additional context**
N/A

jorgeepc:
This PR fixes a couple of bugs in the tests home page. They are related to the infinite scroll implementation (including search and delete functionality). It was causing issues like:
- duplicated items in the list
- search not working
- deleted items not being removed from the list

## Changes

- Implemented a new `Pagination` component and `usePagination` hook to handle pages logic.
- Fixes:
   - Duplicated items
   - Search
   - Deleted items

## Fixes

- fixes #1063 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom

https://www.loom.com/share/416bfa068b654fee9b76843f2c65d40d


web/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,69 @@
...
+  hasPrev,
+  isEmpty,
+  isFetching,
+  isLoading,
+  loadingComponent,
+  loadNext,
+  loadPrev,
+}: IProps) => {
+  const handleNextPage = () => {
+    if (isLoading || !hasNext) return;
```
xoscar, line 33: Instead of having this logic here, why we don't just disable the buttons based on the same condition?

xoscar (approved):
I agree with the solution @jorgeepc, having the infinite scroll working with the delete functionality can become really tricky and we'll need to mess up with the redux cache system and such. I would suggest sending this to @olha23 so she can take a look at the UI portion.

PR #1311, opened 2022-09-30
make run IDs sequence independant for each test

schoren:
This PR makes the run ID sequence to be independat for each test. That means that we can have something like:

- test A
  - run 1
  - run 2 
  - run 3
- test B
  - run 1
  - run 2 
  - run 3

This makes it impossible to gather any data from an URL, except the amount of runs a single test has had

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


mathnogueira (approved):
Looks good

PR #1317, opened 2022-09-30
Fix run delete

schoren:
This PR fixes the query to delete a run

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/openapi/api.go
```diff
@@ -49,20 +49,20 @@ type ApiApiServicer interface {
...
 	CreateTestFromDefinition(context.Context, TextDefinition) (ImplResponse, error)
 	DeleteTest(context.Context, string) (ImplResponse, error)
-	DeleteTestRun(context.Context, string, int32) (ImplResponse, error)
-	DryRunAssertion(context.Context, string, int32, TestSpecs) (ImplResponse, error)
-	ExportTestRun(context.Context, string, int32) (ImplResponse, error)
-	GetRunResultJUnit(context.Context, string, int32) (ImplResponse, error)
+	DeleteTestRun(context.Context, string, string) (ImplResponse, error)
+	DryRunAssertion(context.Context, string, string, TestSpecs) (ImplResponse, error)
+	ExportTestRun(context.Context, string, string) (ImplResponse, error)
+	GetRunResultJUnit(context.Context, string, string) (ImplResponse, error)
```
mathnogueira, line 55: did you change that by hand? I'm not seeing any changes in the openapi spec

schoren: In a previous PR, I tried to change the type from string to int, but was way to big of a change, and I could treat the int as a string, so I reverted it. I probably forgot to run the `generate` cmd, so all the generated code was using INT, but the openapi was using string.

I run a `generate` now, and this happend 🤷

mathnogueira (approved):

PR #1322, opened 2022-10-03
feat: cmd configure uses ui package and enable setting values from flags

mathnogueira:
This PR extracts the UI code used in the installer to its own package and changes the configure cmd to use that package instead of using the fmt package directly.

Also, it enables setting values using the flags `--endpoint` and `--analytics`

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

### DIFF
--- a/cli/installer/constants.go
+++ b/cli/installer/constants.go
@@ -1,0 +1,3 @@
+package installer
+
+const createIssueMsg = "If you need help, please create an issue: https://github.com/kubeshop/tracetest/issues/new/choose"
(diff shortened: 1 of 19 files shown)


cli/installer/constants.go
```diff
@@ -0,0 +1,3 @@
+package installer
+
+const createIssueMsg = "If you need help, please create an issue: https://github.com/kubeshop/tracetest/issues/new/choose"
```
schoren, line 3: this could go to the `tracetest.go`, `installer.go` or even `util.go` file and we can remove this useless file

schoren (commented):
just a nit on code organization, otherwise looks great!

schoren (approved):

PR #1332, opened 2022-10-04
Transactions

schoren:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


tracetesting/transactions/http.yaml
```diff
@@ -0,0 +1,9 @@
+type: Transaction
+spec:
```
schoren, line 2: new format, so we can distinguish between transactions and tests

tracetesting/transactions/test-ids.yaml
```diff
@@ -0,0 +1,15 @@
+type: Transaction
+spec:
+  id: TmAKdfV4g
+  name: Test IDs
+  env:
```
schoren, line 5: This could also be named `variables`. It represents the env/vars initial values that will be passed to every test.
Note that this is a initial value, and can be overwritten by a test output with the same name.

This is not canonical, meaning that a test could create an output that is not defined here, and that output will be also passed to subsequent tests.

tracetesting/transactions/tests/create-nonexisting-id.yaml
```diff
@@ -0,0 +1,43 @@
...
+              }
+            }
+          }
+        }
+  specs:
+  - selector: span[name = "Tracetest trigger"]
+    assertions:
+      - tracetest.selected_spans.count = 1
+      - tracetest.response.status = 200
+      - ${attr:tracetest.response.body | json_path '.id' } != ""
```
schoren, line 39: Proposal: allow the assertion attribute to be parsed as an expression, so we can apply "filters" on it.

In this example, we have a `json_path` filter that returns the value that matches the passed jsonPath query. This could be thoug to be equivalent as:
```
func json_path(body string, jsonPathQuery string) string {
  // this is an imaginary function to exemplify the usage
  return jsonpath.Value(body, query)
}
```

tracetesting/transactions/tests/create-nonexisting-id.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - ${attr:tracetest.response.body | json_path '.id' } = "${env:EXAMPLE_TEST_ID}" #
```
schoren, line 40: same here, but comparing against a variable expected value

tracetesting/transactions/tests/http/test-create.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - tracetest.selected_spans.count = 1
+  - selector: span[name = "exec INSERT"]
+    assertions:
+    - tracetest.selected_spans.count = 1
+  - selector: span[name = "exec INSERT"]:first
+    assertions:
+    - sql.query contains "INSERT INTO tests"
+
+  outputs:
+    CREATED_TEST_ID: "${attr:tracetest.request.body | json_path '.id' }"
```
schoren, line 65: this is where the `filter` concept becomes really useful. When setting outputs, you need to set it to a string value. Right now, we don't have a way to extract a single value from a json, so we cannot do this. 

Using the `json_path` hypothetical filter, you could extract any value from a json response. We could provide other filters, like `substring` for example.

tracetesting/transactions/tests/http/test-create.yaml
```diff
@@ -0,0 +1,67 @@
...
+    OVER_DURATION: "${attr:tracetest.span.duration - env:MAX_SPAN_DURATION}"
```
schoren, line 67: Antother useful feature is to support expressions in outputs, so we could do some basic calculations.
In this example, I'm trying to set an output with the amount of time the request went over a defined max duration, defined [here](https://github.com/kubeshop/tracetest/pull/1332/files#diff-26bb94c58b71d986e2ebb16a42a445d743f56f332e181abfc6466861538f9de9R7)

tracetesting/transactions/tests/http/test-create.yaml
xoscar: We have added a regexp field in the front end to have at least some basic support for users. It can be seen in this loom video 

https://www.loom.com/share/a57ef209f68a4e20888e8a39559affec

xoscar (commented):
Hey @schoren I think the proposal is quite good and simple to have something to start with. I added a couple of questions and comments. Let me know what you think 😄

tracetesting/transactions/http.yaml
```diff
@@ -0,0 +1,11 @@
...
+  id: DWKeDBV4g
+  name: HTTP
+  description: Test all http trigger related functions
+  env:
+    MAX_SPAN_DURATION: 500ms
```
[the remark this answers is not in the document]
schoren: That's actually a pretty good idea. I thought about this as a "transaction local env", but we can add this concept as the "global env"

tracetesting/transactions/tests/http/test-create.yaml
schoren: I don't think regex is enough to handle complex cases. We could totally support a `regex` filter, like `${attr:tracetest.request.body | json_path '.id' | regex "\d+" }`  (this would get the ID from the response body, and from that value, apply a regex to get only the numeric parts).

But we need to have a way to get very specific outputs. If we need to rely on regexp to parse complex outputs, it won't work.

Also, what if the response is not json/text, but protobuf encoded? that's a binary format, and regex won't work there

PR #1337, opened 2022-10-05
feat: support assertion names

mathnogueira:
This PR enables the user to define a name to a group of assertions. This is retro-compatible with the existing test spec format.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/model/json.go
+++ b/server/model/json.go
@@ -71,6 +71,29 @@ func (a Assertion) MarshalJSON() ([]byte, error) {
 	})
 }
 
+func (na *NamedAssertions) UnmarshalJSON(data []byte) error {
+	type encodedNamedAssertions struct {
+		Name       string
+		Assertions []Assertion
+	}
+
+	var namedAssertion encodedNamedAssertions
+	if err := json.Unmarshal(data, &namedAssertion); err != nil {
+		// this might be an []Assertion from the older format
+		// try to parse []Assertions instead
+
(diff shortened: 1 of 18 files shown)


server/model/json.go
```diff
@@ -71,6 +71,29 @@ func (a Assertion) MarshalJSON() ([]byte, error) {
...
 }
 
+func (na *NamedAssertions) UnmarshalJSON(data []byte) error {
+	type encodedNamedAssertions struct {
+		Name       string
+		Assertions []Assertion
+	}
+
+	var namedAssertion encodedNamedAssertions
+	if err := json.Unmarshal(data, &namedAssertion); err != nil {
```
schoren, line 81: do we need to keep this old code? I know it's nice, but it's also a maintenance overhead, and we are not in a stable point, so I think it's easier to clean the very few DBs that uses this than to keep this soon-to-rot code

schoren (approved):
LGTM! just left a comment about retro compatibility code, but feel free to merge

server/model/json.go
mathnogueira: I think we should at least make patch versions retro-compatible. The way I look at our versioning right now is that we our `minor` versions are our beta major versions. 

I'm thinking about removing this as soon as we get to v0.8

PR #1350, opened 2022-10-06
feat: new expression engine

mathnogueira:
This PR implements the new expression engine we discussed.

This engine can handle:

* Expressions on both sides of the statement (`1 + 2 = 2 + 1`)
* Attribute referencing (`attr:error_message = "failed to get trace"`)
* String interpolation with expression support (`attr:error_message = "You must be a ${attr:minimum_permission_level} to execute ${attr:action_name}"`)
* Arithmetic operations (addition, subtraction, division, and multiplication)
* Filters: ways of getting just part of the expression result
    *  json_path: filter part of a JSON using JSON path `| json_path '.id'`
    *  regex: return an array of strings that match a given regex `| regex '"id": \d+'`
    *  regex_group: return only the group information from a regex in a flat array `| regex_group '"id": (\d+)'`
    *  get_index: given an array, allows you to get a specific object using its index on the array `| get_index 2`
    

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/model/json_test.go
```diff
@@ -149,29 +148,6 @@ func TestRunEncoding(t *testing.T) {
 	}
 }
 
-func TestOldAssertionSerialization(t *testing.T) {
```
mathnogueira, line 152: We are totally breaking retro-compatibility with this PR, so I'm taking away this test that was here to ensure retro-compatibility

mathnogueira (commented):
Some thoughts about the PR. Would like to hear your opinions @schoren

server/expression/executor_test.go
```diff
@@ -0,0 +1,202 @@
...
+			ShouldPass: true,
+		},
+		{
+			Name:       "should_allow_multiplication",
+			Query:      "8 / 2 = 4",
+			ShouldPass: true,
+		},
+		{
+			Name:       "should_add_durations",
+			Query:      "100ms + 200ms = 300ms",
```
schoren, line 76: 🤯

server/expression/types/types.go
```diff
@@ -0,0 +1,32 @@
+package types
+
+import "regexp"
+
+type Type uint
+
+const (
+	TYPE_NIL       Type = 0
```
schoren, line 8: In go, it's usually preferred to have constant following the [same format as variables](https://stackoverflow.com/a/22688926). Also, you can use `iota` here for easier maintenance (https://golangbyexample.com/iota-in-golang/).

I would rewrite this block as:

```
const (
	TypeNil Type = iota
	TypeString
	TypeNumber
	TypeAttribute
	TypeDuration
)
```

schoren (approved):
this is insane! love how it looks, and how easy it is to understand what's going on. I left a few very minor comments, but this is totally ready to ship!

mathnogueira:
This is part of the changes needed by #1332

PR #1362, opened 2022-10-11
Add FE env configuration layer

### LINKED ISSUE #1347: [Tech Debt] Configuration Object Abstraction
We had a recent bug that was caused because we weren't properly checking one of the configurations coming from the Backend, it was causing the FE app to crash and rendering the app unusable.

We need to create an abstraction layer as protection where the rest of the app can safetly pull values from the config without having to run the checks manually but rather having it only ones at the abstraction layer.

jorgeepc:
This PR adds a FE Env utility to read env variables used by Tracetest. It also parses the env variables to a proper type.

## Changes

- Env util to read parsed variables
- Parse env variables

Types support:

Screen Shot 2022-10-11 at 17 38 15
Screen Shot 2022-10-11 at 17 37 39

## Fixes

- fixes #1347 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/public/index.html
+++ b/web/public/index.html
@@ -23,7 +23,7 @@
     -->
     <title>Tracetest</title>
     <script>
-      const getParser = {
+      const parser = {
         toArray(value) {
           try {
             return JSON.parse(value) || [];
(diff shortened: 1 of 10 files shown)


xoscar (approved):
Hey @jorgeepc gj man!

PR #1365, opened 2022-10-12
fix formatting when not waiting for results

schoren:
This PR makes the CLI show tests as successful when not waiting for resutls. Before, when a test was run without the `--wait-for-result` option, it would show the test as FAILED, when in fact it still didn't have the result.

Also it pretty prints json so it is easier to read while still being parseable

Before: 
Screen Shot 2022-10-12 at 10 50 36

After:
Screen Shot 2022-10-12 at 10 54 33

JSON indented:
Screen Shot 2022-10-12 at 10 57 40

## Changes

-

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/cli/actions/run_test_action.go
+++ b/cli/actions/run_test_action.go
@@ -130,8 +130,9 @@ func (a runTestAction) runDefinition(ctx context.Context, params runTestParams) 
 	}
 
 	tro := formatters.TestRunOutput{
-		Test: test,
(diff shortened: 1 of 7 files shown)


jorgeepc (approved):
LGTM!

xoscar (approved):

Issue #1372, opened 2022-10-13
Multiple HTTP Request Headers with same Key 


abcl-trifork:
**Problem**
Some services in our cluster sometimes receive multiple headers with same key, for example 2 Accept-Language headers with different values.
Tracetest currently sends the header-key once, with the header-value of the last applied Header in Headers List.
We want to test our handling of Field Order rules:  [https://httpwg.org/specs/rfc7230.html#rfc.section.3.2.2](https://httpwg.org/specs/rfc7230.html#rfc.section.3.2.2)

**Solution**
We would like a way to test how our services handle this type of request by:
1. Allowing sending of multiple headers with same key.
2. Allowing assertions on the count or entry of headers, in case there are multiple headers.

**Considerations**

Here are some considerations of how to specify assertions on count/entries:

```
# assert on count of specific header-key
  assertions:
  - count(http.request.header.accept_language) = 2
  - http.request.header.accept_language.count = 2

# assert on first/last header matching key - fails assertion when no header matches key
  assertions:
  - http.request.header.accept_language:first = "["duplicate1"]"
  - http.request.header.accept_language:last = "["duplicate2"]"

# assert on entry of specific duplicated header - fails assertion when entry of headers matching key doesn't exist
  assertions:
  - http.request.header.accept_language[0] = "["duplicate1"]"
  - http.request.header.accept_language[1] = "["duplicate2"]"
```

**Additional context**
Tracetest UI currently doesn't allow multiple headers with same key Add 2 headers:

![image](https://user-images.githubusercontent.com/110457840/195593400-305e304a-aa9e-4f37-8326-a1b3bedd7b12.png)

Header is sent once, with the last value:

![image](https://user-images.githubusercontent.com/110457840/195592988-68cf3bda-aa8d-406a-a262-77ce9e9c0eb2.png)

mathnogueira:
Hello @abcl-trifork 

I'm glad to hear you are using Tracetest! I'll create an issue to make it possible for Tracetest to send multiple values using the same header name.

Regarding your considerations, we are working on a new way of specifying assertions and we expect to release it soon. It will allow you to achieve exactly what you wanted to do in your examples, but with a slightly different syntax. It will be released after #1378 is merged.

mathnogueira:
Using the new syntax, the assertions would look like this:

```yaml
# assert on count of specific header-key
  assertions:
  - http.request.header.accept_language | count = 2

[...]

PR #1377, opened 2022-10-13
feat: add count filter

mathnogueira:
This PR adds a new filter called `count` which returns the size of the input array. If it's a single value, it will return `1`. This was added to help with the use case described in #1372 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test


jorgeepc (approved):
🔥

schoren (approved):
LGTM!

PR #1385, opened 2022-10-13
chore: remove old e2e tests that were disabled for a while now

mathnogueira:
This PR removes our old e2e test suite for the CLI and removes prism. None of those things were in use for a couple of months.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
kill it!!

PR #1388, opened 2022-10-14
feat: support variables in expressions

mathnogueira:
This PR allow users to reference variables in expressions.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

### DIFF
--- a/docs/expressions.md
+++ b/docs/expressions.md
@@ -5,6 +5,7 @@ Tracetest allows you to add expressions when writing your tests. They are a nice
 ## **Features**
 
 * Reference span attributes
+* Reference variables
 * Arithmetic operations
 * String interpolation
 * Filters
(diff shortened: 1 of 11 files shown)


docs/expressions.md
```diff
@@ -17,6 +18,10 @@ When building assertions, you might need to assert if a certain span contains an
 attr:service.name = "cart-api"
 ```
 
+### **Reference Span Attributes**
```
schoren, line 21:
```suggestion
### **Reference Varaibles*
```

schoren (approved):
nit about docs, otherwise looks great!

docs/expressions.md
```diff
@@ -17,6 +18,10 @@ When building assertions, you might need to assert if a certain span contains an
...
+
+Tracetest allows you to create variables based on the trace obtained by the test to enable assertions that require values from other spans. Variables use the prefix `var:` and its name. For example, a variable called `user_id` would be referenced as `var:user_id` in an expression.
```
jfermi, line 23:
```suggestion
Create variables in Tracetest based on the trace obtained by the test to enable assertions that require values from other spans. Variables use the prefix `var:` and its name. For example, a variable called `user_id` would be referenced as `var:user_id` in an expression.
```

Issue #950, opened 2022-07-20
Provide more options for response assertions


mrwatson:
**Is your feature request related to a problem? Please describe.**
Our potential use case would require in-depth assertions on JSON responses returned from tests. Assertions against trace data is helpful, but does not provide the full picture in our case. For example, if we had a structure returned that looked like this:

```json
{
  "id": "abc",
  "revision": 3,
  "title": "Demo Protocol Version 1",
  "fullVersion": {
    "major": 3,
    "minor": 2
  },
  "nested": {
    "id": "def",
    "revision": 4,
    "nestedAgain": {
      "id": "klm",
      "revision": 5,
      "owner": {
        "id": "demo"
      }
    },
    "identifier": "yyy"
  }
}
```

We would want to be able to write assertions of the form:
- `$.id` is a string
- `$.revision` is a number
- `$.nested.id` != `$.id`
- `$.nested.revision` is a number and does not match `$.revision`
- `$.nested.nestedAgain.owner.id` is a string matching 'demo'
- `$.nested.identifier` is a string of at least 3 characters

**Describe the solution you'd like**
Provide additional assertion types matching the use cases listed above.

**Describe alternatives you've considered**
Running a test from the command line and passing the data through JSONPath or some such tool to validate the response. This is certainly doable, but more difficult.

**Additional context**
N/A

schoren:
I think we can add "filters" for attributes. For example

```
- tracetest.response.body | jsonpath $.identifier | length > 3 
- tracetest.response.body | jsonpath $.id | type = string
- tracetest.response.body | jsonpath $.nested.id != tracetest.response.body | jsonpath $.nested.id
```

This relates to https://github.com/kubeshop/tracetest/issues/782 too

kdhamric:
@mathnogueira I believe we can close this. Can you review and close if you believe we have this capability based on the jsonpath work that was done?

mathnogueira:
Yes, this was implemented back (probably v0.8). 

In that version, we build our assertion engine capable of retrieving data from a data source (environment variables, span attributes), applying fields to transform that data, and comparing it with another value.

The examples you used can be asserted as:

```python
# $.id is a string
attr:my_json | json_path '$.id' | type = "string"

# $.revision is a number
attr:my_json | json_path '$.revision' | type = "number"

# $.nested.id != $.id
attr:my_json | json_path '$.nested.id' != attr:my_json | json_path '$.id'

# $.nested.revision is a number and does not match $.revision
[...]

PR #1397, opened 2022-10-19
feat: add "type" filter and support arrays in expressions

mathnogueira:
This PR adds a new `type` filter and allows users to declare arrays to run comparisons.

This was based on #950. It is useful for checking if some field is a number or a string

### Motivation

1. Imagine a field can be either a `string` or a `number`. We don't have a way of writing an assertion like that. So instead, we could work with arrays to overcome this issue:

```yaml
# check if id is either a number or a string
["string", "number"] contains attr:my_json | json_path '.id' | type
```

2. Imagine you want to compare multiple values from a json_path query at once:

```yaml
# check if ids are 35 and 38 in this exact order
# in the future we can override the equal operator to make the order of the array irrelevant. But 
# I'm not adding this to this PR
attr:my_json | json_path '$.array[*].id' = [35, 38]
```

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

### DIFF
--- a/docs/expressions.md
+++ b/docs/expressions.md
@@ -88,14 +88,34 @@ You can select the last item from a list by specifying `'last'` as the argument 
 '{ "array": [1, 2, 3] }' | json_path '$.array[*]' | get_index 'last' = 3
 ```
 
-### **Count**
+### **Length**
 
 Return the size of the input array. If it's a single value, it will return 1. Otherwise it will return `length(input_array)`.
 
 ```
-'{ "array": [1, 2, 3] }' | json_path '$.array[*]' | count = 3
+'{ "array": [1, 2, 3] }' | json_path '$.array[*]' | length = 3
 ```
 
 ```
-"my string" | count = 1
+"my string" | length = 1
+```
+
+### **Type**
+
+Return the type of the input as a string.
+
+```
+2 | type = "number"
+```
+
+```
+2s | type = "duration"
+```
+
+```
+"something" | type = "string"
+```
+
+```
+[1, 2s, "this is a string"] | type = "array"
 ```
--- a/docs/expressions.md
+++ b/docs/expressions.md
@@ -119,3 +119,8 @@ Return the type of the input as a string.
 ```
 [1, 2s, "this is a string"] | type = "array"
 ```
+
+```
+# check if attribute is either a number of a string
+["number", "string"] contains attr:my_attribute | type
+```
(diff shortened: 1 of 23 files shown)


server/expression/executor.go
```diff
@@ -262,6 +285,26 @@ func (e Executor) executeFilter(input filters.Value, filter *Filter) (filters.Va
 	return newValue, nil
 }
 
+func (e Executor) convertArrayIntoFilterValue(input types.TypedValue) (filters.Value, error) {
+	var array Array
+	err := json.Unmarshal([]byte(input.Value), &array)
```
mathnogueira, line 290: By using JSON here, we don't lose the types of each value in the array when converting them to `filters.Value` later in the pipeline.

server/expression/executor_test.go
```diff
@@ -243,6 +243,53 @@ func TestFunctionExecution(t *testing.T) {
...
+			Name:       "incomplete_arrays_should_not_be_equal",
+			Query:      `[1,2,3] = [1, 2]`,
+			ShouldPass: false,
+		},
+		{
+			Name:       "arrays_can_be_compared_with_other_arrays_generated_by_filters",
+			Query:      `'{ "array": [{ "name": "john", "age": 37 }, { "name": "jonas", "age": 38 }]}' | json_path '$.array[*].age' = [37, 38]`,
+			ShouldPass: true,
+		},
+	}
```
mathnogueira, line 288: This test made me refactor the code to use `value.Value` instead of string.

server/expression/comparison.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func compare(comparatorName string, leftValue, rightValue value.Value) error {
+	comparatorFunction, err := comparator.DefaultRegistry().Get(comparatorName)
+	if err != nil {
+		return fmt.Errorf("comparator not supported: %w", err)
+	}
+
+	if leftValue.IsArray() && comparatorName == "contains" {
+		return compareArrayContains(leftValue, rightValue)
```
mathnogueira, line 17: This way we can use te `contains` operator to check if an array contains a certain value in it. Just like javascript's `includes` function.

docs/expressions.md
```diff
@@ -88,14 +88,39 @@ You can select the last item from a list by specifying `'last'` as the argument
...
+```
+
+### **Type**
+
+Return the type of the input as a string.
+
+```
+2 | type = "number"
+```
+
```
schoren, line 110: can we add something like `attr:myapp.operations | get_index 2 | type = string`?

schoren (approved):

PR #1403, opened 2022-10-20
feat: allow span[] as a valid selector

mathnogueira:
This PR enables users to define `span[]` as a valid selector to select all spans.

### Motivation

Without this, there's no way of selecting all spans that are descendent of a parent span. With this change, you can define it as `span[parent-attrs] span[]`

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/assertions/selectors/selector_test.go
+++ b/server/assertions/selectors/selector_test.go
@@ -109,6 +109,11 @@ func TestSelector(t *testing.T) {
 			ExpectedSpanIds: []trace.SpanID{updatePokemonDatabaseSpanID},
 		},
 		{
+			Name:            "Selector_to_select_all_children_spans",
+			Expression:      `span[service.name="Pokeshop" tracetest.span.type="http"] span[]`,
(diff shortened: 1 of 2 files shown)


xoscar (approved):

jorgeepc (approved):

server/assertions/selectors/selector_test.go
```diff
@@ -69,57 +69,67 @@ func TestSelector(t *testing.T) {
 		ExpectedSpanIds []trace.SpanID
 	}{
 		{
-			Name:            "Empty selector should select all spans",
+			Name:            "Empty_selector_should_select_all_spans",
```
schoren, line 72: do you mind making this `CamelCase`? it's more consistent with go's naming

schoren (approved):

server/assertions/selectors/selector_test.go
mathnogueira: we would have to change not just here, but most of our tests.

schoren: yea, I've been doing that in the tests I updated

PR #1422, opened 2022-10-24
fix: make failed assertions to return the formatted duration field

### LINKED ISSUE #1421: duration value is not formatted when assertion fails
**Describe the bug**
When a duration field is used in an assertion and the comparison fails, the observed value is presented as an integer and not as a duration field.

**Expected behavior**
It should show `169ms` instead of `169000000`

**Screenshots**
Screenshot from 2022-10-24 15-22-03

mathnogueira:
This PR fixes #1421

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [X] added a test

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -276,11 +276,6 @@ jobs:
           cd cli
           go build -o ./tracetest main.go
 
-      - name: Build Current CLI
-        run: |
-          cd cli
-          go build -o ./tracetest main.go
-
       - name: Install CLI intergation version (latest main)
         run: |
           export GOPATH=$(go env GOPATH)
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -298,7 +298,8 @@ jobs:
       - name: Run integration tests
         run: |
           cd tracetesting
-          TRACETEST_CLI_MAIN="$(go env GOPATH)/bin/tracetest" \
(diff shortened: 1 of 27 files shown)


server/expression/executor.go
```diff
@@ -60,10 +60,6 @@ func (e Executor) Statement(statement string) (string, string, error) {
 	}
 
 	err = compare(parsedStatement.Comparator, leftValue, rightValue)
-	if err != nil {
-		return leftValue.String(), rightValue.String(), err
-	}
-
```
mathnogueira, line 66: If you look at the next lines, you will see that this `err` will be returned after formatting the duration fields. So it's safe to remove this condition.

xoscar (approved):
LGTM!

jorgeepc (approved):

.github/workflows/pull-request.yaml
```diff
@@ -276,15 +276,10 @@ jobs:
...
-      - name: Build Current CLI
-        run: |
-          cd cli
-          go build -o ./tracetest main.go
-
       - name: Install CLI intergation version (latest main)
         run: |
           export GOPATH=$(go env GOPATH)
-          go install -v github.com/kubeshop/tracetest/cli@latest
+          go install -v github.com/kubeshop/tracetest/cli@main
```
mathnogueira, line 282: We were installing the latest released version instead of the latest main version.

Issue #1425, opened 2022-10-25
[Transactions] Implement Environment CRUD


xoscar:
Based on the schema definition provided in this issue https://github.com/kubeshop/tracetest/issues/1161

We must implement the environment CRUD services for the FE client and the CLI.
A new definition file type for Environments can be added as a test definition. 

Acceptance Criteria

**AC1**
As a frontend developer
I should be able to integrate environment CRUD services
With the UI

PR #1430, opened 2022-10-26
1425 adding basic environments CRUD

xoscar:
This PR adds the very basic environments CRUD to the backend services.
We still need to add the following things to complete the ticket requirements:
1. Include the definition files for environments
2. For test runs, the environment snapshot should be stored

## Changes

- Open API spec changes
- Models, repositories, and services to support Environments
- DB Migrations

## Fixes

- https://github.com/kubeshop/tracetest/issues/1425 (Partially)

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/model/tests.go
```diff
@@ -103,6 +103,9 @@ type (
 		Fail          int
 
 		Metadata RunMetadata
+
+		// environment
+		EnvSnapshot []EnvironmentValue
```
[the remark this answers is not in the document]
xoscar: Maybe I'm losing the plot haha so please keep me honest if you think that it doesn't makes sense 😄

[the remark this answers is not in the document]
mathnogueira: > An environment can be completely removed and recreated with the same name

I don't think that would be a problem, to be honest. We would still have the variables available at the moment of the run. As runs are immutable, that is more of a debug tool than anything else.

server/http/controller.go
```diff
@@ -483,6 +485,20 @@ func metadata(in *map[string]string) model.RunMetadata {
 	return model.RunMetadata(*in)
 }
 
+func environment(ctx context.Context, testDB model.Repository, environmentId string) model.Environment {
+	if environmentId != "" {
+		environment, err := testDB.GetEnvironment(ctx, environmentId)
+
+		if err != nil {
+			return model.Environment{}
+		}
```
mathnogueira, line 494: We have a lot of different errors that can happen here (invalid SQL syntax, connection errors, etc) that we need to handle. I'd suggest adding an error return as well, but that would totally kill the reason for this function to exist. Maybe that logic could be done in the `RunTest` function instead. So you can handle it using `handleDBError`

server/http/controller.go
```diff
@@ -514,3 +530,89 @@ func (c *controller) executeTest(ctx context.Context, test model.Test, runInfo o
...
+func (c *controller) DeleteEnvironment(ctx context.Context, environmentId string) (openapi.ImplResponse, error) {
+	environment, err := c.testDB.GetEnvironment(ctx, environmentId)
+	if err != nil {
+		return handleDBError(err), err
+	}
+
+	err = c.testDB.DeleteEnvironment(ctx, environment)
+	if err != nil {
+		return openapi.Response(http.StatusInternalServerError, err.Error()), err
+	}
```
mathnogueira, line 572: Maybe you should use `handleDBError` instead.

mathnogueira (commented):
Very good work, @xoscar! Just left a couple of comments regarding error handling.

server/model/tests.go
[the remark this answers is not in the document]
xoscar: This is done 😄 I added the environment json as snapshot to the test run table

server/http/controller.go
xoscar: I added the error handling, let me know if it makes sense

server/http/controller.go
xoscar: Done 😄

mathnogueira (approved):
LGTM. I really liked the db tests, they make it very easy to review the db code :D

PR #1431, opened 2022-10-26
[Transactions] Integrate Test Outputs

### LINKED ISSUE #1420: [Transactions] Integrate Test Outputs
**Objective:** integrate the frontend implementation of test outputs with the backend api.

**AC1:**
As a user in the trigger tab of a test run,
I want to create an output with the following fields:
- name
- selector
- value

**AC2:**
As a user in the trigger tab of a test run,
I want to edit an output and should be able to change the following fields:
- name
- selector
- value

**AC3:**
As a user in the trigger tab of a test run,
I want to delete an output

jorgeepc:
This PR integrates the frontend implementation of test outputs with the backend api.

## Changes

- Update the FE logic to support "draft" outputs
- New `/outputs` endpoint to set outputs to a test

## Fixes

- fixes #1420

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


web/src/redux/testOutputs/slice.ts
```diff
@@ -0,0 +1,43 @@
...
+const initialState: ITestOutputsState = {
+  initialOutputs: [],
+  outputs: [],
+};
+
+const testOutputsSlice = createSlice({
+  name: 'testOutputs',
+  initialState,
+  reducers: {
+    outputsInitiated(state, action: PayloadAction<TTestOutput[]>) {
```
jorgeepc, line 19: Redux strong recommendation: [Model Actions as Events, Not Setters](https://redux.js.org/style-guide/#model-actions-as-events-not-setters)

server/http/controller.go
```diff
@@ -277,6 +277,32 @@ func (c *controller) RunTest(ctx context.Context, testID string, runInformation
...
+	if err != nil {
+		return handleDBError(err), err
+	}
+
+	newOutputs, err := c.mappers.In.Outputs(testOutputs)
+	if err != nil {
+		return openapi.Response(http.StatusBadRequest, err.Error()), nil
+	}
+
+	newTest, err := model.BumpVersionIfOutputsChanged(test, newOutputs)
```
xoscar, line 291: Question, do we need to keep sending the full body when updating a test? keeping track of the outputs and such? if so, I think instead of having two separate services to keep track of the same data we should stick with only one

web/src/redux/testOutputs/selectors.ts
```diff
@@ -0,0 +1,41 @@
+import {createSelector} from '@reduxjs/toolkit';
```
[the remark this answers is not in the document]
xoscar: I like this!

web/src/redux/testOutputs/slice.ts
xoscar: Makes sense, but I think as this is just a preference I won't worry too much as it doesn't change the behavior of functionality

web/src/types/TestOutput.types.ts
```diff
@@ -0,0 +1,19 @@
+import {Model, TTestSchemas} from './Common.types';
+
+export type TRawTestOutput = TTestSchemas['TestOutput'];
+
+export type TTestOutput = {
```
xoscar, line 5: Do we still need this? I think it would be better to compose the type from the open API spec and overwrite the fields we need, that we can avoid changing it "that" often

xoscar (changes requested):
Hey @jorgeepc gj! I think we need to see if there is a way we can display the value to the user while creating the output and during execution as right now we only display the created metadata but not the result.

I also added a couple of comments to some of the changes. Let me know what you think my dude!

web/src/redux/testOutputs/slice.ts
jorgeepc: Yes, agree. However it could change the way you think about actions. Probably right now we are dispatching multiple actions for 1 single event. For example, we dispatch an action to set a value in the store but also to clean something else.

server/http/controller.go
jorgeepc: The `test update endpoint` updates everything. So, if we want to update only outputs we will need to send the trigger info, specs, etc. My idea here was to follow the same approach we have today for specs. We have a separate `PUT /definition endpoint` to set specs for a particular test.

web/src/types/TestOutput.types.ts
jorgeepc: Yeah, hmm that was my first approach. I can change that, I don't have a strong opinion about it. However not a big fan of using the `Model<>` type. I think it makes "not easy" to understand your types when you are developing.

jorgeepc:
Thanks @xoscar for the feedback. Added my responses to your comments. Regarding this:
> I think we need to see if there is a way we can display the value to the user while creating the output and during execution as right now we only display the created metadata but not the result.

I'm displaying the execution value of the output. We are getting that from the `Run` response. You can see that in the selector we are getting those values (`testRunOutputs`):
https://github.com/kubeshop/tracetest/blob/2d9083122bf33b199879f0e133c58ea8973311b3/web/src/redux/testOutputs/selectors.ts#L19-L28

The idea of displaying the value while creating the output is great. I was waiting for your `parse statement service` to see how we can reuse that.

server/http/controller.go
xoscar: Makes sense!

web/src/types/TestOutput.types.ts
xoscar: Not a blocker, let's discuss the `Model<>` approach to see what we can do for the future

xoscar (approved):
Gj @jorgeepc thanks for answering my comments 🦜

PR #1432, opened 2022-10-27
feat: transaction crud

mathnogueira:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/migrations/13_transactions.up.sql
```diff
@@ -0,0 +1,23 @@
...
+);
+
+CREATE TABLE transaction_steps (
+    transaction_id varchar NOT NULL,
+    transaction_version integer NOT NULL,
+    test_id varchar NOT NULL,
+    step_number int NOT NULL,
+
+    CONSTRAINT transaction_steps_transactions_fk
+      FOREIGN KEY (transaction_id, transaction_version) REFERENCES transactions(id, "version")
```
schoren, line 20: don't you need an fk here to the tests table?

tracetesting/definitions/transaction_create.yml
```diff
@@ -0,0 +1,32 @@
...
+      url: ${TARGET_URL}/api/transactions
+      method: POST
+      headers:
+        - key: Content-Type
+          value: application/json
+      body: |
+        {
+            "name": "test-transaction",
+            "description": "a transaction",
+            "steps": []
```
schoren, line 17: it would be better to have a test here

schoren (approved):
left 2 nits, but should be good to go!

tracetesting/definitions/transaction_create.yml
schoren: by test I mean step

server/migrations/13_transactions.up.sql
mathnogueira: Thing is that the PK of tests are ìd` and `version`. But as we discussed a couple of days ago, we are always going to use the latest version available. So I decided to not keep the fk here

api/openapi.yaml
```diff
@@ -34,6 +34,139 @@ paths:
                 $ref: "./definition.yaml#/components/schemas/ExecuteDefinitionResponse"
         422:
           description: "trying to create a test with an already existing ID"
+  /transactions:
+    get:
```
jorgeepc, line 38: Based on the UI design, we are going to need one single endpoint that returns `tests` and `transactions` (should support search, pagination, etc.). We could keep the `get /transactions` endpoint and implement the unified one in a new PR.

tracetesting/definitions/transaction_create.yml
mathnogueira: Done

server/testdb/postgres.go
```diff
@@ -107,7 +107,15 @@ func (td *postgresDB) ServerID() (id string, isNew bool, err error) {
 }
 
 func (td *postgresDB) Drop() error {
-	return dropTables(td, "test_runs", "tests", "server", "schema_migrations")
+	return dropTables(
+		td,
+		"transaction_steps",
+		"transactions",
+		"test_runs",
+		"tests",
```
xoscar, line 115: I forgot to add the environments table here, is it necessary?

xoscar (approved):
Hey @mathnogueira great job man!, I left some non-blocking comments!

server/testdb/postgres.go
mathnogueira: yes, it is. I'll add it

server/http/mappings/tests.go
```diff
@@ -282,6 +298,27 @@ func (m OpenAPI) Runs(in []model.Run) []openapi.TestRun {
 type Model struct {
 	comparators           comparator.Registry
 	traceConversionConfig traces.ConversionConfig
+	testRepository        model.TestRepository
+}
+
+func (m Model) Transaction(ctx context.Context, in openapi.Transaction) (model.Transaction, error) {
+	tests := make([]model.Test, len(in.Steps))
+	for i, testID := range in.Steps {
```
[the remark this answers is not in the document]
mathnogueira: I'll try to solve this in another PR

Issue #1434, opened 2022-10-27
[Transactions] Environment definition execution


xoscar:
Enable environment definition files to be executed from the server side from outside requests.

PR #1435, opened 2022-10-28
1434 [Transactions] Environment definition execution

xoscar:
This PR adds the definition yaml execution support for environment files, same logic as for tests, if the id exits we update, if not we create a new brand entry to the DB.

## Changes

- Adds execution support for environment definition files

## Fixes

- #1434 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/78716627311d45209f14214d6cb0c961 [NO SOUND]

### DIFF
--- a/server/http/controller.go
+++ b/server/http/controller.go
@@ -484,7 +484,36 @@ func (c *controller) ExecuteDefinition(ctx context.Context, testDefinition opena
 		return c.executeTest(ctx, test.Model(), testDefinition.RunInformation)
 	}
 
+	if environment, err := def.Environment(); err == nil {
+		return c.createEnvFromDefinition(ctx, environment)
+	}
+
 	return openapi.Response(http.StatusUnprocessableEntity, nil), nil
+}
+
+func (c *controller) createEnvFromDefinition(ctx context.Context, def yaml.Environment) (openapi.ImplResponse, error) {
+	environment := def.Model()
+
+	if environment.ID != "" {
+		_, err := c.testDB.UpdateEnvironment(ctx, environment)
+
+		if err != nil {
+			return handleDBError(err), err
+		}
+	} else {
+		env, err := c.testDB.CreateEnvironment(ctx, environment)
+		environment.ID = env.ID
+
+		if err != nil {
+			return handleDBError(err), err
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


jorgeepc (approved):
🔥 LGTM!

server/model/yaml/environment.go
```diff
@@ -0,0 +1,52 @@
...
+			Key:   ev.Key,
+			Value: ev.Value,
+		})
+	}
+
+	return mevs
+}
+
+func (e Environment) Validate() error {
+	if e.Name == "" {
```
mathnogueira, line 36: Shouldn't we validate if the variables names are set?

server/http/controller.go
```diff
@@ -484,9 +484,38 @@ func (c *controller) ExecuteDefinition(ctx context.Context, testDefinition opena
 		return c.executeTest(ctx, test.Model(), testDefinition.RunInformation)
 	}
 
+	if environment, err := def.Environment(); err == nil {
+		return c.createEnvFromDefinition(ctx, environment)
```
mathnogueira, line 488: This is returning the wrong response. It should return the testId, not te env ID.

mathnogueira (changes requested):

server/http/controller.go
xoscar: I'm a little confused here, when creating an environment from a definition file there won't be any info regarding the test so my idea was to return the environment id as the creation output, if that's not expected, do you think I should add a new field to the response to be environment id instead?

server/model/yaml/environment.go
xoscar: Done 😄

server/http/controller.go
mathnogueira: Maybe this shouldn't be here then. Maybe the CLI should create the env before running the test and we only reference the environment here. Another thing, even if we leave this here, shouldn't this happen before the `executeTest` function call?

To execute a test we still need the transaction to exist.

mathnogueira: Oh I see what you did there. It is only be executed if the resource is an environment. Ignore my comments. Makes sense.

mathnogueira (approved):
Looks good

PR #1465, opened 2022-11-08
feat: transaction run crud and execution

mathnogueira:
This PR adds endpoints for handling transaction runs and to execute a transaction

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/executor/transaction_runner.go
```diff
@@ -0,0 +1,160 @@
...
+func (r PersistentTransactionRunner) runTransactionStep(ctx context.Context, transactionRun model.TransactionRun, stepIndex int, environment model.Environment) (model.TransactionRun, error) {
+	step := transactionRun.Steps[stepIndex]
+	testRun := r.testRunner.Run(ctx, step, transactionRun.Metadata, environment)
+
+	testRun, err := r.waitTestRunIsFinished(ctx, testRun)
+	if err != nil {
+		return model.TransactionRun{}, fmt.Errorf("could not wait for step execution: %w", err)
+	}
+
+	transactionRun.StepRuns = append(transactionRun.StepRuns, testRun)
```
schoren, line 111: runs tend to be really big, since they hold the trace. Is it really necessary to have them in memory like this? you could hold a `map[testRunID]status` or something similar holding a small subset of the info if it's required somewhere else (which I don't think it is, except for transmiting it to the FE)

server/executor/transaction_runner.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+	}
+
+	return testRun, nil
+}
+
+func (r PersistentTransactionRunner) patchEnvironment(baseEnvironment model.Environment, run model.TransactionRun) model.Environment {
+	newEnvVariables := make([]model.EnvironmentValue, 0)
+
+	for i := 0; i < run.CurrentTest; i++ {
```
schoren, line 145: I think this loop is not required. Each step already contains the env from the previous one, so you can just apply the outputs from the last step

server/executor/transaction_runner.go
```diff
@@ -0,0 +1,160 @@
...
+
+	var err error
+
+	for i := range run.Steps {
+		run, err = r.runTransactionStep(ctx, run, i, environment)
+		if err != nil {
+			return fmt.Errorf("could not execute step %d of transaction %s: %w", i, run.TransactionID, err)
+		}
+
+		if run.State == model.TransactionRunStateExecuting {
```
schoren, line 90: Why do you need this condition? I imagine it's to prevent updating the env on error. In that case, I'd use an exit guard, like 
```
if run.state == error {
  break
}
environment = r.patchEnvironment(environment, run) 
```

The current implementation might actually lead to a bug, or unexpected behaviour, if the last step returns a state  other than `exectuing`, where the environments wouldn't be correctly replaced. Might not be a case now, but might happen later

server/model/environments.go
```diff
@@ -0,0 +1,47 @@
...
+	"time"
+)
+
+type (
+	Environment struct {
+		ID          string
+		Name        string
+		Description string
+		CreatedAt   time.Time
+		Values      []EnvironmentValue
```
schoren, line 14: shouldn't this be a `map[string]string`? maybe an `OrderedMap`?

server/model/environments.go
```diff
@@ -0,0 +1,47 @@
...
+		Value string
+	}
+)
+
+func (e Environment) GetSlug() string {
+	return strings.ToLower(strings.ReplaceAll(strings.TrimSpace(e.Name), " ", "-"))
+}
+
+func (e Environment) Merge(env Environment) Environment {
+	values := make(map[string]string)
```
schoren, line 28: I guess the idea of this is to avoid duplicated env keys. but in doing this, it's loosing the order guaranteed in a slice, which I imagine is the point of `e.Values` being a slice.

The only way to keep order and uniqueness of a map is using `OrderedMap`, even if the code looks a bit to verbose

server/model/environments_test.go
```diff
@@ -0,0 +1,43 @@
...
+	assert.Equal(t, env1.ID, newEnv.ID)
+	assert.Equal(t, env1.Name, newEnv.Name)
+	assert.Equal(t, env1.Description, newEnv.Description)
+	assert.Equal(t, env1.CreatedAt, newEnv.CreatedAt)
+
+	require.Len(t, newEnv.Values, 4)
+	assert.Contains(t, newEnv.Values, model.EnvironmentValue{Key: "URL", Value: "http://localhost"})
+	assert.Contains(t, newEnv.Values, model.EnvironmentValue{Key: "PORT", Value: "8085"})
+	assert.Contains(t, newEnv.Values, model.EnvironmentValue{Key: "apiKey", Value: "abcdef"})
+	assert.Contains(t, newEnv.Values, model.EnvironmentValue{Key: "apiKeyLocation", Value: "header"})
```
schoren, line 42: it would be nice to have an override key in this test

schoren: Also, if the idea of having an array is to keep the order, it would be great to have a test to guarantee the ordering

server/testdb/transactions_test.go
```diff
@@ -233,82 +233,3 @@ func TestGetTransactionsWithMultipleVersions(t *testing.T) {
 		assert.Equal(t, 2, test.Version)
 	}
 }
-
-// func TestTransactionSummary(t *testing.T) {
```
schoren, line 237: can all this commented code be deleted?

schoren (changes requested):
The basic idea of the runner looks fine, but there are a few things that needs to be fixed for the app to keep consistent behaviour (such as order of things) and performance

server/model/transactions.go
```diff
@@ -16,8 +16,57 @@ type (
...
+)
+
+const (
+	TransactionRunStateCreated   TransactionRunState = "CREATED"
+	TransactionRunStateExecuting TransactionRunState = "EXECUTING"
+	TransactionRunStateFailed    TransactionRunState = "FAILED"
+	TransactionRunStateFinished  TransactionRunState = "FINISHED"
 )
 
+func (rs TransactionRunState) IsFinal() bool {
```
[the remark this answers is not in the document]
mathnogueira: I prefer to have a different state because some states in the test state doesn't make sense for transactions, such as `AWAITING_TRACE`

server/executor/transaction_runner.go
```diff
@@ -0,0 +1,160 @@
...
+			environment = r.patchEnvironment(environment, run)
+		}
+	}
+
+	if !run.State.IsFinal() {
```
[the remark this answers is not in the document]
mathnogueira: You are correct. Instead of creating an `IsError` function, I'm going to just compare it with the failed value.

server/model/environments.go
```diff
@@ -0,0 +1,47 @@
...
+		Values      []EnvironmentValue
+	}
+
+	EnvironmentValue struct {
+		Key   string
+		Value string
+	}
+)
+
+func (e Environment) GetSlug() string {
```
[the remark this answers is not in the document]
mathnogueira: All environment changes are only things I moved to this file. They were introduced in #1430

server/model/environments.go
mathnogueira: The order of environments doesn't matter, so no need for OrderedMap. `map[string]string` might be a better alternative than an array.

server/model/environments.go
mathnogueira: yeah, the order doesn't matter in this case. I'll change it to `map[string]string` in a different PR

server/testdb/transactions_test.go
mathnogueira: I forgot to remove it in my old PR. We don't have transaction summaries yet, so I decided to remove it and reintroduce it later.

server/executor/transaction_runner.go
```diff
@@ -0,0 +1,160 @@
...
+	err = r.db.UpdateTransactionRun(ctx, transactionRun)
+	if err != nil {
+		return model.TransactionRun{}, fmt.Errorf("could not update transaction run: %w", err)
+	}
+
+	return transactionRun, nil
+}
+
+func (r PersistentTransactionRunner) waitTestRunIsFinished(ctx context.Context, testRun model.Run) (model.Run, error) {
+	ticker := time.NewTicker(r.checkTestStateDelay)
```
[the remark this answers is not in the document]
mathnogueira: Implemented it.

server/testdb/transaction_runs.go
```diff
@@ -0,0 +1,329 @@
...
+	return strings.ReplaceAll(sql, runSequenceName, seqName)
+}
+
+func (td *postgresDB) CreateTransactionRun(ctx context.Context, transactionRun model.TransactionRun) (model.TransactionRun, error) {
+	jsonSteps, err := json.Marshal(transactionRun.Steps)
+	if err != nil {
+		return model.TransactionRun{}, fmt.Errorf("failed to marshal transaction steps: %w", err)
+	}
+
+	jsonStepRuns, err := json.Marshal(transactionRun.StepRuns)
```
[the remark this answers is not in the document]
mathnogueira: Now it keeps simpler objects containing only the essentials for running the transaction

server/model/transactions.go
```diff
@@ -16,8 +16,76 @@ type (
 		Steps       []Test
 		Summary     Summary
 	}
+
+	TransactionRunState string
+
+	transactionStep struct {
```
schoren, line 22:
```suggestion
	TransactionStep struct {
```

server/model/transactions.go
```diff
@@ -16,8 +16,76 @@ type (
...
+func (rs TransactionRunState) IsFinal() bool {
+	return rs == TransactionRunStateFailed || rs == TransactionRunStateFinished
+}
+
 func (t Transaction) HasID() bool {
 	return t.ID != ""
 }
+
+func NewTransactionRun(transaction Transaction) TransactionRun {
+	testIds := make([]transactionStep, 0, len(transaction.Steps))
```
schoren, line 76:
```suggestion
	testIds := make([]TransactionStep, 0, len(transaction.Steps))
```

server/model/transactions.go
```diff
@@ -16,8 +16,76 @@ type (
...
+
+	for _, test := range transaction.Steps {
+		testIds = append(testIds, transactionStep{ID: test.ID, Name: test.Name})
```
schoren, line 79:
```suggestion
		testIds = append(testIds, TransactionStep{ID: test.ID, Name: test.Name})
```

schoren (approved):
Looks awesome! thanks for addressing everything!

xoscar (approved):

PR #1468, opened 2022-11-10
feat: validate cli config

mathnogueira:
This PR makes the CLI validate if it is configured before running any command.

It also makes the CLI logger print logs as messages and not JSON messages. This doesn't affect test outputs or other formatted outputs. Example:

Screenshot from 2022-11-10 18-37-47

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/cmd/config.go
+++ b/cli/cmd/config.go
@@ -20,6 +20,7 @@ func setupCommand(cmd *cobra.Command, args []string) {
 	setupOutputFormat()
 	setupLogger(cmd, args)
 	loadConfig(cmd, args)
+	validateConfig(cmd, args)
 	analytics.Init(cliConfig)
 }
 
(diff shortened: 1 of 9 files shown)


mathnogueira:
Question: should we skip this step for some commands? Such as `tracetest server install`?

jorgeepc (approved):

schoren:
@mathnogueira `tracetest server install` should not be validated, because you might be running the tool for the 1st time to setup the server.

Maybe we could add, on top of validation, a default value of `http://loaclhost:11633` that users can later override?

mathnogueira:
Added a config for skipping the config validation

schoren (approved):
LGTM!

PR #1484, opened 2022-11-15
Add Flame Graph as a Trace view

cescoferraro:
This PR add a new trace view called Flame view.
We use the same library jager is using called @pyroscope/flamegraph
So we can benefit from future development for the specific trace visualization goal like this [one](https://github.com/pyroscope-io/pyroscope/pull/1679) 

Screen Shot 2022-11-14 at 23 59 43

Screen Shot 2022-11-14 at 23 57 36

## Changes

- Add new Trace view

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/web/src/components/RunDetailTest/Visualization.tsx
+++ b/web/src/components/RunDetailTest/Visualization.tsx
@@ -1,6 +1,3 @@
-import {useCallback, useEffect} from 'react';
-import {Node, NodeChange} from 'react-flow-renderer';
-
 import {VisualizationType} from 'components/RunDetailTrace/RunDetailTrace';
 import SkeletonDiagram from 'components/SkeletonDiagram';
 import {useTestSpecForm} from 'components/TestSpecForm/TestSpecForm.provider';
@@ -8,6 +5,8 @@ import DAG from 'components/Visualization/components/DAG';
 import Timeline from 'components/Visualization/components/Timeline';
 import {TestState} from 'constants/TestRun.constants';
 import {useSpan} from 'providers/Span/Span.provider';
+import {useCallback, useEffect} from 'react';
+import {Node, NodeChange} from 'react-flow-renderer';
 import {useAppDispatch, useAppSelector} from 'redux/hooks';
 import {initNodes, onNodesChange as onNodesChangeAction} from 'redux/slices/DAG.slice';
 import DAGSelectors from 'selectors/DAG.selectors';
@@ -15,6 +14,7 @@ import TraceAnalyticsService from 'services/Analytics/TestRunAnalytics.service';
 import TraceDiagramAnalyticsService from 'services/Analytics/TraceDiagramAnalytics.service';
 import {TSpan} from 'types/Span.types';
 import {TTestRunState} from 'types/TestRun.types';
+import {Flame} from '../Visualization/components/Flame/Flame';
 
 export interface IProps {
   runState: TTestRunState;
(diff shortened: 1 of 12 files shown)


olha23:
@cescoferraro can we use the same color here as we use in graph view or timeline view?

cescoferraro:
@olha23 It's not trivial. We are using a proprietary library from pyroscope which is the same one Jager uses to avoid the issues with concurrent spans. So there are no many controls over the color. I will try to do it with raw CSS

cescoferraro:
@olha23 figured out how to roll our own version of the graph library

these are all the colors it needs to create a palette. Can you help me?
Screen Shot 2022-11-16 at 11 50 46

Rperry2174:
This is pretty cool! I'm curious did you make a conscious decision to not include the toolbar?

image

web/src/components/RunDetailTest/Visualization.tsx
```diff
@@ -70,26 +70,43 @@ const Visualization = ({runState, spans, type}: IProps) => {
...
+          matchedSpans={matchedSpans}
+          nodes={nodes}
+          onNavigateToSpan={onNavigateToSpan}
+          onNodesChange={onNodesChange}
+          onNodeClick={onNodeClick}
+          selectedSpan={focusedSpan}
+        />
+      )}
+
+      {type === VisualizationType.Timeline && (
```
xoscar, line 88: In a follow up version, lets consider moving this to a map instead of the ifs

web/src/components/RunDetailTrace/Flame/Flame.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const Flame: React.FC<IProps> = ({spans}) => {
+  const {run} = useTestRun();
+  const profile = useMemo(
+    () => convertJaegerTraceToProfile(convertTracetestSpanToJaeger(run.trace?.traceId || '', spans)),
+    [spans, run.trace?.traceId]
+  );
+  return (
+    <div style={{padding: 24}}>
```
xoscar, line 25: Styled component?

web/src/components/RunDetailTrace/Flame/convertTracetestSpanToJaeger.tsx
```diff
@@ -0,0 +1,41 @@
+import {Trace} from '@pyroscope/models';
+import {TSpan} from 'types/Span.types';
+
+type Span = Trace['spans'][number];
+
```
xoscar, line 5: Why does this have a `.jsx` extension?

web/src/components/RunDetailTrace/Flame/convertTracetestSpanToJaeger.tsx
```diff
@@ -0,0 +1,41 @@
...
+export function tracetestSpanToJagerSpan(id: string, spans: TSpan[], s: TSpan): Span {
+  const milliseconds = 1000;
+  return {
+    operationName: s.name,
```
xoscar, line 9: Let's create a model and put this under a service

xoscar (changes requested):
Hey @cescoferraro GJ I left some questions and concerns. Let me know what you think!

web/src/components/RunDetailTest/Visualization.tsx
cescoferraro: to do this, I would need to ensure all Graphs have the same props.
I don't think it's very relevant for this. But I can send a separate PR to deal with this. @xoscar

web/src/components/RunDetailTrace/Flame/Flame.tsx
```diff
@@ -0,0 +1,29 @@
...
+  spans: TSpan[];
+  width?: number;
+  onNavigateToSpan(spanId: string): void;
+  onNodeClick(spanId: string): void;
+}
+
+export const Flame: React.FC<IProps> = ({spans}) => {
+  const {run} = useTestRun();
+  const profile = useMemo(
+    () => convertJaegerTraceToProfile(convertTracetestSpanToJaeger(run.trace?.traceId || '', spans)),
```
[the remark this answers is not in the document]
cescoferraro: because the graph library expects a trace in the Jager format, so we need to emulate this format based on Tracetest spans

xoscar (approved):

PR #1489, opened 2022-11-15
fix: env merging and output processing

mathnogueira:
This PR fixes the output processor and environment merging when running transactions.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/executor/outputs_processor.go
+++ b/server/executor/outputs_processor.go
@@ -66,7 +66,7 @@ func outputProcessor(ctx context.Context, outputs model.OrderedMap[string, model
 		value := ""
 		spans.
 			ForEach(func(_ int, span traces.Span) bool {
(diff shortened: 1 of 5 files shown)


server/executor/outputs_processor.go
```diff
@@ -66,7 +66,7 @@ func outputProcessor(ctx context.Context, outputs model.OrderedMap[string, model
 		value := ""
 		spans.
 			ForEach(func(_ int, span traces.Span) bool {
-				value = extractAttr(traces.Span{}, stores, out.expr)
+				value = extractAttr(span, stores, out.expr)
```
mathnogueira, line 69: I tried creating an output, but it was always coming as empty. Then I noticed we were passing an empty span instead of using the one selected by the selector.

schoren (approved):

PR #1492, opened 2022-11-15
fix execution time calc

schoren:
This PR fixes the execution time calculation when test is running

## Changes

-

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


xoscar (approved):
LGTM!

server/model/run_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const ms = 100000000
```
mathnogueira, line 11: Cant you use duration.Microsecond instead?

mathnogueira (approved):
Just a small comment

PR #1494, opened 2022-11-15
1491 transactions websockets and versioning for transactions

xoscar:
This PR enables the websockets usages from the FE as part of the transaction run page.

## Changes

- Listens to events during the transaction run execution.

## Fixes

- [[Transactions] Websockets and Versioning for Transactions#1491](https://github.com/kubeshop/tracetest/issues/1491)

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/http/websocket/subscribe.go
+++ b/server/http/websocket/subscribe.go
@@ -62,6 +62,11 @@ func (e subscribeCommandExecutor) ResourceUpdatedEvent(resource interface{}) Eve
 		mapped = e.mappers.Out.Run(&v)
 	case *model.Run:
 		mapped = e.mappers.Out.Run(v)
+	case model.TransactionRun:
+		mapped = e.mappers.Out.TransactionRun(v)
+	case *model.TransactionRun:
+		var value model.TransactionRun = *v
+		mapped = e.mappers.Out.TransactionRun(value)
 	default:
 		fmt.Printf("type %T mapping not supported\n", v)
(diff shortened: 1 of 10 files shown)


mathnogueira (approved):
BE looks fine

server/http/websocket/subscribe.go
```diff
@@ -62,6 +62,11 @@ func (e subscribeCommandExecutor) ResourceUpdatedEvent(resource interface{}) Eve
 		mapped = e.mappers.Out.Run(&v)
 	case *model.Run:
 		mapped = e.mappers.Out.Run(v)
+	case model.TransactionRun:
+		mapped = e.mappers.Out.TransactionRun(v)
+	case *model.TransactionRun:
+		var value model.TransactionRun = *v
+		mapped = e.mappers.Out.TransactionRun(value)
```
mathnogueira, line 69: Can you use this instead?

```suggestion
		mapped = e.mappers.Out.TransactionRun( *v)
```

xoscar: I tried that but the TransactionRun function expects a model instance value not a pointer 😢

schoren: Are you sure you tried the exact code as Matheus suggested? It’s the exact same way we’ve done it for test runs, so it should work

Issue #1503, opened 2022-11-16
[Transactions] Allow users create outputs from attributes


xoscar:
We need a way to allow users fast create outputs based on span attributes

Acceptance Criteria

**AC1**
As a user looking at the test run trace mode
And I open the attributes drawer for any span
And by clicking on the three dots
I should be able to see a create output option

**AC2**
As a user looking at the test run test mode
And I open the attributes drawer for any span
And by clicking on the three dots
I should be able to see a create output option

**AC3**
As a user looking a the create output option from the attribute drawer
After clicking on it
I should be redirected to the outputs section
And the output modal should be displayed with the initial values from the selected attribute and span

PR #1508, opened 2022-11-16
1503 [Transactions] Allow users create outputs from attributes

xoscar:
This PR adds the ability for users to create outputs based on span attributes by clicking one of the options that show up when hovering over one of the attributes.

## Changes

- Moves output management to a provider
- Opens output modal and redirects from the new attribute option
- Trigger tabs are now accessible through the query params

## Fixes

- #1503 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/b5d4bffd7043474cbcf6d21129dee00c


web/src/components/RunDetailLayout/RunDetailLayout.tsx
```diff
@@ -42,29 +43,31 @@ const RunDetailLayout = ({test: {id, name, trigger, version = 1}, test}: IProps)
...
-        </Tabs.TabPane>
-        <Tabs.TabPane tab="Trace" key={RunDetailModes.TRACE}>
-          <RunDetailTrace run={run} testId={id} />
-        </Tabs.TabPane>
-        <Tabs.TabPane tab="Test" key={RunDetailModes.TEST}>
-          <RunDetailTest run={run} testId={id} />
-        </Tabs.TabPane>
-      </Tabs>
-    </S.Container>
+    <TestOutputProvider testId={id} runId={run.id}>
```
jorgeepc, line 46: It seems that we have all the providers for the TestRun page here `/src/pages/RunDetail/RunDetail.tsx`. Any reason why not having the same pattern for this provider?

jorgeepc (approved):
Looks great, one minor question.

web/src/components/RunDetailLayout/RunDetailLayout.tsx
xoscar: The only reasoning was to avoid having to render it before we actually need it, I'll move it up just so it matches with the rest of the stuff

PR #1513, opened 2022-11-16
fix step duplication in transactions

mathnogueira:
This PR fixes the bug where transactions would have duplicated steps when a new test version was created.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/testdb/tests.go
+++ b/server/testdb/tests.go
@@ -173,9 +173,8 @@ const (
 	testMaxVersionQuery = `
 	INNER JOIN (
 		SELECT id as idx, max(version) as latest_version FROM tests GROUP BY idx
-		) as latest_tests ON latest_tests.idx = t.id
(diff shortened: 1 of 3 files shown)


schoren (approved):

server/testdb/tests.go
```diff
@@ -173,9 +173,8 @@ const (
 	testMaxVersionQuery = `
 	INNER JOIN (
 		SELECT id as idx, max(version) as latest_version FROM tests GROUP BY idx
-		) as latest_tests ON latest_tests.idx = t.id
-
-		WHERE t.version = latest_tests.latest_version `
```
mathnogueira, line 178: Basically moving the condition from the `WHERE` to the `ON` clause.

PR #1528, opened 2022-11-17
feat: support running transactions from cli

schoren:
This PR allos runnign transactions from CLI. 
If a transaction step starts with `./`, it is treated as a file reference, and so is upserted in the server and replaced by the corresponding ID before executing the transaction.

No support has been added for `wait results` because that part needs to be reworked a bit before being practical for waiting for multiple steps.

Example Output:
Screenshot 2022-11-17 at 17 27 44

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/http/controller.go
```diff
@@ -783,13 +822,36 @@ func (c *controller) buildDataStores(ctx context.Context, info openapi.ResolveRe
...
 	}
 
-	createdTransaction, err := c.testDB.CreateTransaction(ctx, transactionModel)
+	return c.doCreateTransaction(ctx, transaction)
+}
+
+var errTransactionExists = errors.New("transaction already exists")
+
+func (c *controller) doCreateTransaction(ctx context.Context, transaction model.Transaction) (openapi.ImplResponse, error) {
+	// if they try to create a test with preset ID, we need to make sure that ID doesn't exists already
```
mathnogueira, line 837:
```suggestion
	// if they try to create a transaction with preset ID, we need to make sure that ID doesn't exists already
```

mathnogueira (approved):
Good job. Just found a comment you forgot to update.

PR #1537, opened 2022-11-18
feat(cli): add `--environment` flag

schoren:
This PR adds the `--environment` or `-e` flag. This allows to pass an environment ID to be used when running the test/transaction.

## Example using ID: 
Screenshot 2022-11-18 at 12 47 51

This will use the environment with ID `localhost`

## Example using file path:
Screenshot 2022-11-18 at 14 04 26

This will create (or update) an env with name/id of `localhost.env` (base path) and run the test/transaction using that id

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/actions/run_test_action.go
+++ b/cli/actions/run_test_action.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	cienvironment "github.com/cucumber/ci-environment/go"
+	"github.com/joho/godotenv"
 	"github.com/kubeshop/tracetest/cli/config"
 	"github.com/kubeshop/tracetest/cli/file"
 	"github.com/kubeshop/tracetest/cli/formatters"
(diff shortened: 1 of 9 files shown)


cli/actions/run_test_action.go
```diff
@@ -54,28 +57,96 @@ func (a runTestAction) Run(ctx context.Context, args RunTestConfig) error {
...
+func stringReferencesFile(path string) bool {
+	return strings.HasPrefix(path, "./")
+}
+
+func (a runTestAction) processEnv(ctx context.Context, envID string) (string, error) {
+	if !stringReferencesFile(envID) { //not a file, do nothing
+		return envID, nil
+	}
+
+	envVars, err := godotenv.Read(envID)
```
mathnogueira, line 98: I'd expect the environment file to be in the following format and not an .env file:

```yaml
---
type: Environment
id: something
name: something
description: an env
values:
  - key: HOST
    value: localhost
  - key: PORT
    value: 1245

File: cli/actions/run_test_action.go [hunk 2]
```diff
@@ -54,28 +57,96 @@ func (a runTestAction) Run(ctx context.Context, args RunTestConfig) error {
...
+	name := filepath.Base(envID)
+
+	req := openapi.Environment{
+		Id:     &name,
+		Name:   &name,
+		Values: values,
+	}
+
+	body, resp, err := a.client.ApiApi.
+		CreateEnvironment(ctx).
```
mathnogueira, line 120: If you read an `openapi.Environment`, you can create/update the env using `ExecuteDefinition` instead.

mathnogueira (changes requested):
Comments about the file format.

mathnogueira (approved):
After our discussion, this was the chosen env file format. So, looks good!

Issue #1534, opened 2022-11-18
Fix adding otput flow


olha23:
https://www.loom.com/share/0adad34c4a6a4c3680529ff171763ff2

PR #1538, opened 2022-11-18
fix(frontend): fixing output selector input

xoscar:
This PR fixes an issue we have with the creative output modal where the selector field doesn't work as expected.

## Changes

- Moving span provider to wrap the output modal provider.

## Fixes

- [Fix adding otput flow#1534](https://github.com/kubeshop/tracetest/issues/1534)

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


jorgeepc (approved):
Looks good. A suggestion for the commit. I don't think we need to include the issue number.
`fix(frontend): fix output selector input`
What do you think?

PR #1547, opened 2022-11-21
feat: allow tests to reference outputs declared in them

mathnogueira:
This PR fixes two problems with outputs:

1. Right now it's not possible to export output and use it within the same test. This fixes this limitation by injecting the outputs into the environment before running the test assertions;
2. When updating a test that contains outputs, the new version of the test doesn't contain those outputs. 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
LGTM!

jorgeepc (approved):

Issue #1544, opened 2022-11-21
Improve graph swisher on the flame view


olha23:
https://www.loom.com/share/e7de61fe69634df187b436165c02b157

olha23:
@cescoferraro tagged you here, because you worked on it

cescoferraro:
@olha23 sure! working on it right now!

PR #1549, opened 2022-11-21
Fix flamegraph alignment

cescoferraro:
This PR fix flamegraph alignment reported on #1544

- Flame graph container was not accounting to the fact that the default search bar is not present on the flame trace view
- Fix max-width issue on test view

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/components/Visualization/components/Flame/Flame.styled.tsx
+++ b/web/src/components/Visualization/components/Flame/Flame.styled.tsx
@@ -1,5 +1,6 @@
 import styled from 'styled-components';
 
-export const Container = styled.div`
+export const Container = styled.div<{$isTrace: boolean}>`
   padding: 24px;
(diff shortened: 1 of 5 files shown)


web/src/components/Visualization/components/Flame/Flame.styled.tsx
```diff
@@ -1,5 +1,6 @@
 import styled from 'styled-components';
 
-export const Container = styled.div`
+export const Container = styled.div<{$isTrace: boolean}>`
   padding: 24px;
+  max-width: ${({$isTrace}) => ($isTrace ? 'unset' : '60vw')};
```
jorgeepc, line 5: Why `60vw`? is it going to work fine in different screens?

jorgeepc (approved):
Looks good. Added one question.

PR #1559, opened 2022-11-22
docs: add concepts for transactions and environments

mathnogueira:
This PR adds concepts that are new in 0.8

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


docs/concepts/environments.md
```diff
@@ -0,0 +1,46 @@
+# Environments
+
+A common use case for tests is to assert the same behavior across multiple environments (dev, staging, and production, for example). But to make sure all environments will have the same behavior, it's important that the tests executed against those environments test the same aspects. To reduce risks of diverging tests, Tracetest allows you to organize different environments configurations using global objects called Environments.
```
jfermi, line 3:
```suggestion
A common use case for tests is to assert the same behavior across multiple environments (dev, staging, and production, for example). To make sure all environments will have the same behavior, it is important that the tests executed against those environments test the same aspects. To reduce the risks of diverging tests, Tracetest allows you to organize different environments configurations using global objects called **Environments**.
```

docs/concepts/environments.md
```diff
@@ -0,0 +1,46 @@
...
+
+## How environments work
```
jfermi, line 5:
```suggestion
## **How Environments Work**
```

docs/concepts/environments.md
```diff
@@ -0,0 +1,46 @@
...
+
+Environments are objects containing variables that can be referenced by tests. You can have the same test and just swap the information for each environment by changing which environment object to be used to run the test. To illustrate it, consider an app that is deployed in three stages: `dev`, `staging`, and `production`. We can execute the same test in all those environments, however, both `URL` and `credentials` change from environment to environment. To be able to run the same test against the three deployments of the app, you can create three environments:
```
jfermi, line 7:
```suggestion
Environments are objects containing variables that can be referenced by tests. You can use a single test and provide the information on which environment object will be used to run the test. To illustrate this, consider an app that is deployed in three stages: `dev`, `staging`, and `production`. We can execute the same test in all those environments, however, both `URL` and `credentials` change from environment to environment. To run the same test against the three deployments of the app, you can create three environments:
```

docs/concepts/environments.md
```diff
@@ -0,0 +1,46 @@
...
+        url: "${env:URL}/api/users"
+        method: POST
+        body: '{}'
+        authentication:
+          type: bearer
+          bearer:
+            token: "${env:API_TOKEN}"
+```
+
+Both `env:URL` and `env:API_TOKEN` would be replaced by the variables defined in the chosen environment to be used when running the test. So, if the chosen environment was `dev.env`, its values would be replaced by `https://app-dev.com` and `dev-key` respectively.
```
jfermi, line 45:
```suggestion
Both `env:URL` and `env:API_TOKEN` would be replaced by the variables defined in the chosen environment where the test will run. So, if the chosen environment was `dev.env`, its values would be replaced by `https://app-dev.com` and `dev-key`, respectively.
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
+# Transactions
+
+Most of End-to-End tests are not simple to run. They require some setup before the actual test is run. Maybe you have to create a new user, remove all items from a cart, etc. So it's important for you that you can execute multiple steps as part of your test suite. Tracetest introduces the concept of Transactions to achieve this goal.
```
jfermi, line 3:
```suggestion
Most End-to-End tests are not simple to run. They require some setup before the actual test is run. Actions like creating a new user, removing all items from a cart, etc. So, it's important that you can execute multiple steps as part of your test suite. Tracetest introduces the concept of **Transactions** to achieve this goal.
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+## What is a transaction
```
jfermi, line 5:
```suggestion
## What is a transaction?
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+A transaction is defined as a group of steps that are executed in the defined order and can access information exported by previous step executions. Each step is a test.
+
+## Chaining tests
```
jfermi, line 8:
```suggestion
## Chaining Tests
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+The main benefit of using transactions is to be able to chaining tests together and use values obtained from a test in a subsequent test.
```
jfermi, line 9:
```suggestion
The main benefit of using transactions is to be able to chain tests together and use values obtained from a test in a subsequent test.
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+### How values are shared by tests
```
jfermi, line 11:
```suggestion
### How Values are Shared by Tests
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+When a transaction is run, a context object is created with several informations about that specific run. One of those informations is an `environment variables` object which is empty by default. In case you run the transaction when referencing an [environment](environments.md), all values from the selected environments will be copied to the `environment variables` object.
```
jfermi, line 12:
```suggestion
When a transaction is run, a context object is created with information about that specific run. One of those pieces of information is an `environment variables` object, which is empty by default. If the transaction is run when referencing an [environment](environments.md), all values from the selected environments will be copied to the `environment variables` object.
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+When a test is executed within a transaction, if it generates any outputs, its outputs will be injected into the transaction context environment variables object. After the outputs are injected, all subsequent tests to be run within the transaction will be able to reference those values.
+
+> :information_source: Outputs generated by steps don't modify the selected [environment](environments.md). It only modifies the transaction run context object.
+
+Consider you have 3 tests within a transaction: A, B, and C. Both A and B generate outputs called A_OUTPUT and B_OUTPUT respectively. And when running the transaction, we provide an environment which contain a `HOST` environment variable. The execution of test A would only be able to reference `env:HOST`. B would be able to reference `env:HOST`, and `env:A_OUTPUT`. While C would be able to reference all three environment variables: `env:HOST`, `env:A_OUTPUT`, `env:B_OUTPUT`.
```
jfermi, line 18:
```suggestion
Consider you have 3 tests within a transaction: A, B, and C. Transactions A and B generate outputs called A_OUTPUT and B_OUTPUT, respectively. When running the transaction, we provide an environment which contains a `HOST` environment variable. The execution of test A would only be able to reference `env:HOST`. B would be able to reference `env:HOST`, and `env:A_OUTPUT`. While C would be able to reference all three environment variables: `env:HOST`, `env:A_OUTPUT`, `env:B_OUTPUT`.
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+> :information_source: A single test can contain as many outputs as you like.
+
+### Exposing values from a test to other tests
```
jfermi, line 22:
```suggestion
### Exposing Values from a Test to Other Tests
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+Since version v0.8, Tracetest allows tests to declare `outputs`. An output is a value that is extracted from a trace by providing it a [selector](selectors.md) to choose which spans to use to get the information from and an [expression](../expressions.md) to get the value from the selected spans. For example, consider that you want to expose the time a specific job was taken from a queue and began executing. An output would look something like the following:
```
jfermi, line 23:
```suggestion
Since version v0.8, Tracetest allows tests to declare `outputs`. An output is a value that is extracted from a trace by providing a [selector](selectors.md) to choose which spans to use to get the information from, and an [expression](../expressions.md) to get the value from the selected spans. For example, consider that you want to expose the time a specific job was taken from a queue and began executing. An output would look something like the following:
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+```yaml
+outputs:
+    - name: TIME_CANCEL_SUBSCRIPTION_MESSAGE_OBTAINED
+      selector: span[name = "Process request from cancel subscription queue"]
+      expression: attr:tracetest.time.start
+```
+
+This would create an output called `TIME_CANCEL_SUBSCRIPTION_MESSAGE_OBTAINED` that is obtained by reading the attribute `tracetest.time.start` from the span with `name` equals to `Process request from cancel subscription queue`. This value would would then be injected into the environment variables of that transaction to be accessed by other tests within the same transaction run.
```
jfermi, line 32:
```suggestion
This would create an output called `TIME_CANCEL_SUBSCRIPTION_MESSAGE_OBTAINED` that is obtained by reading the attribute `tracetest.time.start` from the span with `name` equal to `Process request from cancel subscription queue`. This value would would then be injected into the environment variables of that transaction to be accessed by other tests within the same transaction run.
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Transactions execution flow
```
jfermi, line 34:
```suggestion
### Transactions Execution Flow
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+Transaction steps are executed sequentially. A next step is only executed after the previous step finished executing and it is successful. A successful step is considered a step which managed to trigger an operation and got a trace back from the data store. Failing assertions do not make a transaction stop executing the next steps.
```
jfermi, line 36:
```suggestion
Transaction steps are executed sequentially. A next step is only executed after the previous step finishes executing successfully. A successful step is one which managed to trigger an operation and received a trace back from the data store. Failing assertions do not make a transaction stop executing the next steps.
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Transaction which one step didn't get executed
+
+* Step 1 (Finished)
+* Step 2 (Failed to fetch trace)
+* Step 3 (not executed)
+
+Result: **FAILED**
+
+### Transaction which all steps were executed, but they assertions failed
```
jfermi, line 48:
```suggestion
### Transaction where all steps were executed, but the assertions failed:
```

docs/concepts/transactions.md
```diff
@@ -0,0 +1,60 @@
...
+* Step 1 (Finished)
+* Step 2 (Finished)
+* Step 3 (Failed assertions)
+
+Result: **FAILED**
+
+### Transaction which all steps succeded
```
jfermi, line 55:
```suggestion
### Transaction where all steps succeeded:
```

jfermi (commented):
Some editing suggestions.

adnanrahic (approved):

PR #1562, opened 2022-11-22
feat: add startTime and endTime span attributes

mathnogueira:
This PR adds `tracetest.span.startTime` and `tracetest.span.endTime` to the span attributes. This allows users to check if an operation happened within X seconds since another operation by using test outputs.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/traces/traces.go
+++ b/server/traces/traces.go
@@ -54,6 +54,8 @@ func augmentSpanData(span *Span) {
 	span.Attributes["name"] = span.Name
 	span.Attributes["tracetest.span.type"] = spanType(span.Attributes)
 	span.Attributes["tracetest.span.duration"] = spanDuration(*span)
+	span.Attributes["tracetest.span.startTime"] = fmt.Sprintf("%d", span.StartTime.UnixNano())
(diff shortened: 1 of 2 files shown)


xoscar (approved):

xoscar:
did something happen to the service name? I think I haven't seen it coming up as part of the attributes anymore

schoren (approved):
Is nanoseconds too much resolution? I would use ms, but not sure If its nano can be more useful

jorgeepc (approved):

mathnogueira:
> did something happen to the service name? I think I haven't seen it coming up as part of the attributes anymore

I agree `ms` would be enough. I used `ns` because we allow operations on that `tracetest.span.startTime + 15us`. I decided to add more precision available so we don't end up creating some rounding problems when users try to use that feature.

mathnogueira:
> Is nanoseconds too much resolution? I would use ms, but not sure If its nano can be more useful

I agree `ms` would be enough for most cases. I used `ns` because we allow operations on that `tracetest.span.startTime + 15us`. I decided to add more precision available so we don't end up creating some rounding problems when users try to use that feature and require more precision.

PR #1575, opened 2022-11-23
fix(frontend): fixing assertion value autocomplete overlap

xoscar:
This PR fixes some UI issues when adding assertions to test specs and updates the expression query language to better support json_path.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


jorgeepc (approved):
Looks good!

PR #1577, opened 2022-11-24
feat: normalize transaction steps info

schoren:
This PR changes the persistence and API related to transaction runs. It now reuses the openapi entities, and uses normalized relational tables instead of json-encoded fields to store test runs.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/transaction_runner.go
```diff
@@ -121,93 +117,81 @@ func (r persistentTransactionRunner) runTransaction(ctx context.Context, run mod
...
+			if testRun.State.IsFinal() {
+				done <- true
+			}
 
-			return r.updateTransactionRun(ctx, stepIndex, &transactionRun, runResult)
+			return nil
 		}),
 	)
+	// TODO: this will block indefinitely. we need to set a timeout or something
+	<-done
```
mathnogueira, line 158: Without the timeout, this can potentially break tracetest again. You probably can use a simple select here. Probably something like:

```suggestion
   select {
     case <-done:
       return tr, err
     case <- time.After(timeout):
       return tr, fmt.Errorf("transaction step execution timed out")
   }
```

server/executor/transaction_runner.go
```diff
@@ -121,93 +117,81 @@ func (r persistentTransactionRunner) runTransaction(ctx context.Context, run mod
...
-		return err
+		return model.TransactionRun{}, fmt.Errorf("could not update transaction run: %w", err)
 	}
 
-	stepRun := createStepRun(testRun)
+	return tr, nil
+}
 
-	transactionRun.StepRuns[stepIndex] = stepRun
+// func (run TransactionRun) InjectOutputsIntoEnvironment(env Environment) Environment {
```
mathnogueira, line 177: Where are you injecting the outputs into the environment?

mathnogueira (changes requested):
It seems you are not injecting the outputs of the step into the following step environment. It would be nice to have the timeout in place to prevent Tracetest from freezing.

server/executor/transaction_runner.go
schoren: I thought of this, but how much time is a timeout? It will depend on the time it takes for the test to finish. Tests don't have a timeout other than the http and trace polling, so any timeout we put here will be the same as waiting for the test to finish.

This comment is not meant for now, but for later, when we implement a more advanced logic for timing tests etc

server/executor/transaction_runner.go
schoren: uppps, fixed!

mathnogueira (approved):

PR #1587, opened 2022-11-24
ci: add adnam to team and remove PR notification

mathnogueira:
This PR adds @adnanrahic to the team and remove the PR notification pipeline

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


.github/workflows/community-pr-notification.yml
```diff
@@ -1,35 +0,0 @@
-# Send us messages on slack when users from the community either open new issues or PRs
```
jorgeepc, line 1: Why are we deleting this one?

jorgeepc (approved):

.github/workflows/community-pr-notification.yml
mathnogueira: Because it doesn't work.

It requires the secret for the slack channel, however, GitHub doesn't allow secrets to be referenced in pipelines that are triggered by people who are not contributors to the project. Otherwise, anyone could open a PR to the project printing all secrets and stealing our credentials.

So, I don't think there's an easy way of notifying us about this using GitHub Actions at the moment.

Issue #1598, opened 2022-11-28
Last run does not appear to be working


kdhamric:
I ran a transaction, then went to the main screen and sorted by 'last run'. The transaction did not show up, even thought tests that were run earlier than it are in the list. Appears that last run never works (possibly), and that it always shows based on created time.

https://www.loom.com/share/1bc63b8fc42d4baaa7f1d09ed45fa59d

jorgeepc:
@kdhamric  It's an API issue. I'm seeing this error in the API response `"pq: column \"last_test_run_time\" does not exist"` cc: @schoren @mathnogueira

schoren:
Just spotted the issue, I'll create a PR for this

kdhamric:
Last run does not seem to show the newly created tests as being 'run last'. For example, it shows the last run test was 2 hours ago.
![image](https://user-images.githubusercontent.com/1385894/204395677-b725b6fb-bc2c-44f9-831f-a1fab2f9b7f7.png)

When you look at recently created, it shows test created in the last 50 minutes (vs 2 hours):
![image](https://user-images.githubusercontent.com/1385894/204395828-d522f384-cc9b-42ff-8f38-b9316fd9c6e4.png)

Expected behaviour: 'Last run' would include tests created recently (ie creation causes a 'run')

PR #1600, opened 2022-11-28
fix(server): map transactions createdAt sort field to correct db field

schoren:
This PR fixes the mapping used to sort transactions by `createdAt` field.

## Fixes

- #1598 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/testdb/environments.go
+++ b/server/testdb/environments.go
@@ -54,11 +54,6 @@ INSERT INTO environments (
 	`
 )
 
-var environmentSortingFields = map[string]string{
-	"created": "e.created_at",
-	"name":    "e.name",
-}
-
 func (td *postgresDB) CreateEnvironment(ctx context.Context, environment model.Environment) (model.Environment, error) {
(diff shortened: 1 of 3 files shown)


server/testdb/environments.go
```diff
@@ -113,7 +108,12 @@ func (td *postgresDB) GetEnvironments(ctx context.Context, take, skip int32, que
 		sql += condition
 	}
 
-	sql = sortQuery(sql, sortBy, sortDirection, environmentSortingFields)
+	sortFields := map[string]string{
```
jorgeepc, line 111: just for consistency, should we name this `sortingFields` instead of `sortFields`?

jorgeepc (approved):

server/testdb/environments.go
schoren: given it's just an internal, temporary variable, it's not very relevant, but I'll fix it

PR #1611, opened 2022-11-30
PoC: allow running tracetest locally without dependencies

schoren:
This PR allows the server to run without dependencies by directly running the `tracetest-server` binary locally.

It embed the web build in the server binary, and uses a (unfinished) filesystem based testDB.

To run this version, you need to checkout this branch and build (you need [goreleaser](https://goreleaser.com/install/) installed):
```
# on project root 
goreleaser release --skip-announce  --skip-publish --snapshot --rm-dist
```
This will create a `dist/[architecture]/tracetest-server` file. You can run that file to start the server

---

Now you need to run the server, passing a root "database" dir. This will hold your "ignorable" db (like runs and config) and your "persistable" db (tests, envs, transactions).
You can use the `tracetesting` dir from the project:

```
dist/[architecture]/tracetest-server ./tracetesting/transactions
```

---

Finally, go to `http://localhost:11633` 

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/main.go
```diff
@@ -26,14 +28,16 @@ var cfg = flag.String("config", "config.yaml", "path to the config file")
 func main() {
 
 	flag.Parse()
+	root := flag.Arg(0)
```
mathnogueira, line 31: I think we should have a default here. Probably some directory in the working directory, such as: `./.tracetest` or so.

server/testdb/fs/fsdb.go
```diff
@@ -0,0 +1,250 @@
...
+		})
+	if err != nil {
+		return nil, err
+	}
+
+	return files, nil
+}
+
+func (td *fsDB) Drop() error {
+	panic("not implemented")
```
mathnogueira, line 243: I think it's worth implementing this because it's very simple:

```suggestion
	return os.RemoveAll(td.root)
```

mathnogueira (commented):
Looks good! Just a few minor comments

PR #1613, opened 2022-11-30
feat: adding definition files for transactions and environments

xoscar:
This PR adds both the BE and FE features to support definition files for transactions and environments.

## Changes

- New API services for transaction and environment definition files
- New UI to support the new BE services
- Updates FE endpoint services to use a single entry point for all resources definitions.

## Fixes

- [[Transactions] Display definition file for transactions#1526](https://github.com/kubeshop/tracetest/issues/1526)

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/http/custom_routes.go
```diff
@@ -124,6 +127,42 @@ func (c *customController) GetTestVersionDefinitionFile(w http.ResponseWriter, r
 	w.Write(result.Body.([]byte))
 }
 
+// GetTransactionVersionDefinitionFile - Get the test definition as an YAML file
+func (c *customController) GetTransactionVersionDefinitionFile(w http.ResponseWriter, r *http.Request) {
```
xoscar, line 131: At some point I think it would be nice to have a better way to define custom routes without having to duplicate the most of the code everytime

server/http/controller.go
```diff
@@ -710,6 +710,20 @@ func (c *controller) GetEnvironments(ctx context.Context, take, skip int32, quer
 	}), nil
 }
 
+func (c *controller) GetEnvironmentDefinitionFile(ctx context.Context, environmentId string) (openapi.ImplResponse, error) {
```
xoscar, line 713: The controller file is getting quite large, let's consider a refactor to split it up. I like DDD for API backend development

web/src/components/FileViewerModal/FileViewerModal.provider.tsx
```diff
@@ -41,28 +42,40 @@ const FileViewerModalProvider = ({children}: IProps) => {
...
       setFileViewerData({data, type: 'junit'});
+      setProps(propsMap.junit);
     },
     [getJUnit]
   );
 
-  const loadTestDefinitionYaml = useCallback(
-    async (testId: string, version: number) => {
-      const data = await getTestDefinitionYaml({testId, version}).unwrap();
+  const loadDefinition = useCallback(
```
xoscar, line 63: This provider now handles the definition fetch and display for tests, transactions and environments

server/http/controller.go
schoren: +1 to both comments. I think we need to move on from the openapi generator for anything other than entities

cli/openapi/api_api.go
```diff
@@ -47,8 +47,8 @@ CreateEnvironment Create new environment
 
 Create new environment action
 
- @param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
- @return ApiCreateEnvironmentRequest
+	@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
```
schoren, line 50: @xoscar how are you generating this files? we shouldn't have spacing diffs here

server/model/yaml/yamlconvert/environments.go
```diff
@@ -0,0 +1,28 @@
...
+	valueList := make([]yaml.EnvironmentValue, len(in.Values))
+
+	for index, value := range in.Values {
+		valueList[index] = yaml.EnvironmentValue{
+			Key:   value.Key,
+			Value: value.Value,
+		}
+	}
+
+	out.Values = valueList
```
schoren, line 22: I think you can just `dc.DeepCopy(in.Values, &out.Values)`

server/model/yaml/yamlconvert/transactions.go
```diff
@@ -0,0 +1,25 @@
...
+	out := yaml.Transaction{}
+	dc.DeepCopy(in, &out)
+
+	stepList := make([]string, len(in.Steps))
+
+	for index, step := range in.Steps {
+		stepList[index] = string(step.ID)
+	}
+
+	out.Steps = stepList
```
schoren, line 19:
```suggestion
	out.Steps = make([]string, 0, len(in.Steps))

	for _, step := range in.Steps {
		out.Steps = append(out.Steps, step.ID.String()
	}
```

schoren (changes requested):
backend looks great, except for a few details that would make the code easier to read and more consistent. great work!

mathnogueira:
Oscar is at the airport. I rebased and fixed the things you suggested, @schoren

schoren (approved):

cli/openapi/api_api.go
mathnogueira: That was strange. At first, I thought it was a versioning issue, but we are pinning the generator version in the makefile.

PR #1620, opened 2022-12-01
fix: remove partial test update endpoints

### LINKED ISSUE #1553: Duplicate variable entries get added, and deletion fails
Added a variable based on an attribute.
Then edited the variable.
Ended up with 2 duplicate variables
In trying to delete the two variables, I could not successfully delete them both
Seems reproducible - see video

Not sure if this is a backend or frontend issue.

https://www.loom.com/share/464b4dc79e78432e941f4e14acac1938

### LINKED ISSUE #1605: All outputs can not be removed from test
**Describe the bug**
With the changes introduced here https://github.com/kubeshop/tracetest/pull/1547 it's not possible to remove all the outputs from a test.

**To Reproduce**
1. Go to a test run page, select the Trigger tab
2. Click on `Outputs`, add one output and click on `Save & Run`
3. Delete the output and click on `Save & Run`.
4. See that the output is not deleted.

**Expected behavior**
All the outputs should be deleted.

**Screenshots**

**Additional context**
Code preventing the deletion
https://github.com/kubeshop/tracetest/blob/27f5d6ae6396636a2f34e36dbbc3ebb9825e1b24/server/testdb/tests.go#L107-L110

### LINKED ISSUE #1608: Remove partial test update endpoints
**Context:** currently we are using two partial test update endpoints to interact with `specs` and `outputs`. We want to remove those endpoints and use the main `test update` service.

**AC1:**
As a user interacting with test `specs`,
I should be able to create, update and delete specs by using the main `test update` service.

**AC2:**
As a user interacting with test `outputs`,
I should be able to create, update and delete outputs by using the main `test update` service.

jorgeepc:
Currently we are using two partial test update endpoints to interact with `specs` and `outputs`. This PR removes those endpoints and use the main test update service.

## Changes

- Remove BE specs/outputs endpoints
- Change FE code to use the main test update service

## Fixes

- fixes #1608 
- fixes #1605  
- fixes #1553 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
love this "remove code" PRs ❤️

web/src/providers/Test/Test.provider.tsx
```diff
@@ -30,13 +31,15 @@ export const Context = createContext<IContext>({
...
-  version?: number;
   children: React.ReactNode;
 }
 
 export const useTest = () => useContext(Context);
 
-const TestProvider = ({children, testId, version = 0}: IProps) => {
+const TestProvider = ({children, testId}: IProps) => {
+  const {run} = useTestRun();
+  const version = run.testVersion || 0;
```
xoscar, line 41: What's the advantage of this change? I've seen that we use the TestProvider outside of the test run context, I wouldn't like to have this coupled

web/src/providers/TestOutput/TestOutput.provider.tsx
```diff
@@ -101,12 +105,16 @@ const TestOutputProvider = ({children, testId, runId}: IProps) => {
 
   const onSave = useCallback(
     async (outputs: TTestOutput[]) => {
-      await setTestOutputs({testId, testOutputs: toRawTestOutputs(outputs)}).unwrap();
-      const run = await reRunTest({testId, runId}).unwrap();
+      const plugin = TriggerTypeToPlugin[test?.trigger?.type || TriggerTypes.http];
```
xoscar, line 108: Should we encapsulate this into a single layer so we don't have to keep adding it for the different sections?

xoscar (approved):
Hey @jorgeepc gj man, I left some small comments, let me know what you think

Issue #1498, opened 2022-11-15
Clicking on the test spec indication next to an attribute is not transitioning to test spec details


kdhamric:
When you see the indication that a span or attribute is 'under test' in the trace view or when looking at a list of attributes, you should be able to click on it and view the details of the test spec. Looks like this is not working.

Recorded a video: https://www.loom.com/share/0481ff294cd6439b99742095f7c8ed95

PR #1636, opened 2022-12-02
fix(frontend): fixes summary checks not opening the proper test spec

xoscar:
This PR fixes an issue caused by the expression changes introduced with Transactions. Before showing the results checks for each attribute span we would match the attribute string, as that changed it broke the logic.

## Changes

- Now we are looking at the assertion string to see if the attribute exists as part of the operation.

## Fixes

- https://github.com/kubeshop/tracetest/issues/1498

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/components/AssertionResultChecks/AssertionResultChecks.styled.ts
+++ b/web/src/components/AssertionResultChecks/AssertionResultChecks.styled.ts
@@ -1,0 +1,57 @@
+import {Badge} from 'antd';
+import styled, {css} from 'styled-components';
+
+export const Container = styled.div`
+  display: flex;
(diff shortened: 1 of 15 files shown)


jorgeepc (approved):
Hey @xoscar great job. Looks good, just a comment, I noticed that now we have a separate click in the check here, and this could open the dropdown when you have multiple checks. In this case we are getting duplicated values because 3 of the checks corresponds to the same selector. 
Just want to confirm that this was expected.

Screenshot 2022-12-05 at 10 00 39

PR #1641, opened 2022-12-05
feature(backend): config services open api definition

xoscar:
This PR adds the initial openAPI definition for the config services

## Changes

- Includes the open API changes
- Adds the definitions for data stores and exporters

### DIFF
--- a/api/config.yaml
+++ b/api/config.yaml
@@ -52,7 +52,7 @@ components:
         openSearch:
           $ref: "#/components/schemas/OpenSearch"
         signalFx:
-          $ref: "#/components/schemas/SignalFx"
+          $ref: "#/components/schemas/SignalFX"
     OpenSearch:
       type: object
       properties:
@@ -146,7 +146,7 @@ components:
             openSearch:
               $ref: "#/components/schemas/OpenSearch"
             signalFx:
-              $ref: "#/components/schemas/SignalFx"
+              $ref: "#/components/schemas/SignalFX"
             collector:
               $ref: "#/components/schemas/CollectorConfig"
     TestConnectionResponse:
--- a/api/config.yaml
+++ b/api/config.yaml
@@ -1,0 +1,170 @@
+openapi: 3.0.0
+components:
+  schemas:
+    Config:
+      type: object
+      properties:
+        telemetry:
+          $ref: "./components/schemas/TelemetryConfig"
+        server:
+          $ref: "./components/schemas/Server"
+    TelemetryConfig:
+      type: object
+      properties:
+        dataStores:
+          type: array
+          items:
+            $ref: DataStore
+        exporters:
+          type: array
+          items:
+            $ref: "./components/schemas/ExporterOption"
+    ExporterOption:
+      type: object
+      properties:
+        serviceName:
+          type: string
+        sampling:
+          type: number
+        exporter:
+          $ref: "./components/schemas/Exporter"
+    Exporter:
+      type: object
+      properties:
+        type:
+          type: string
+        collector:
+          $ref: "./components/schemas/CollectorConfig"
+    CollectorConfig:
+      type: object
+      properties:
+        endpoint:
+          type: string
+    DataStore:
+      type: object
+      properties:
+        type:
+          type: string
+        jaeger:
+          $ref: "./components/schemas/GRPCClientSettings"
+        tempo:
+          $ref: "./components/schemas/GRPCClientSettings"
+        openSearch:
+          $ref: "./components/schemas/OpenSearch"
+        signalFx:
+          $ref: "./components/schemas/SignalFx"
+    OpenSearch:
+      type: object
+      properties:
+        addresses:
+          type: array
+          items:
+            type: string
+        username:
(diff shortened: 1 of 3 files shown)


api/config.yaml
```diff
@@ -0,0 +1,170 @@
+openapi: 3.0.0
+components:
+  schemas:
+    Config:
+      type: object
+      properties:
+        telemetry:
+          $ref: "./components/schemas/TelemetryConfig"
+        server:
+          $ref: "./components/schemas/Server"
```
mathnogueira, line 10: This PR is only for data stores, right? This config won't allow users to set polling timeouts and retry delays.

xoscar: Exactly, not yet, I just wanted to focus on data stores & exporters

schoren:
This PR itself looks fine. My only concern is the amount of duplication this introduces. Can't we figure out a way to send config updates without the need to rewrite the entire config in openapi?

In this case, I think it's particularly impractical because some of our definitions are "external", like the tempo/jaeger gRPC endpoints.

What if we have a single endpoint, like  `PUT /api/server/config` that accepts an entire yaml config file? Then the UI can implement the parts that are neccesary and the backend can "merge" partial updated with existing config. 
This approach would leave to a much more flexible endpoint and config format, and decouples the UI from the actual server implementation.

Thoughts?

mathnogueira:
> What if we have a single endpoint, like `PUT /api/server/config` that accepts an entire yaml config file? Then the UI can implement the parts that are neccesary and the backend can "merge" partial updated with existing config. This approach would leave to a much more flexible endpoint and config format, and decouples the UI from the actual server implementation.

If we provide a way of validating the config, I'm ok with that.

xoscar:
> This PR itself looks fine. My only concern is the amount of duplication this introduces. Can't we figure out a way to send config updates without the need to rewrite the entire config in openapi?
> 
> In this case, I think it's particularly impractical because some of our definitions are "external", like the tempo/jaeger gRPC endpoints.
> 
> What if we have a single endpoint, like `PUT /api/server/config` that accepts an entire yaml config file? Then the UI can implement the parts that are neccesary and the backend can "merge" partial updated with existing config. This approach would leave to a much more flexible endpoint and config format, and decouples the UI from the actual server implementation.
> 
> Thoughts?

My concern is that we are killing the usage of openAPI, if we cannot generate the FE types based on it, and we get it as a raw object we'll have to parse it to YAML in the FE

Also, we still need a GET endpoint right? because the FE needs to know what config is already there

schoren:
Those are valid points in favor of openapi mappings. But in the same line, there's a lot of manual, error prone mapping needed in the backend to convert from an openapi entity to an internal config type, so it's a matter of at what stage we want to loose the "type safety" of openapi.

xoscar:
> Those are valid points in favor of openapi mappings. But in the same line, there's a lot of manual, error prone mapping needed in the backend to convert from an openapi entity to an internal config type, so it's a matter of at what stage we want to loose the "type safety" of openapi.

@schoren do you mind explaining what would be the steps a client app (FE, CLI) would have to follow to fetch, manipulate and send the update request with that approach?

Like in what file format do we get it, do we have to parse it to YAML in our end? etc etc.

xoscar:
> @xoscar sure. Something like this:
> 
> 1. `GET /server/config` returns the entire yaml config file. The client is responsible for parsing it and getting the parts it needs
> 2. `PUT /server/config`, with the body containing the entire file, or a part of it. Partial updates would probably require to keep the "tree", like if you want to update the `jaeger` datastore, you would need to send something like this:
> 
> ```
> telemetry:
>   dataStores:
>     jaeger:
>       jaeger:
>         endpoint: updated endpont # this value is the only one being modifed
> ```
> 
> That's it. Each client needs to understand the relevant parts of the config file, but only those parts. This provides the flexibility to have a CLI that doesn't need to understand anything, just send an entire (or partial, for that matter) config file, and the only thing it needs to understand is the server response code with validation details.
> 
> The UI will need to know a lot more about how to config each data store, but IMO that's the FE responsibility, in the sense that you cannot escape that part. The FE will need to know how to configure a Tempo DS, even if it's just to create the UI.
> 
> The server will always validate the config, even if it's already validated by the UI. So even adding a lot of validation on the FE would be redundant in this case.
> 
> My point comes around my understanding that the FE will be "hardcoded" one way or the other. If this is not the case, my thought would be very different. But if there will be hand written code that replicates the structure of the config file, then there's no point of duplicating it in the FE, the BE, and the openAPI specs, when it can be only written on the FE, and leave the BE to handle configs in a more flexible way.
> 
> The only way I imagine this extra step of redefining everything in openapi is if you are planning to have a dynamic UI, that can automatically render forms based on the generated types from openAPI. In this case, the extra effort would be paid by the fact that future updates to the config, like adding new backends, or supporting new options in existing ones would come "for free".

@schoren makes sense, let's go with that approach, I'll update this PR 😄

xoscar:
@schoren @mathnogueira what do you think about the test connection services? anything that you'll want to see differently?

Issue #1622, opened 2022-12-01
[In-App Config] Setup Wizard


kdhamric:
The setup wizard is the main in-app config feature that allows users to configure tracetest to start supporting traces. 

Acceptance Criteria:
AC1:
As a user looking at the setup wizard
I should be able to see a brief explanation about why a data store is needed
And a way to select the data store of my preference

AC2 (Special Case for Otel Collector):
As a user looking at the UI setup wizard
And I select the Otel Collector as my data store
I should see a small tutorial with the steps I need to take to configure it

AC3 (Data Store Step):
https://github.com/kubeshop/tracetest/issues/1627

AC7:
As a user looking at the UI setup wizard
After filling in the details to configure the data store
I should be able to save the changes
Without testing the connection

AC8:
As a user looking at the UI setup wizard
And I have clicked save changes
I should see a prompt message letting me know that Tracetest will be restarted and there will be some downtime after confirmation.

AC9:
As a user that just went through the setup process
I should see a change in the UI that will tell me that Tracetest is ready to support traces

kdhamric:
@mathnogueira @schoren Can one of you fill in the details of what needs to be gathered to connect to each of the natively supported data stores?

PR #1652, opened 2022-12-07
feature(frontend): adding base logic for the data store config form

xoscar:
This PR adds the base logic for the data store config form.

## Changes

1. Plugin logic for the different data store types
2. Config draft object and reducer
3. Adding parsing services for the draft and initial values

## Fixes

- [[In-App Config] Setup Wizard#1622](https://github.com/kubeshop/tracetest/issues/1622)

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


web/src/components/DataStoreForm/DataStoreForm.tsx
```diff
@@ -0,0 +1,38 @@
+import {Form} from 'antd';
+import {useMemo} from 'react';
+import SetupConfigService from 'services/SetupConfig.service';
+import {TConfig, TDraftConfig, TDraftConfigForm} from 'types/Config.types';
+import DataStoreComponentFactory from '../DataStorePlugin/DataStoreComponentFactory';
+import * as S from './DataStoreForm.styled';
+
+export const FORM_ID = 'edit-test';
```
jorgeepc, line 8: let's change this `edit-test`

web/src/providers/SetupConfig/SetupConfig.provider.tsx
```diff
@@ -0,0 +1,84 @@
...
+  onReset: noop,
+});
+
+interface IProps {
+  children: React.ReactNode;
+}
+
+export const useSetupConfig = () => useContext(Context);
+
+const SetupConfigProvider = ({children}: IProps) => {
```
jorgeepc, line 33: What is the reason for having a new Provider? Could this be part of the `Config.provider`?

jorgeepc (approved):
Great job @xoscar. Added a couple of questions. Also, what do you think about moving all the config related components to `web/src/components/Settings/`? Not a big deal but we can analyze a new organization for the new features.

web/src/providers/SetupConfig/SetupConfig.provider.tsx
xoscar: I was thinking that the ConfigProvider would be for read-only, while the setup config one is the one we'll be using to manipulate the config draft for each of the tabs and sections of the settings page. Does it make sense?

Issue #1656, opened 2022-12-08
[In-App Config] Data Store Config endpoint


xoscar:
As discussed, we are moving away from the idea of having a single endpoint for the entire tracetest configuration object, instead, we are going to trigger updates based on sections.

For the initial in-app config options we need to have a way to only update the data store configuration.

Acceptance criteria:

AC1:
As a user looking at the data store config form
And I update the configuration
At the same time, one other user changes another section of the config
Any of the two should be overwritten but rather both should be saved to the app configuration

AC2:
As a user looking at configuring tracetest using the config file
And I want to add the data store section
I should be able to understand the name/type of each supported data store
And it should match whatever I see in the frontend

AC3:
As a user looking at configuring tracetest using the config file
And I want to add the data store configuration
I should always add the default data store in the server section
So the frontend UI can display the proper datastore config

xoscar:
@mathnogueira I'm going to start working on the FE changes required for this. Let me know if you want to talk about the open API specs and such!

xoscar:
FE And openAPI changes https://github.com/kubeshop/tracetest/pull/1659

PR #1659, opened 2022-12-08
feature(frontend): prepare FE for granular data store service update

xoscar:
This PR updates the openAPI specs to now use a custom config endpoint just for data stores.
It also removes a lot of the FE logic to handle the full configuration object for the settings page.

## Changes

- Updates openAPI
- Removes unnecessary reducer and code

## Fixes

- https://github.com/kubeshop/tracetest/issues/1656

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


api/config.yaml
```diff
@@ -53,6 +53,15 @@ components:
...
           $ref: "#/components/schemas/SignalFX"
+    UpdateDataStoreConfigRequest:
+      type: object
+      properties:
+        dataStores:
+          type: array
+          items:
+            $ref: "#/components/schemas/DataStore"
+        defaultDataStore:
+          $ref: "#/components/schemas/SupportedDataStores"
```
mathnogueira, line 64: I think this should be a string instead. The user might have 2 different jaeger instances and want to switch between them. This should reference the data store id

api/openapi.yaml
```diff
@@ -1073,24 +1073,23 @@ paths:
             application/json:
               schema:
                 $ref: "./config.yaml#/components/schemas/Config"
+  /config/datastore:
```
mathnogueira, line 1076: Should the `get` method be included in this PR? It's ok if the answer is 'no'.

mathnogueira (changes requested):
Small thing about the openapi spec

api/config.yaml
xoscar: Great callout  @mathnogueira in that case, should we include a name entry or something like that?

mathnogueira (commented):
OpenAPI spec looks good

xoscar:
@mathnogueira @jorgeepc this is ready to be re-reviewed, if you think it looks good please approve so we can merge and move forward. Thanks guys!

xoscar:
@mathnogueira this PR is kind of a blocker for the FE, if you think everything looks fine, do you mind approving?

mathnogueira (approved):

PR #1680, opened 2022-12-13
fix(frontend): remove tabs in settings page

jorgeepc:
This PR removes the tabs in the settings page because right now we only support Data Store configuration.

## Changes

- Remove tabs in settings page

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Screenshots

Screenshot 2022-12-13 at 16 04 53

### DIFF
--- a/web/src/pages/Settings/Content.tsx
+++ b/web/src/pages/Settings/Content.tsx
@@ -1,11 +1,11 @@
-import {Tabs} from 'antd';
+// import {Tabs} from 'antd';
 import DataStore from 'components/Settings/DataStore';
 import {useDataStoreConfig} from 'providers/DataStoreConfig/DataStoreConfig.provider';
 import * as S from './Settings.styled';
 
-const TabsKeys = {
+/* const TabsKeys = {
   DataStore: 'dataStore',
-};
+}; */
 
 const Content = () => {
   const {dataStoreConfig} = useDataStoreConfig();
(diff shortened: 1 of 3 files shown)


web/src/pages/Settings/Content.tsx
```diff
@@ -1,28 +1,28 @@
...
     <S.Container>
       <S.Header>
-        <S.Title>Settings</S.Title>
+        <S.Title>Configure Data Store</S.Title>
       </S.Header>
 
-      <S.TabsContainer>
+      {/* <S.TabsContainer>
         <Tabs size="small">
-          <Tabs.TabPane key={TabsKeys.DataStore} tab="Configure Data Store">
```
xoscar, line 21: I guess you did this so we can go back to tabs more easily once we need it right?

xoscar (approved):
Hey @jorgeepc code changes look good! 

Tbh I don't see why having a single tab might be "bad" but I don't have a strong opinion either. 

I think is time for @olha23 and @kdhamric to decide haha

web/src/pages/Settings/Content.tsx
jorgeepc: That's right. I think we will add more options soon.

kdhamric: I personally would leave the tab... does not distract me. It is, however, @olha23 UX.

xoscar: @olha23 for final approval 😄

olha23: we will add tabs when there is more than one

xoscar: Sold! feel free to merge @jorgeepc

Issue #1679, opened 2022-12-13
[In-App Config] Test Trace Data Store Connection 


kdhamric:
We need to provide the user with information about the test(s) ran when pressing the 'Test Trace Data Store' button when configuring a trace data store.

Providing more details forAC6 from #1622, ie:

AC6: (from other issue, but all applies)
As a user looking at the UI setup wizard
After filling in the details to configure the exporter section
And I choose the test connection option
I should be able to see the result (error or success)
And details about the results

Specific to these requirements, these are expanded acceptance criteria:

AC1: 
As a user, 
I should be able to tell that the 'test connection' button was pressed
And that we are awaiting results

AC2: 
As a user, 
I should be able to tell the exact test(s) that ran as part of the connection testing (1-4 total tests)
And have details for each individual test
Including information about 1) what was tested with the each test, 2) the result / error message from the test and 3) the status of it (pass / failed)
As well as the overall test result (passed / failed)

olha23:
mockups here https://www.figma.com/file/LBN4SKVPq3ykegrPKbHT2Y/0.8-0.9-Release-Tracetest?node-id=1103%3A21127&t=OQa7f7SEOtqIGzz0-1

PR #1683, opened 2022-12-13
chore(backend): updating the open API contract to support datastore test connection changes

xoscar:
This PR adds the OpenAPI contract for the test connection service response

## Changes

- Adds the DataStoreTestConnectionChecks

## Fixes

- #1679 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/api/config.yaml
+++ b/api/config.yaml
@@ -105,14 +105,25 @@ components:
           type: string
         maxVersion:
           type: string
-    TestConnectionRequest:
+    DataStoreTestConnectionRequest:
       $ref: "#/components/schemas/DataStore"
-    TestConnectionResponse:
+    DataStoreTestConnectionResponse:
       type: object
       properties:
-        successful:
+        allPassed:
           type: boolean
-        errorMessage:
+        checks:
+          type: array
+          items:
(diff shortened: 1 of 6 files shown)


api/config.yaml
```diff
@@ -105,14 +105,25 @@ components:
...
       properties:
-        successful:
+        allPassed:
           type: boolean
-        errorMessage:
+        checks:
+          type: array
+          items:
+            $ref: "#/components/schemas/DataStoreTestConnectionCheck"
+    DataStoreTestConnectionCheck:
```
xoscar, line 119: Depending on the data store and strategy we want to use, the response could have n number of checks

api/openapi.yaml
```diff
@@ -1100,11 +1100,11 @@ paths:
         content:
           text/json:
             schema:
-              $ref: "./config.yaml#/components/schemas/TestConnectionRequest"
+              $ref: "./config.yaml#/components/schemas/DataStoreTestConnectionRequest"
```
jorgeepc, line 1103: Should we change the path `/config/connection` to something more data store specific? And should we document the `4xx` status code when the test connection fails?

PR #1690, opened 2022-12-14
feat(server): add data stores CRUD API

### LINKED ISSUE #1688: [In-App Config] Create database design for Trace Data Store Configuration
We need to have a database design for storing the trace data store. Details to be determined by the technical team. Would be good, but not required, if the design takes into account that we will likely have multiple data stores (ie rows) in the next 6 months.

jorgeepc:
This PR adds the initial data stores db schema and the basic CRUD API.

## Changes

- New data stores CRUD API

## Fixes

- fixes #1688 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/migrations/16_data_stores.up.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+CREATE TABLE "data_stores" (
+  "id" varchar not null primary key,
+  "name" varchar,
+  "type" varchar not null,
+  "is_default" boolean not null,
```
mathnogueira, line 7: We just have to ensure there are not two default data stores at the same time.

mathnogueira (approved):

mathnogueira (changes requested):
I think it's worth adding some component tests to the db queries. Those are terrible/maintain so we need a good way of ensuring it works fine.

mathnogueira (approved):

Issue #1628, opened 2022-12-02
[In-App Config] No-Tracing Mode


kdhamric:
When the app is not ready to support data stores, traces, spans, etc. It should be in a specific mode called no-tracing, where it should function just with the immediate request/response and tracetest triggering spans.

Acceptance Criteria:
AC1:
As a user that just installed Tracetest
And I have decided to not provide the data store info at the beginning
I should be able to check if the app is in no-tracing mode

AC2:
As a user that just installed Tracetest
And I have decided to not provide the data store info at the beginning
I should be able to create tests

AC3:
As a user that just created a test in no-tracking mode
I should be redirected to the run page
And see the immediate result of the service

AC4:
As a user looking at a no-tracking mode test
I should be able to move to the trace and test modes
And the trace should display a single trace trigger span

AC5:
As a user looking at a no-tracing mode test
I should be able to create test specs on the trace trigger span

xoscar:
@olha23 I think we need to add some designs on a global way to display the no tracking mode for users, it could be a warning badge in the nav menu that opens something or anything like that.

For the backend, the tasks here are:
1. Create the fake trace and trigger span
2. Do not wait for the full trace

CC: @mathnogueira @kdhamric

xoscar:
bump @mathnogueira

PR #1695, opened 2022-12-15
feature(backend): adding no-tracing mode support for the BE

xoscar:
This PR enables the BE to support the no-tracing mode by adding a synchronous job processing by creating the root span directly from Tracetest.

## Changes

- For no tracing mode, the poller executor now adds the root span, finishes the trace fetch process, and runs assertions
- For tracing mode, we are now injecting the root span directly into the trace

## Fixes

- #1628 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/f97f9ce5aedd46d0b0c03e58acbea567


server/config/config.go
```diff
@@ -118,14 +118,26 @@ func (c Config) MaxWaitTimeForTraceDuration() time.Duration {
...
+	}
+
 	return &dataStoreConfig, nil
 }
 
+func (c Config) IsDataStoreConfigured() bool {
+	selectedStore := c.Server.Telemetry.DataStore
+	_, found := c.Telemetry.DataStores[selectedStore]
+
+	return found
```
schoren, line 138: We could avoid logic duplication by reusing the `DataStore` method

```suggestion
	ds, err := c.DataStore()
	return err == nil && ds != nil
```

server/executor/trace_poller.go
```diff
@@ -108,11 +111,38 @@ func (tp tracePoller) Stop() {
...
+
+	if tp.isDataStoreConfigured {
+		tp.enqueueJob(job)
+	} else {
+		tp.syncProcessJob(job)
+	}
+}
+
+func (tp tracePoller) syncProcessJob(job PollingRequest) {
+	run := job.run
```
schoren, line 131: this is not a great go pattern. I'd remove this and directly access the values with the extra `run.`

server/model/spans.go
```diff
@@ -323,3 +160,46 @@ func decodeChildren(parent *Span, children []encodedSpan) ([]*Span, error) {
...
+	span := Span{
+		ID:         IDGen.SpanID(),
+		Name:       "Tracetest Trigger",
+		StartTime:  run.CreatedAt,
+		EndTime:    run.ServiceTriggerCompletedAt,
+		Attributes: Attributes{},
+		Children:   []*Span{},
+	}
+
+	span.setMetadataAttributes()
```
schoren, line 201: is this working correctly? The code is making a call to a reference method, but the var is not a reference

schoren, line 201: I would replace the pointer receivers with a by value calls:


```
func (span Span) setTriggerResultAttributes(run *Run) Span {
  // do things on span...
  return span //this is now updated
}
// ...
span = span.setMetadataAttributes()
```

server/model/traces.go
```diff
@@ -0,0 +1,182 @@
...
+type Trace struct {
+	ID       trace.TraceID
+	RootSpan Span
+	Flat     map[trace.SpanID]*Span `json:"-"`
+}
+
+func NewTrace(traceID string, spans []Span) Trace {
+	spanMap := make(map[string]*Span, 0)
+	for _, span := range spans {
+		spanCopy := span
```
schoren, line 21: With my suggested changes above, you can do this:
```
spanCopy := span.setMetadataAttributes()
spanID := span.ID.String()
spanMap[spanID] = &spanCopy
```

server/tracedb/tracedb.go
```diff
@@ -21,20 +20,23 @@ const (
...
 	Close() error
 }
 
 var ErrInvalidTraceDBProvider = fmt.Errorf("invalid traceDB provider: available options are (jaeger, tempo)")
 
 func New(c config.Config, repository model.RunRepository) (db TraceDB, err error) {
 	selectedDataStore, err := c.DataStore()
+
+	if selectedDataStore == nil && err == nil {
+		return nil, nil
```
schoren, line 33: Returning nil here will force the rest of the code to do nil checks before accessing the datastore. Returning a `noopDataStore` is a safer approach. You can copy one from here: https://github.com/kubeshop/tracetest/blob/c78b2737da070ca529d2bd5f6443bae64f60719a/server/tracedb/tracedb.go#L28-L33

schoren (changes requested):
Most of my comments are more personal opinions than facts, so use your judgement.

The only required change is the one of using a noopDataStore, because that can create nil pointer errors

xoscar:
Hey @schoren thanks for the review man! I just pushed some of the cleanup changes 😄

schoren (approved):
Looks great!

PR #1727, opened 2022-12-21
feat: create/update/list data stores via CLI

mathnogueira:
This PR adds a command `tracetest datastore apply -f <file>` which creates/updates a data store based on the file. And also creates a command for listing data stores

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/testdb/data_stores.go
+++ b/server/testdb/data_stores.go
@@ -76,7 +76,7 @@ func (td *postgresDB) UpdateDataStore(ctx context.Context, dataStore model.DataS
 
 	// keep the same creation date to keep sort order
 	dataStore.CreatedAt = oldDataStore.CreatedAt
-	dataStore.ID = dataStore.Slug()
+	dataStore.ID = oldDataStore.ID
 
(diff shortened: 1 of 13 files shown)


server/testdb/data_stores.go
```diff
@@ -76,7 +76,7 @@ func (td *postgresDB) UpdateDataStore(ctx context.Context, dataStore model.DataS
 
 	// keep the same creation date to keep sort order
 	dataStore.CreatedAt = oldDataStore.CreatedAt
-	dataStore.ID = dataStore.Slug()
+	dataStore.ID = oldDataStore.ID
```
mathnogueira, line 79: I removed the auto generated ID when updating a data store because this might lead to inconsistencies. Even with the CLI it was a weird behavior because if you run two updates changing the name of the data store, the second one fails because the first changed the data store id.

jorgeepc (approved):
LGTM!

PR #1730, opened 2022-12-22
fix(server): replace fallback DataStore with "create on first run" approach

schoren:
This PR changes the way we handle initial datastore config. Now, on first server run, if there is any datastore configured on the config file, it will insert it to the db.
## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/app/app.go
+++ b/server/app/app.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net/http"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/gorilla/handlers"
 	"github.com/kubeshop/tracetest/server/analytics"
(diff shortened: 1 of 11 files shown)


mathnogueira (approved):

server/app/app.go
```diff
@@ -102,11 +102,14 @@ func (a *App) Start() error {
 		return err
 	}
 
+	fmt.Println("New install?", isNewInstall)
```
xoscar, line 105: 😬

xoscar (approved):

jorgeepc (approved):

PR #1738, opened 2022-12-22
chore: Speedup docker multiarch build

schoren:
This PR changes the build pipeline to use more of goreleaser's features, and improving cache and parallelization. The end goal is to have faster multi arch docker builds, but also speed our pipelines in general.

This PR changes the deploy main and release pipelines. This has not been totally tested because the only way to really test is is to do a real deploy main and release. It has been tested up to a point where it shouldn't be different from the previos pipeline.

Example runss:
- Release: https://github.com/schoren/tracetest/actions/runs/3761226129
- Deploy main: https://github.com/schoren/tracetest/actions/runs/3761226125

In both examples the pipeline fails because of invalid credentials, so everything works until the actual publishing of artifacts.


.github/workflows/deploy-main.yml
```diff
@@ -1,4 +1,4 @@
-name: Deploy main
+name: Deploy main 2
```
mathnogueira, line 1: I suppose that was a test

mathnogueira (approved):

Dockerfile
```diff
@@ -1,58 +1,15 @@
...
-RUN cd cli && go mod download
-
-COPY ./cli ./cli
-
-COPY .goreleaser.yaml .
-RUN cd cli && make build
-
 FROM alpine AS release
+# Enable machine-id on alpine-linux (https://gitlab.alpinelinux.org/alpine/aports/-/issues/8761)
+RUN apk add dbus
```
danielbdias, line 3: We removed this lib on purpose because we were having a lot of problems with duplicate machineids on docker at the beginning of January. The PR with the fallback is here: https://github.com/kubeshop/tracetest/pull/1797

danielbdias (approved):

PR #1741, opened 2022-12-26
chore: update tracetest github action to use transactions to simplify test scripts

danielbdias:
This PR aims to simplify our `tracetest` step on Github Actions to rely more on the transaction infra instead of unique tests.

## Changes

- Update structure inside on `./tracetesting` folder to use tracetest "transaction" structure to create test suites, moving test suite control from the set o bash scripts to tracetest server
- Removed fixed ids from tests to guarantee that two test runs won't affect each other

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


danielbdias:
This first trigger was just made to force gitlab actions run and ensure that everything is right. The next commits will have the changes to the process.

mathnogueira (changes requested):
Looks great. Just try to improve the assertions I mentioned

mathnogueira (approved):

PR #1756, opened 2023-01-02
fix(cli): changing dotenv file validation to check if a file exists

danielbdias:
This PR aims to fix the dotenv file validation that we have on CLI, to check if a file really exists instead of relying only on the filename having a `./` prefix.

This allow an user to call `tracetest` CLI using a relative path like this:
```sh
tracetest test run -d test_definition.yaml  -e my_env_file.env --wait-for-result
```

## Changes

- Updated file validation to check if a string is a file path by checking if a file really exists.
- Update dogfood to add validation on these changes

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -287,24 +287,24 @@ jobs:
           TARGET_URL="http://tracetest-pr-${{ github.event.pull_request.number }}.tracetest-pr-${{ github.event.pull_request.number }}:11633" \
           ./run.bash
 
-  cleanup:
-    runs-on: ubuntu-latest
-    needs: [trace-testing, e2e]
-    if: always()
-    steps:
-      - uses: google-github-actions/setup-gcloud@94337306dda8180d967a56932ceb4ddcf01edae7
-        with:
-          service_account_key: ${{ secrets.GKE_SA_KEY }}
-          project_id: ${{ secrets.GKE_PROJECT }}
+  # cleanup:
+  #   runs-on: ubuntu-latest
+  #   needs: [trace-testing, e2e]
+  #   if: always()
+  #   steps:
+  #     - uses: google-github-actions/setup-gcloud@94337306dda8180d967a56932ceb4ddcf01edae7
+  #       with:
+  #         service_account_key: ${{ secrets.GKE_SA_KEY }}
+  #         project_id: ${{ secrets.GKE_PROJECT }}
 
-      - uses: google-github-actions/get-gke-credentials@fb08709ba27618c31c09e014e1d8364b02e5042e
-        with:
-          cluster_name: ${{ secrets.GKE_CLUSTER }}
-          location: ${{ secrets.GKE_ZONE }}
-          credentials: ${{ secrets.GKE_SA_KEY }}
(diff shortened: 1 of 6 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -289,8 +283,7 @@ jobs:
       - name: Run integration tests
         run: |
           cd tracetesting
-          TRACETEST_CLI_MAIN="/tmp/tracetest-main/cli/dist/tracetest" \
-          TRACETEST_CLI_TARGET="../cli/tracetest" \
+          TRACETEST_CLI="../cli/tracetest" \
```
danielbdias, line 286: I'm changing our dogfood process to use the PR CLI instead of the main CLI. By doing it we will be able to dogfood the CLI too.

tracetesting/run.bash
```diff
@@ -25,16 +25,14 @@ echo "Preparing to run tests on API..."
...
-TRACETEST_CLI_MAIN=$TRACETEST_CLI_MAIN
+echo "TRACETEST_CLI:      $TRACETEST_CLI"
+echo "TARGET_URL:         $TARGET_URL"
+echo "TRACETEST_ENDPOINT: $TRACETEST_ENDPOINT"
+echo "DEMO_APP_URL:       $DEMO_APP_URL"
+echo "DEMO_APP_GRPC_URL:  $DEMO_APP_GRPC_URL"
+
+cat << EOF > .env
 TARGET_URL=$TARGET_URL
-TRACETEST_MAIN_ENDPOINT=$TRACETEST_MAIN_ENDPOINT
```
danielbdias, line 37: Since this variable is only used on this script, does not make sense to add it on `.env` file.

tracetesting/run.bash
```diff
@@ -61,7 +59,7 @@ run_test_suite_for_feature() {
   junit_output='results/'$feature'_test_suite.xml'
   definition='./features/'$feature'/_test_suite.yml'
 
-  $TRACETEST_CLI_MAIN --config ./config.main.yml test run --definition $definition --environment ./.main.env --wait-for-result --junit $junit_output
+  $TRACETEST_CLI --config config.yml test run --definition $definition --environment .env --wait-for-result --junit $junit_output
```
danielbdias, line 62: Since we don't have to connect on `integration` and `target` environment at same time anymore, I've removed the references for `tracetest-main` for now.
Later we can add a README here to explain better how this dogfooding test is working.

.github/workflows/pull-request.yaml
mathnogueira: This might break things. What happens if we change something in the server/CLI which is not retro-compatible? The PR CLI will not be able to communicate with the integration instance (which is deployed using the latest main code)

mathnogueira (commented):
Looks good, only one thing that can go wrong.

.github/workflows/pull-request.yaml
danielbdias: Makes sense! I'll revert the changes and keep our dogfooding using only the stable/main CLI for now. Later we can add a step just to test the CLI better.

mathnogueira (approved):

Issue #1739, opened 2022-12-23
Still collecting and showing trace info when in no trace mode 


kdhamric:
When setup to use the OpenTelemetry Collector, if you delete the configuration and enter the no trace mode, you sometimes get a trace or pieces of spans from the trace. Expected behavior would be that you would only receive the top 'response' span.

Showing getting a full trace while in no trace mode:
https://www.loom.com/share/a11e1a97743c4496acd513b3bf6893b7

Showing getting a messaging span while in no trace mode:
![image](https://user-images.githubusercontent.com/1385894/209348633-421cefbb-8c91-425b-8969-57a495d3abb0.png)

mathnogueira:
I finally understood why that's happening. Because even though the data store is not configured, tracetest is still running the OTLP server (which is always on) and your otel collector is stlll sending the trace to it.

So the poller is working fine. The problem is that the trace is being updated in the background when the collector sends the trace.

We would probably need a way of disabling that server to prevent it from updating the test run trace

kdhamric:
Can we just ‘dump’ the packets unless the trace backend is set to
opentelemetry collector?

kdhamric:
Should we just 'dump' the spans as they come in unless we have the Opentelemetry collector mode set to on?

PR #1758, opened 2023-01-02
fix(backend): ignoring incoming otel collector traffic when data store is not enabled

xoscar:
This PR adds a check to the Export function that is part of the otlp listener to validate if a default data store is enabled and its the otlp one.

## Changes

- Adds otlp default data store validation

## Fixes

- #1739 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/71080bf4b5a64192a2867bbdb148b1b1

### DIFF
--- a/local-config/collector.config.yaml
+++ b/local-config/collector.config.yaml
@@ -33,5 +33,5 @@ service:
   pipelines:
     traces/1:
       receivers: [otlp]
-      processors: [tail_sampling, batch]
+      processors: [batch]
       exporters: [otlp/1]
--- a/server/otlp/server.go
+++ b/server/otlp/server.go
@@ -18,12 +18,12 @@ type Server struct {
 	pb.UnimplementedTraceServiceServer
 
 	addr string
-	db   model.RunRepository
+	db   model.Repository
 
 	gServer *grpc.Server
 }
 
-func NewServer(addr string, db model.RunRepository) *Server {
+func NewServer(addr string, db model.Repository) *Server {
 	return &Server{
 		addr: addr,
 		db:   db,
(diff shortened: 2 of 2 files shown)


local-config/collector.config.yaml
```diff
@@ -33,5 +33,5 @@ service:
   pipelines:
     traces/1:
       receivers: [otlp]
-      processors: [tail_sampling, batch]
```
xoscar, line 36: I removed this because its not in use anymore, we need to work on a fix if we want to add it back

server/otlp/server.go
```diff
@@ -18,12 +18,12 @@ type Server struct {
...
 
 	addr string
-	db   model.RunRepository
+	db   model.Repository
 
 	gServer *grpc.Server
 }
 
-func NewServer(addr string, db model.RunRepository) *Server {
+func NewServer(addr string, db model.Repository) *Server {
```
xoscar, line 26: We are passing the whole db model to the constructor function from the `app.go` file so I just changed the type to have access to the `DefaultDataStore` function

mathnogueira (approved):

server/otlp/server.go
```diff
@@ -45,6 +45,13 @@ func (s *Server) Stop() {
 }
 
 func (s Server) Export(ctx context.Context, request *pb.ExportTraceServiceRequest) (*pb.ExportTraceServiceResponse, error) {
+	ds, err := s.db.DefaultDataStore(ctx)
+
+	if err != nil || ds.Type != "otlp" {
```
xoscar, line 50: Ignoring if no data store or if is not type otlp

danielbdias (approved):

Issue #1643, opened 2022-12-06
[Chaining V2] Prompt for values when running a test or transaction with undefined variables


kdhamric:
When a user runs a test or a transaction, any variables that will be needed but are not defined by the time it is used should be prompted for.

Here are some use cases:
- a user wants a test or transaction they can run on a particular user, order id, etc that is configurable at run time. This would make running an adhoc test in an environment, even production, very easy and convenient. In this case, the user would reference the variable, but not add it to the environment. Each time they run the test or transaction, they would be prompted for the unspecified variables.
- a user wants to define 3 tests as part of a transaction. The first test has an output variable, and this is used by the second test. They define the first test. They then go to define the second test, and reference the variable. When they run the second test initially it will not work because the variable is not defined as they are running outside of a transaction. They could alter the environment to hard code the value, create the full transaction, then remove the environmental variable, but this is a hack and not user friendly.

This capability should be present on both the UI and CLI. On the CLI, we would need options to:
- interactively answer any undefined vars
- specify any variables that will be undefined by passing them as part of the cli vs waiting for a prompt (not sure about this... could just pass an env with the vars defined - we need to discuss this)
- for both UI and CLI, the promoting screen should show the variable name and ask for the value. If there have been previous runs of the test, the value of the variable for the last run should be used if available.

xoscar:
- What happens when a user creates a brand new test with missing variables (create & run scenario)?
- Work on diagrams and use cases/edge cases for this ticket.

xoscar:
The following sequence diagrams describe how the two new services will work to get the missing variables from a test or a transaction.

## Test:

```mermaid
sequenceDiagram
    Frontend->>+Server: GET /${testId}/variables?environmentId=<envId>
    Server->>+Database: Fetch Test
    Database-->>-Server: Return test
    alt is environment Id present
        Server->>+Database: Fetch Environment
        Database-->>-Server: Return Environment
        Server->>Server: Initializes env from database result
    else
        Server->>Server: Initializes empty env
    end
    Server->>Server: Adds test outputs to the env
    Server->>Server: Analyzes test looking for interpolations
    Server->>Server: Generates variables based on interpolations
    Server->>Server: Matches variables with env to find missing entries
    Server-->>-Frontend: Returns env, variables and missing
```

## Transaction:

```mermaid
sequenceDiagram
    Frontend->>+Server: GET /{tId}/variables?environmentId=<envId>
    Server->>+Database: Fetches Transaction
[...]

PR #1766, opened 2023-01-04
feature(backend): adding variables services for tests and transactions

xoscar:
This PR adds the variables services to both transactions and tests to return the environment variables required from the test, what is already set and which ones are missing.

## Changes

- Adds new API services for both transactions and tests
- Creates a new Variables struct to handle environment and spec env interpolations

## Fixes

- #1643 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

# Diagrams
The following sequence diagrams describe how the two new services will work to get the missing variables from a test or a transaction.

## Test:

```mermaid
sequenceDiagram
    Frontend->>+Server: GET /${testId}/variables?environmentId=<envId>
    Server->>+Database: Fetch Test
    Database-->>-Server: Return test
    alt is environment Id present
        Server->>+Database: Fetch Environment
        Database-->>-Server: Return Environment
        Server->>Server: Initializes env from database result
    else
        Server->>Server: Initializes empty env
    end
    Server->>Server: Adds test outputs to the env
    Server->>Server: Analyzes test looking for interpolations
    Server->>Server: Generates variables based on interpolations
    Server->>Server: Matches variables with env to find missing entries
    Server-->>-Frontend: Returns env, variables, and missing
```

## Transaction:

```mermaid
sequenceDiagram
    Frontend->>+Server: GET /{tId}/variables?environmentId=<envId>
    Server->>+Database: Fetches Transaction
    Database-->>-Server: Returns Transaction
    alt is environment Id present
        Server->>+Database: Fetch Environment
        Database-->>-Server: Return Environment
        Server->>Server: Initializes env from database result
    else
        Server->>Server: Initializes empty env
    end
    loop For Every Test in Transaction
        Server->>+Database: Fetches Test
        Database-->>-Server: Returns Test
        Server->>Server: Adds test outputs to the env
        Server->>Server: Analyzes test looking for interpolations
        Server->>Server: Generates variables based on interpolations
        Server->>Server: Matches variables with env to find missing entries
        Server->>Server: Update environment store for next tests
    end
    Server-->>-Frontend: Returns env, variables and missing
```
https://www.loom.com/share/6eb73136470947d8ad1966629392bea9


server/http/controller.go
```diff
@@ -1221,3 +1225,92 @@ func (c *controller) TestConnection(ctx context.Context, dataStore openapi.DataS
...
+	if err != nil {
+		return openapi.Response(http.StatusInternalServerError, err.Error()), err
+	}
+
+	return openapi.Response(200, c.mappers.Out.TestVariables(testVariables)), nil
+}
+
+func processTestVariables(ctx context.Context, test model.Test, variables expression.Variables, triggerRegistry *trigger.Registry) (expression.TestVariables, error) {
+	environmentVariables := variables.GetEnvironmentVariables(test)
+	specVariables, err := variables.GetSpecsVariables(test)
```
jorgeepc, line 1261: Do we need a similar step for `Outputs`? Similar to specs we can reference a `env:` var in an output value, right?

xoscar: You are right, I'm adding it

jorgeepc (commented):
Also saw in you loom that we get duplicated variables in the response. What do you think about getting unique values?

server/http/controller.go
xoscar: Done 😄

kdhamric:
One requirement is to provide a default for each missing variable based on the previous run of this same test - I did not see that covered in the loom video:

requirement:
for both UI and CLI, the promoting screen should show the variable name and ask for the value. If there have been previous runs of the test, the value of the variable for the last run should be used if available.

xoscar:
> One requirement is to provide a default for each missing variable based on the previous run of this same test - I did not see that covered in the loom video:
> 
> requirement: for both UI and CLI, the promoting screen should show the variable name and ask for the value. If there have been previous runs of the test, the value of the variable for the last run should be used if available.

@kdhamric this is only the API implementation to grab the missing variables, we can ensure that connection when working on the client side. This is not a fully ready feature yet.

xoscar:
> Also saw in you loom that we get duplicated variables in the response. What do you think about getting unique values?

Done 😄

danielbdias (commented):
At first glance, the code is ok for me! I have two questions about it: 
1) the idea of using [samber/lo](https://github.com/samber/lo) is because of some parts of repetitive code, like in the grpc and HTTP executor, right? 
2) Do you have an example of an API call that we could do to test this feature? (rationale behind it: if we can make a curl exercising this use case, maybe we can add a dogfood/tracetesting case in our CI later)

server/executor/trigger/grpc.go
```diff
@@ -144,6 +145,52 @@ func (t *grpcTriggerer) Resolve(ctx context.Context, test model.Test, opts *Trig
...
+	grpc := test.ServiceUnderTest.GRPC
+
+	if grpc == nil {
+		return triggerVariables, fmt.Errorf("no settings provided for GRPC triggerer")
+	}
+	address, err := executor.StatementTermsByType(WrapInQuotes(grpc.Address, "\""), expression.EnvironmentType)
+	if err != nil {
+		return triggerVariables, err
+	}
+	triggerVariables = triggerVariables.MergeStringArray(address)
```
danielbdias, line 159: Small / not important suggestion: to make it easier to read it makes sense to group the guard clauses if the statements that we are checking, like the `grpc := ...` line and `if grpc == nil { ...` line

```suggestion
	triggerVariables := expression.VariablesMap{}
	
	grpc := test.ServiceUnderTest.GRPC
	if grpc == nil {
		return triggerVariables, fmt.Errorf("no settings provided for GRPC triggerer")
	}
	
	address, err := executor.StatementTermsByType(WrapInQuotes(grpc.Address, "\""), expression.EnvironmentType)
	if err != nil {
		return triggerVariables, err
	}
	
	triggerVariables = triggerVariables.MergeStringArray(address)
```

server/executor/trigger/http.go
```diff
@@ -199,3 +199,70 @@ func mapResp(resp *http.Response) model.HTTPResponse {
...
+	if http == nil {
+		return triggerVariables, fmt.Errorf("no settings provided for HTTP triggerer")
+	}
+
+	urlVariables, err := executor.StatementTermsByType(WrapInQuotes(http.URL, "\""), expression.EnvironmentType)
+
+	if err != nil {
+		return triggerVariables, err
+	}
+	triggerVariables = triggerVariables.MergeStringArray(urlVariables)
```
danielbdias, line 216: Same suggestion as the gRPC endpoint:

```suggestion
	triggerVariables := expression.VariablesMap{}
	
	http := test.ServiceUnderTest.HTTP
	if http == nil {
		return triggerVariables, fmt.Errorf("no settings provided for HTTP triggerer")
	}

	urlVariables, err := executor.StatementTermsByType(WrapInQuotes(http.URL, "\""), expression.EnvironmentType)
	if err != nil {
		return triggerVariables, err
	}
	
	triggerVariables = triggerVariables.MergeStringArray(urlVariables)
```

server/executor/trigger/grpc.go
xoscar: Done 😄

xoscar:
Hey @danielbdias thanks for the review man, the PR should be ready for its final review now 😄

danielbdias (approved):
:shipit:

PR #1769, opened 2023-01-04
feat(frontend): redesign test outputs

### LINKED ISSUE #1573: [Chaining V2] Redesign Test Outputs modal and location
Currently, the defining a test output takes you to the Trigger page and the outputs are shown in a tab. There are a couple issues with this:
- when defining an output, you typically are working with attributes on either the Trace or Test page, and you may need to see the data contained and shown in the attribute when defining it. As an example, if you wanted to extract a value from the json in an attribute, viewing the json while creating the output is helpful.
- It would be helpful to have an indication that a particular attribute in a span is involved in setting an output. As user could be looking at an attribute and think they need to create a variable based on it, when in actuality they have already created one. This is covered in #1565

jorgeepc:
This PR includes the new design for `Test Outputs`.

## Changes

- New `Test Outputs` design consistent with `Test Specs`
- Unified `revert`/`publish` actions for `Test Outputs` and `Test Specs`

## Fixes

- fixes #1573 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom

https://www.loom.com/share/b334b43d81494c2aaf849a6057903f9d


web/cypress/e2e/TestRunDetail/Outputs.spec.ts
```diff
@@ -94,7 +93,7 @@ describe('Outputs', () => {
     cy.get('[data-cy=output-pending-tag]').should('have.length', 1);
 
     // Revert
-    cy.get('[data-cy=output-reset-button]').click();
+    cy.get('[data-cy=trace-actions-revert-all]').click();
```
danielbdias, line 96: I don't know if here is a good place for it, but question: by applying the "Revert All" action, it makes sense to query the user before doing it (something like "Are you sure that you want to revert the last X changes?")?
It seems that this is the type of action that if the user clicks by mistake they can undo a good bunch of their work.

danielbdias (approved):
It sounds ok! I left just one question about usability, which if makes sense, we can see on other PRs.

web/src/components/RunDetailLayout/HeaderRight.tsx
```diff
@@ -9,21 +9,25 @@ import {useTestRun} from 'providers/TestRun/TestRun.provider';
...
 
 interface IProps {
   testId: string;
   testVersion: number;
 }
 
 const HeaderRight = ({testId, testVersion}: IProps) => {
-  const {isDraftMode} = useTestSpecs();
+  const {isDraftMode: isTestSpecsPending} = useTestSpecs();
+  const isTestOutputsPending = useAppSelector(selectIsPending);
```
xoscar, line 22: Shouldn't be good to have a provider for outputs similar to what we have for test specs? or maybe a test draft provider which uses both 👀

web/src/components/RunDetailTest/RunDetailTest.tsx
```diff
@@ -77,8 +151,69 @@ const RunDetailTest = ({run, testId}: IProps) => {
...
+                  isLoading={isLoading}
+                  onCancel={onClose}
+                  onSubmit={onSubmitTestOutput}
+                  output={output}
+                  runId={run.id}
+                  testId={testId}
+                />
+              )}
+
+              {!isTestSpecFormOpen && !isTestOutputFormOpen && (
```
xoscar, line 168: It feels like we are trying to jiggle the two concepts and if feels weird haha.

Not for this PR but maybe we can consider consolidating a single draft mode where multiple things can be added to the bucket

web/src/components/TestActions/TestActions.tsx
```diff
@@ -1,13 +1,25 @@
...
-  const pendingCount = specs.filter(({isDraft}) => isDraft).length;
+interface IProps {
+  runId: string;
+  testId: string;
+}
+
+const TestActions = ({runId, testId}: IProps) => {
+  const {specs, publish, cancel: onCancelTestSpecs} = useTestSpecs();
+  const {onCancel: onCancelTestOutputs} = useTestOutput();
+  const outputs = useAppSelector(state => selectTestOutputs(state, testId, runId));
```
xoscar, line 19: I think this value should come from a provider similar to the specs one

web/src/components/TestResults/TestResults.tsx
```diff
@@ -90,23 +47,12 @@ const TestResults = () => {
       {!isLoading && Boolean(assertionResults) && (
         <TestSpecs
           assertionResults={assertionResults!}
-          onDelete={handleDelete}
-          onEdit={handleEdit}
+          onDelete={onDelete}
```
xoscar, line 50: Should these functions be defined at the TesSpecs component level instead?

xoscar (changes requested):
Hey @jorgeepc feature looks awesome! I have some questions and thinks to consider, let me know what you think my dude

web/src/components/RunDetailLayout/HeaderRight.tsx
jorgeepc: Done 👍  Moved that state to the provider. I have mixed feelings with adding the state (`outputs`) to the provider but it's ok. We should have consistency with the `TestSpecs` one. 

And totally agree with your idea of one single `test draft` provider. I didn't want to refactor test specs as part of this PR but I can work on that refactor in a new PR.

web/src/components/RunDetailTest/RunDetailTest.tsx
jorgeepc: Like the idea of one single draft mode.

web/src/components/TestActions/TestActions.tsx
jorgeepc: Done

web/src/components/TestResults/TestResults.tsx
jorgeepc: We are also using those functions at a higher level. Probably they can be moved to the TestSpecs provider.

xoscar (approved):
❤️

PR #1796, opened 2023-01-09
fix(server): update analytics unique id generation

danielbdias:
This PR aims to change how we generate the server id, to avoid using `machineid`. This is necessary because we don't have any guarantee that the `machineid` will be unique across many Tracetest server containers.

## Changes

- Change ServerID generation to use shortid instead of machineid

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/testdb/postgres.go
+++ b/server/testdb/postgres.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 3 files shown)


mathnogueira (approved):

danielbdias:
For now we opted for keeping the `machineid` strategy

PR #1797, opened 2023-01-09
fix: machine id fallback

mathnogueira:
This PR makes tracetest consider the hostname as a fallback in case it cannot get the machine-id from the host. It also will make old servers reset the server table, so all old users will get new server ids.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/config/machine-id.go
+++ b/server/config/machine-id.go
@@ -9,15 +9,25 @@ import (
 )
 
 func GetMachineID() string {
-	id, err := machineid.ProtectedID("tracetest")
-	// fallback to hostname based machine id in case of error
-	if err != nil {
-		name, err := os.Hostname()
-		if err != nil {
-			return "default"
-		}
-		sum := md5.Sum([]byte(name))
-		return hex.EncodeToString(sum[:])
+	id := getMachineID()
+	if len(id) > 10 {
+		return id[:10] // limit lenght to avoid issues with GA
 	}
+
 	return id
+}
+
+func getMachineID() string {
+	id, err := machineid.ProtectedID("tracetest")
+	if err == nil {
+		return id
+	}
+
+	// fallback to hostname based machine id in case of error
+	name, err := os.Hostname()
+	if err != nil {
+		return "default"
+	}
+	sum := md5.Sum([]byte(name))
+	return hex.EncodeToString(sum[:])
 }
--- a/server/config/machine-id.go
+++ b/server/config/machine-id.go
@@ -1,0 +1,23 @@
+package config
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"os"
+
+	"github.com/denisbrodbeck/machineid"
+)
+
+func GetMachineID() string {
+	id, err := machineid.ProtectedID("tracetest")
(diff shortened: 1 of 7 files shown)


server/config/machine-id.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/denisbrodbeck/machineid"
+)
+
+func GetMachineID() string {
+	id, err := machineid.ProtectedID("tracetest")
+	// fallback to hostname based machine id in case of error
+	if err != nil {
+		name, err := os.Hostname()
+		if err != nil {
+			return "default"
```
danielbdias, line 17: In a worst-case scenario, it make sense to use a random string here?

server/config/machine-id.go
```diff
@@ -0,0 +1,23 @@
...
+	"encoding/hex"
+	"os"
+
+	"github.com/denisbrodbeck/machineid"
+)
+
+func GetMachineID() string {
+	id, err := machineid.ProtectedID("tracetest")
+	// fallback to hostname based machine id in case of error
+	if err != nil {
```
danielbdias, line 14: Small suggestion to avoid one more nesting: we can invert the `if` here and do an early return:

```golang
func GetMachineID() string {
    id, err := machineid.ProtectedID("tracetest")
    if err == nil {
      return id
    } 
    // do fallback case
}
```

server/testdb/postgres.go
```diff
@@ -81,11 +81,7 @@ func (td *postgresDB) ServerID() (id string, isNew bool, err error) {
 
 	// no id, let's creat it
 	isNew = true
-	id, err = machineid.ProtectedID("tracetest")
-	if err != nil {
-		err = fmt.Errorf("could not get machineID: %w", err)
-		return
-	}
+	id = config.GetMachineID()
 	id = id[:10] // limit lenght to avoid issues with GA
```
danielbdias, line 85: Does make sense to move this `id` trim to `config.GetMachineID`?

server/testdb/postgres.go
```diff
@@ -81,11 +81,7 @@ func (td *postgresDB) ServerID() (id string, isNew bool, err error) {
 
 	// no id, let's creat it
 	isNew = true
-	id, err = machineid.ProtectedID("tracetest")
-	if err != nil {
-		err = fmt.Errorf("could not get machineID: %w", err)
-		return
-	}
+	id = config.GetMachineID()
```
danielbdias, line 84: Does make sense try to use [this unit test](https://github.com/kubeshop/tracetest/pull/1796/files#diff-c4d616c3d0d491af1cba1ba8028c52a51f97ffbb2461732204bb5936e8f0583b) here?

danielbdias (approved):
Sounds fine. To guarantee the we are using the hostname logic, it makes sense to remove the [`dbus` installation](https://github.com/kubeshop/tracetest/blob/main/Dockerfile#L49) from our image?

server/config/machine-id.go
mathnogueira: I rather have something that cannot be confused with a real id. So we can detect it

Issue #1777, opened 2023-01-06
Support for Elastic APM as a backend


michaelhyatt:
Elastic APM is another popular APM solution that has a free tier offering. It supports OpenTelemetry and can be used to store both: traces captured by OpenTelemetry agents and traces generated by Elastic own APM agents. Adding it as a supported backend should be easy since it should be very similar to the existing OpenSearch backend.

michaelhyatt:
Hi @kdhamric 

OpenSearch being a fork of Elasticsearch is effectively using the same API for document retrieval, so everything your team built for OpenSearch in terms of the retrieval of documents by their id (i.e. retrieval by the `trace.id`) will work. The only thing different that Jaeger and Elastic APM will be doing is field mapping or naming going from OpenTelemetry to Elastic Common Schema format used to store Elastic APM data.

I run the `quick-start-nodejs` test configuring the OTel collector to send data into Elastic APM. Here are the resulting 4 Elasticsearch documents representing the trace generated by the `Hello World` test app. You will notice the top level span being called `transaction` in Elastic APM, whereas the underlying spans will be called `span`.

```json
[
  {
    "_index": ".ds-traces-apm-default-2023.01.06-000001",
    "_id": "9IEQhoUBOBwak0ckjABO",
    "_version": 1,
    "_score": null,
    "fields": {
      "span.name": [
        "middleware - query"
      ],
      "labels.process_command": [
        "/usr/src/app/app.js"
      ],
      "labels.process_runtime_description": [
        "Node.js"
      ],
      "labels.process_executable_name": [
        "node"
      ],
      "labels.express_name": [
        "query"
      ],
      "trace.id": [
        "6386505b1d8f4cbd1b5aac0bb34b230a"
      ],
      "span.duration.us": [
        44
      ],
      "processor.event": [
        "span"
      ],
      "agent.name": [
        "opentelemetry/nodejs"
      ],
      "event.agent_id_status": [
        "missing"
      ],
      "event.outcome": [
        "unknown"
      ],
      "service.name": [
        "unknown_service_node"
      ],
      "data_stream.namespace": [
        "default"
      ],
      "processor.name": [
        "transaction"
      ],
      "span.subtype": [
        "internal"
      ],
      "labels.http_route": [
        "/"
      ],
      "span.id": [
        "f0175493026fc419"
      ],
      "observer.hostname": [
        "elastic-agent"
      ],
      "data_stream.type": [
        "traces"
      ],
      "span.type": [
[...]

PR #1805, opened 2023-01-10
feat: add elastic search as data store

xoscar:
This PR adds the frontend piece to support elastic APM as a data store, it is a partial update as it still requires the work done for the backend client to be fully wired up.

## Changes

- Updates the open search client to be a generic elastic search client
- Adds logic for the elastic search client to support both apm and open search

## Fixes

- #1777 

## Checklist

- [] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


api/dataStores.yaml
```diff
@@ -21,13 +21,15 @@ components:
...
+          $ref: "#/components/schemas/ElasticSearch"
+        elasticApm:
+          $ref: "#/components/schemas/ElasticSearch"
         signalFx:
           $ref: "#/components/schemas/SignalFX"
         createdAt:
           type: string
           format: date-time
-    OpenSearch:
+    ElasticSearch:
```
xoscar, line 32: Switched this to be a generic elastic search structure

server/config/config.go
```diff
@@ -86,11 +87,12 @@ type (
...
-		Addresses []string
-		Username  string
-		Password  string
-		Index     string
+	ElasticSearchDataStoreConfig struct {
+		Addresses   []string
+		Username    string
+		Password    string
+		Index       string
+		Certificate string
```
xoscar, line 95: Including the certificate string

server/testdb/data_stores.go
```diff
@@ -223,12 +225,36 @@ func (td *postgresDB) readDataStoreRow(ctx context.Context, row scanner) (model.
...
 
-		return dataStore, nil
+		return patchDataStoreValues(dataStore), nil
 	default:
 		return model.DataStore{}, err
 	}
 }
 
+func patchDataStoreValues(dataStore model.DataStore) model.DataStore {
+	if dataStore.Values.ElasticApm == nil {
```
xoscar, line 235: When adding the elastic APM object to the datastore values, it was causing an issue because the object was empty and when trying to run a deep copy it would fail, this check would patch any further incompatibility issues that we might encounter.

web/src/components/Settings/DataStorePlugin/forms/ElasticSearch/ElasticSearch.tsx
```diff
@@ -38,6 +39,14 @@ const OpenSearch = () => {
           </Form.Item>
         </Col>
       </Row>
+
+      <Row gutter={[16, 16]}>
```
xoscar, line 43: Adding new field for the certificate

server/tracedb/elasticsearchdb.go
```diff
@@ -0,0 +1,326 @@
...
+	spans := make([]model.Span, 0)
+	for _, result := range searchResponse.Hits.Results {
+		span := convertElasticSearchSpanIntoSpan(result.Source)
+		spans = append(spans, span)
+	}
+
+	return model.NewTrace(traceID, spans)
+}
+
+func convertElasticSearchSpanIntoSpan(input map[string]interface{}) model.Span {
```
danielbdias, line 194: Question (not important, out of curiosity): is there a reason on why we are parsing a span here using `flatten` and getting thought a map?

Other possibility it was create a temporary type for the Elastic span and use `json.Unmarshal` to get the values, in a code similar to this:

```golang
type elasticDuration struct {
    Us float64 `json:"us"`
}

type elasticCommonInfo struct {
    Id string `json:"id"`
    Name string `json:"name"`
    Duration `json:"duration"`
}

type elasticSearchSpan struct {
    Transaction elasticCommonInfo `json:"transaction"`
    Span elasticCommonInfo `json:"span"`
    Timespan string `json:"@transaction"`
}

// to do that we need to think how to bring input as an array of bytes (maybe using json.RawMessage?)
func convertElasticSearchSpanIntoSpan(input []byte) model.Span {
    var esSpan elasticSearchSpan 
    err := json.Unmarshal(input, &esSpan)

    // after handling error we could retrieve this values like this
    transactionId := esSpan.Transaction.Id
    spanId := esSpan.Span.Id

    transactionName := esSpan.Transaction.Name
    spanName := esSpan.Span.Name

    transactionDuration := esSpan.Transaction.Duration.Us
    spanDuration := esSpan.Span.Duration.Us

    timespan = esSpan.Timespan
}

```

danielbdias (approved):

server/tracedb/elasticsearchdb.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+type FlattenOptions struct {
+	Prefix    string
+	Delimiter string
+	Safe      bool
+	MaxDepth  int
+}
+
+func flatten(prefix string, depth int, nested interface{}, opts *FlattenOptions) (flatmap map[string]interface{}, err error) {
```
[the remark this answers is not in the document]
xoscar: We need to consider this when we decide what to do for the BE refactoring, right now its too hard to understand where we can add helper functions like this

server/tracedb/elasticsearchdb.go
jorgeepc: Talked with @danielbdias about this. I think we can go with the current approach (flatten + map). The thing about creating a struct is that we don’t know all the fields of the span, also maybe it is not good try to parse the fields that we don't know, because they can change through elastic versions.

server/tracedb/elasticsearchdb.go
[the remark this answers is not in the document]
jorgeepc: We were talking about creating a `helpers` folder, but we need to sync with the team to see/define some standards.

Issue #1793, opened 2023-01-09
Add data stores for each trace provider we integrate with via OpenTelemetry Collector


kdhamric:
Currently, the trace data store configuration shows trace providers that we have direct integrations with and has the OpenTelemetry Collector tab for 'all others' that we cannot directly integrate with. We want to add vendors such as Lightstep and NewRelic that we have documented how to connect via the OpenTelemetry collector and show this information on an individual tab.

AC1:
As a user that uses Lightstep, 
I want to be able to see a Lightstep option in the trace data store configuration screeen
And select it and see exactly how to configure the OpenTelemetry collector to enable Lightstep

AC2:
As a user that selected Lightstep as my trace data store,
I want this selection to be respected when I come back to the trace data store configuration screen in the future
And that 'tab' should be selected by default

AC3:
The 'Need more information' link with be specific to the vendor and will go to a specific documentation page
 for each vendor under the 'connecting to data stores' area of our docs (@adnanrahic will be adding those this week)

AC4:
The trace data store configuration screen should be able to handle 15 possible choices in the future without a redesign.

AC5:
Our 'storage/persistence' will need to have new types added to it (NewRelic and Lightstep immediately)

AC6:
Add Lightstep as a configuration choice

AC7:
Add NewRelic as a configuration choice

AC8:
Use custom OTel collector configs for both NewRelic & Lightstep

AC9:
Change CLI to support new work

AC10:
Add New Relic & Lightstep under the Connect to Data Store docs

olha23:
added mockups with 2 options https://www.figma.com/file/LBN4SKVPq3ykegrPKbHT2Y/0.8-0.9-Release-Tracetest?node-id=1337%3A23100&t=I2JbRJQzCfpLJGMB-1

kdhamric:
Lets go with option #2
![image](https://user-images.githubusercontent.com/1385894/211414695-f9d62226-ecfc-4fd1-ae5a-9d53f62b0037.png)

olha23:
added svg icons:

![Image](https://user-images.githubusercontent.com/5725840/211657594-12708b8f-0d61-482d-9935-389574f58f0c.svg)
![Image](https://user-images.githubusercontent.com/5725840/211657593-adbab1d0-8544-42af-8b4f-3211685851df.svg)
![Image](https://user-images.githubusercontent.com/5725840/211657592-37fbe819-d241-4bc7-ae77-84856cc91114.svg)

PR #1815, opened 2023-01-11
feature: adding support for new relic and lighstep

xoscar:
This PR adds support for both new relic and light step to show up as data store options under the settings page.

## Changes

- Updates the UI to show a different input for the list
- Adds new relic and lightstep

## Fixes

- #1793 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/4c289d7351f846029d2521958029a15b


server/model/data_stores.go
```diff
@@ -77,6 +79,8 @@ func DataStoreFromConfig(dsc config.TracingBackendDataStoreConfig) DataStore {
 		ds.Type = openapi.OPEN_SEARCH
 	case signalfx:
 		ds.Type = openapi.SIGNAL_FX
+	case newrelic:
+	case lighstep:
```
jorgeepc, line 83: If you have `newrelic` or `lighstep` in your config file, is this going to change the type to `otlp`? Do we want that behavior?

jorgeepc (commented):
Looks great. I added one question and a minor suggestion: I would put `OpenSearch` and `Elastic` together in the list.

xoscar:
> Looks great. I added one question and a minor suggestion: I would put `OpenSearch` and `Elastic` together in the list.

I added them based on the list from the designs, I'll put them together 😄

server/model/data_stores.go
xoscar: Nice catch, fixed!

jorgeepc (approved):

PR #1835, opened 2023-01-13
Added an example with native Elastic APM agent

michaelhyatt:
This PR adds an example with Elastic APM Nodejs agent and Elastic APM stack.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


michaelhyatt:
I need to make a quick fix for the index name, will re-open it shortly.

PR #1840, opened 2023-01-13
feat(server): add traceid trigger

### LINKED ISSUE #1824: [k6] Trafcetest trace id trigger support
As part of supporting k6 as a new trigger, we need to allow users to create tests whose trigger includes a trace id that tracetest will try to find in the configured data store.

## Acceptance Criteria
AC1:
As a user utilizing the tracetest CLI 
I should be able to create a test with trigger type for TRACE_ID
And the trace id input could be an environment variable

AC2:
As a user utilizing the tracetest CLI 
I should be able to run a test with TRACE_ID as trigger

## Sequence Diagram

```mermaid
sequenceDiagram
    Client->>+Server: Createst test with trace id as trigger
    Server->>+Database: Saves test
    Database-->>-Server: Returns Test
    Server-->>-Client: Test Object
    Client->>+Server: Fetch Test variables
    Server->>-Client: Returns Tests Variables
    alt Test uses Variables
        Client-->>Client: Adds variables to run request body
    end
    Client->>+Server: Runs Test with trace id
    Server->>Server: Executes noop trigger
    Server-->>+Database: Saves trigger results to DB
    Server->>Server: Gets configured trace DB
    Server->>Server: Polls for trace id
    alt Was trace found
        Server->>Server: Runs Test specs
        Server-->>+Database: Saves test specs results to DB
    end
    Server-->>-Client: Returns test result
```

jorgeepc:
This PR adds support for a new type of `test trigger` based on a `traceID`. This is part of the work needed to interact with the k6 extension. You can find the details and a sequence diagram here https://github.com/kubeshop/tracetest/issues/1824

## Changes

- New `test trigger` based on `traceID`

## Fixes

- fixes #1824 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom

https://www.loom.com/share/48f44a7d29a144f18c63bd5cc46e384f


server/executor/runner.go
```diff
@@ -159,6 +158,13 @@ func (r persistentRunner) processExecQueue(job execReq) {
 		panic(err)
 	}
 
+	if job.test.ServiceUnderTest.Type == model.TriggerTypeTRACEID {
+		traceIDFromParam, err := trace.TraceIDFromHex(job.test.ServiceUnderTest.TRACEID.ID)
+		if err == nil {
```
danielbdias, line 163: Question: If we get an error here, do we want to continue this test? 

At first glance, the test will not work or give inconsistent results in the error case. If we have another block, like [this](https://github.com/kubeshop/tracetest/pull/1840/files#diff-999879534cbd7435721e05ccc9349cc71a860b82fb1e36172a7e128d2e4d0fa6R34), maybe we could add a comment here signalizing this.

server/executor/trigger/traceid.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if traceid == nil {
+		return test, fmt.Errorf("no settings provided for TRACEID triggerer")
+	}
+
+	id, err := opts.Executor.ResolveStatement(WrapInQuotes(traceid.ID, "\""))
+
+	if err != nil {
+		return test, err
+	}
```
danielbdias, line 41: Sometimes to associate an error handling with the code that caused the error we put the code together:
(It is not a standard on Go as far as I saw, but I saw a bunch of Go programmers doing that to help this association)

```suggestion
	traceid := test.ServiceUnderTest.TRACEID
	if traceid == nil {
		return test, fmt.Errorf("no settings provided for TRACEID triggerer")
	}

	id, err := opts.Executor.ResolveStatement(WrapInQuotes(traceid.ID, "\""))
	if err != nil {
		return test, err
	}
```

server/executor/trigger/traceid.go
```diff
@@ -0,0 +1,66 @@
...
+func (t *traceidTriggerer) Type() model.TriggerType {
+	return model.TriggerTypeTRACEID
+}
+
+func (t *traceidTriggerer) Resolve(ctx context.Context, test model.Test, opts *TriggerOptions) (model.Test, error) {
+	traceid := test.ServiceUnderTest.TRACEID
+
+	if traceid == nil {
+		return test, fmt.Errorf("no settings provided for TRACEID triggerer")
+	}
```
danielbdias, line 35: Should we test here if traceid.Id is empty too? Something like:

```suggestion
	if traceid == nil || traceid.Id == "" {
		return test, fmt.Errorf("no settings provided for TRACEID triggerer")
	}
```

server/executor/runner.go
```diff
@@ -159,6 +158,13 @@ func (r persistentRunner) processExecQueue(job execReq) {
 		panic(err)
 	}
 
+	if job.test.ServiceUnderTest.Type == model.TriggerTypeTRACEID {
```
mathnogueira, line 161: So you either have to choose if you want an HTTP test or a test that you provide the trace?

I remember some discussions that we had in the past where we considered it would be nice to be able to provide a trace ID for any kind of test. This way, you could have your test in your CI, but if you found a problem in production, you could run the same test by just referencing the trace ID from production.

xoscar: @mathnogueira that would mean that we'll need a way to create tests without trigger information and a way of test creation only without running. In the case of k6 the idea is to have a static number of tests (created using the tracetest extension) and N number of runs based on the load tests.

api/traceid.yaml
```diff
@@ -0,0 +1,14 @@
+openapi: 3.0.0
+components:
+  schemas:
+    TRACEIDRequest:
```
xoscar, line 4: If these objects are the same should we keep the separated?

xoscar (approved):
Hey @jorgeepc good job man! I left some non-blocking comments

web/src/components/CreateTestPlugins/TraceID/steps/Value/Value.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }, [draftTest, form, onIsFormValid]);
+
+  useEffect(() => {
+    onRefreshData();
+  }, [onRefreshData]);
+
+  return (
+    <Step.Step>
+      <Step.FormContainer>
+        <Step.Title>Enter a trace id or an expression</Step.Title>
```
danielbdias, line 40: Does it make sense to give more emphasis to these words?

```suggestion
        <Step.Title>Enter a Trace ID or an Expression</Step.Title>
```

web/src/components/CreateTestPlugins/TraceID/steps/Value/ValueForm.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Editor from 'components/Editor';
+import {SupportedEditors} from 'constants/Editor.constants';
+
+const ValueForm = () => {
+  return (
+    <S.InputContainer>
+      <Form.Item
+        label="Trace ID"
+        name="id"
+        rules={[{required: true, message: 'Please enter a trace id or an expression'}]}
```
danielbdias, line 13:
```suggestion
        rules={[{required: true, message: 'Please enter a Trace ID or an Expression'}]}
```

web/src/components/CreateTestPlugins/TraceID/steps/Value/ValueForm.tsx
```diff
@@ -0,0 +1,21 @@
...
+      >
+        <Editor type={SupportedEditors.Interpolation} placeholder="Trace id or expression" />
```
danielbdias, line 15:
```suggestion
        <Editor type={SupportedEditors.Interpolation} placeholder="Trace ID or Expression" />
```

danielbdias (approved):

api/traceid.yaml
jorgeepc: I was thinking that probably we want to include extra data in the response. Not really needed for now.

web/src/components/RunDetailTriggerResponse/RunDetailTriggerResponse.tsx
```diff
@@ -62,12 +64,16 @@ const RunDetailTriggerResponse = ({
             updateQuery([['tab', newTab]]);
           }}
         >
-          <Tabs.TabPane key={TabsKeys.Body} tab="Body">
-            <ResponseBody body={body} bodyMimeType={bodyMimeType} />
-          </Tabs.TabPane>
-          <Tabs.TabPane key={TabsKeys.Headers} tab="Headers">
-            <ResponseHeaders headers={headers} />
-          </Tabs.TabPane>
+          {type !== TriggerTypes.traceid && (
```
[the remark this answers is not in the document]
jorgeepc: Changed to approach to a Component factory based on the trigger type

web/src/components/CreateTestPlugins/TraceID/steps/Value/Value.tsx
jorgeepc: Done

web/src/components/CreateTestPlugins/TraceID/steps/Value/ValueForm.tsx
jorgeepc: Done

PR #1841, opened 2023-01-13
feat: add environment apply command

mathnogueira:
This PR introduces the command `tracetest environment apply -f file.yaml` to make the CLI more coherent 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/actions/apply_environment_action.go
+++ b/cli/actions/apply_environment_action.go
@@ -43,7 +43,7 @@ func (a applyEnvironmentAction) Run(ctx context.Context, args ApplyEnvironmentCo
 	}
 
 	if fileContent.Definition().Type != "Environment" {
-		return fmt.Errorf(`file must be of type "DataStore"`)
+		return fmt.Errorf(`file must be of type "Environment"`)
 	}
 
 	fileContentString := fileContent.Contents()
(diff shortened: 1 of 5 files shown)


cli/actions/apply_environment_action.go
```diff
@@ -0,0 +1,73 @@
...
+	)
+
+	fileContent, err := file.Read(args.File)
+	if err != nil {
+		return fmt.Errorf("could not read file: %w", err)
+	}
+
+	if fileContent.Definition().Type != "Environment" {
+		return fmt.Errorf(`file must be of type "DataStore"`)
+	}
```
danielbdias, line 47:
```suggestion
	if fileContent.Definition().Type != "Environment" {
		return fmt.Errorf(`file must be of type "Environment"`)
	}
```

danielbdias (approved):
The PR is ok! 

There is only a log that we need to change. If it is possible, can you add an explanation on PR description on how to use this command? At first glance, he remembers a similar behavior that `kubectx` has, where changes a context (in this case environment) and set it as a standard on the CLI.

mathnogueira:
@danielbdias instead of adding a description here, I updated the docs to explain how to use it.

@jfermi and @adnanrahic might want to take a look at this.

jfermi (approved):

Issue #1857, opened 2023-01-19
Change Trace Id based trigger to always ask for a variable name


kdhamric:
As a User,
When selecting a test with a TraceID based trigger,
I should be asked for a variable name, not a trace id number

Currently, we are asking for a trace id, and provide the user with the option to enter a variable name by entering ${env:mytraceid}

We want to change this so we always expect a variable name to be used when initially defining the traceID based trigger.

The third form in the wizard should use this text:
Title: Enter the variable name to use for the Trace Id

Subheading: Variable name:
text input: Default the value to 'traceId'
Help text: Please enter a variable name for the trace id or accept the default 'traceId' name

When the user clicks create, we should change the variable name to ${env:VARIABLENAMEENTERED} and that value should show when viewing the 'Edit Test' form. We will 'build' the expression properly and show it - they can alter it from there (ie see below)

![image](https://user-images.githubusercontent.com/1385894/213574389-3369f483-d3c3-42d3-b61f-afcc5ff76b61.png)

@olha23 let me know if you have thoughts....

olha23:
yea make sense, i will add the mockups

olha23:
@kdhamric do we need to have ability to add couple variables name? see mockups - https://www.figma.com/file/LBN4SKVPq3ykegrPKbHT2Y/0.8-0.9-Release-Tracetest?node-id=1495%3A25017&t=EgMwXPc0AZCYxZYy-1

PR #1860, opened 2023-01-20
feature(frontend): updating trace id trigger form

xoscar:
This PR updates the trigger trace form to ask users to add the variable name instead of the actual trace id, to have the trace id be always dynamic.

## Changes

- Updates the trace id trigger

## Fixes

- #1857 

## Checklist
- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

Screenshot 2023-01-20 at 10 03 53

### DIFF
--- a/web/src/components/CreateTestPlugins/TraceID/steps/VariableName/VariableName.tsx
+++ b/web/src/components/CreateTestPlugins/TraceID/steps/VariableName/VariableName.tsx
@@ -5,9 +5,9 @@ import * as Step from 'components/CreateTestPlugins/Step.styled';
 import {ComponentNames} from 'constants/Plugins.constants';
 import {useCreateTest} from 'providers/CreateTest/CreateTest.provider';
 import {ITraceIDValues} from 'types/Test.types';
-import ValueForm from './ValueForm';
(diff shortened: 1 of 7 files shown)


web/src/components/CreateTestPlugins/TraceID/steps/VariableName/VariableName.tsx
```diff
@@ -37,9 +37,10 @@ const Value = () => {
   return (
     <Step.Step>
       <Step.FormContainer>
-        <Step.Title>Enter a Trace ID or an Expression</Step.Title>
+        <Step.Title>Enter the variable name to use for the Trace Id</Step.Title>
```
jorgeepc, line 40:
```suggestion
        <Step.Title>Enter the variable name to use for the Trace ID</Step.Title>
```

jorgeepc (approved):
🔥 non blocking suggestions

olha23:
@xoscar i'm just thinking if we have only one input here maybe make sense to unite variable name with previous step? 
i mean like this: 
Screenshot 2023-01-20 at 1 40 22 PM
it will help to avoid one more unnecessary click. What do you think?

PR #1863, opened 2023-01-20
feat: missing variable detection on test/transaction run

### LINKED ISSUE #1871: [Chaining V2] Undefined variables for test and transactions CLI

mathnogueira:
This PR changes how missing variables are detected. Now they are detected at the moment a test/transaction is run. The validation happens before the test is run and it will return a `422 Unprocessable Entity` status code in case of missing variables.

## Changes
- Variable support (`var:abc`) was dropped from the assertion engine (this was never used and we replaced it with env vars (`env:abc`)
- Removed endpoints for validating if there were missing variables for a test and a transaction
- Worked on an approach to use Go tags to choose which fields can contain expressions/statements and only scan those fields when searching for missing variables.

## Fixes
- fixes #1871 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/model/tests.go
```diff
@@ -23,12 +23,12 @@ type (
 
 	Output struct {
 		Selector SpanQuery
-		Value    string
+		Value    string `expr_enabled:"true"`
```
mathnogueira, line 26: Added those tags so we can choose which fields can contain expressions/statements. The idea is to use the same tags to inject resolved values into the test object as well. So we don't have to worry when adding new fields to the test/transaction struct.

server/expression/reflection.go
```diff
@@ -0,0 +1,90 @@
+package expression
+
+import (
+	"fmt"
+)
+
+type ReflectionToken struct {
```
xoscar, line 7: I'm having issues connecting the `reflection` name to what is happening here, what would you say is the main goal of this module?

server/expression/linting/missing_variables.go
```diff
@@ -0,0 +1,118 @@
+package linting
+
+import (
+	"reflect"
```
xoscar, line 4: What does this reflect module does? is it the one that reads the tags you added?

server/expression/linting/missing_variables.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func traverse(target interface{}, f func(reflect.StructField, string)) {
+	typ := reflect.TypeOf(target)
+	value := reflect.ValueOf(target)
+
+	traverseValue(value, typ, reflect.StructField{}, f)
+}
+
+func traverseValue(value reflect.Value, typ reflect.Type, parentField reflect.StructField, f func(reflect.StructField, string)) {
```
xoscar, line 69: Ohhh shit I think I get it haha

server/expression/linting/missing_variables.go
```diff
@@ -0,0 +1,118 @@
...
+			existingMissingVariables[variable] = true
+		}
+	}
+
+	return uniqueMissingVariables
+}
+
+func getTokens(fieldInfo reflect.StructField, value string) []expression.ReflectionToken {
+	if fieldInfo.Tag.Get("expr_enabled") == "true" {
+		exprTokens, err := expression.GetTokensFromExpression(value)
```
xoscar, line 40: So we go through all of the fields from a struct and depending on the type we decide if we recursively traverse over it or we just land into callback which triggers the getTokens function at some point.

Is this a standard way of traversing objects in golang?

server/http/validation/variable.go
```diff
@@ -0,0 +1,143 @@
+package validation
```
xoscar, line 1: What's the idea behind having this module be part of a `validation` subfolder? is this something we can continue to replicate for future scenarios?

server/model/http.go
```diff
@@ -27,14 +27,15 @@ var (
...
 type HTTPHeader struct {
-	Key, Value string
+	Key   string `expr_enabled:"true"`
+	Value string `expr_enabled:"true"`
 }
 
 type HTTPRequest struct {
-	Method  HTTPMethod
-	URL     string
+	Method  HTTPMethod `expr_enabled:"true"`
```
xoscar, line 35: I like this extra metadata approach haha

server/expression/linting/missing_variables.go
```diff
@@ -0,0 +1,118 @@
...
+		if err != nil {
+			// probably not an expression, just skip it
+			return []expression.ReflectionToken{}
+		}
+
+		return exprTokens
+	}
+
+	if fieldInfo.Tag.Get("stmt_enabled") == "true" {
```
xoscar, line 49: What does stmt mean?

xoscar (commented):
Hey dude great job, I tried to follow the new code as much as I could, I left some non-blocking questions for you!

I'm going to be adding the FE changes to this PR on Monday 😄

danielbdias (commented):
The overall PR seems fine. I have only one question: if we merge it as it is now, can we break our front end?

server/expression/reflection.go
mathnogueira: Makes sense. My idea was: I'm inspecting the result of an expression and usually this is called reflection. But just calling them tokens might make sense.

server/expression/linting/missing_variables.go
mathnogueira: It's a module that allows you to get information about a type. You can see its fields, its types, iterate though them, etc.

server/http/validation/variable.go
mathnogueira: Yes. One thing that the backend has to do better is to validate the input. I added this module there and I want to keep adding some validation objects there. `http` already has quite some objects there and I didn't want to make it bigger.

server/expression/linting/missing_variables.go
mathnogueira: Statement.

A statement is `<expr> <operator> <expr>`. Basically our assertion format.

server/expression/linting/missing_variables.go
```diff
@@ -0,0 +1,118 @@
...
+			return
+		}
+
+		realValue := value.Elem().Interface()
+		typ := reflect.TypeOf(realValue)
+		value := reflect.ValueOf(realValue)
+		traverseValue(value, typ, parentField, f)
+	case reflect.Struct:
+		// ignore time fields
+		if !typ.AssignableTo(reflect.TypeOf(time.Now())) {
```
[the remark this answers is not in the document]
mathnogueira: I'll add a comment there. And you found a bug. I shouldn't skip a time field, I should call `f` on it.

mathnogueira:
> can we break our front end?

Yes. It will break our frontend. I suggested @xoscar added the changes for the FE in this PR and we merge it after everything is working.

xoscar:
> > can we break our front end?
> 
> Yes. It will break our frontend. I suggested @xoscar added the changes for the FE in this PR and we merge it after everything is working.

Yep I'll be working on this today

server/expression/reflection.go
mathnogueira: Renamed it to `Token` instead.

danielbdias (approved):

xoscar:
@jorgeepc do you mind taking a look at the FE changes? thanks man!

web/src/providers/Transaction/hooks/useTransactionCrud.ts
```diff
@@ -5,48 +5,53 @@ import {TDraftTransaction, TTransaction} from 'types/Transaction.types';
...
-          name: transaction.name,
-          testsVariables: transactionVariables.variables,
-          onSubmit(variables) {
-            run(variables);
-          },
-        });
+      run();
     },
-    [getTransactionVariables, navigate, onOpen, runTransactionAction, selectedEnvironment?.id]
+    [navigate, onOpen, runTransactionAction, selectedEnvironment?.id]
```
jorgeepc, line 54: We have a missing dep here

jorgeepc (approved):
Clean 💥

PR #1884, opened 2023-01-24
feat: ask for missing variables in the CLI

mathnogueira:
This PR makes the CLI ask for missing variables

Screenshot from 2023-01-24 17-44-16

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/http/controller.go
+++ b/server/http/controller.go
@@ -569,7 +569,7 @@ func (c *controller) executeTest(ctx context.Context, test model.Test, runInfo o
 	testID := id.ID(resp.Body.(openapi.UpsertDefinitionResponse).Id)
 	// test ready, execute it
 	resp, err = c.RunTest(ctx, testID.String(), runInfo)
-	if resp.Code != http.StatusOK {
+	if resp.Code != http.StatusOK || err != nil {
 		return resp, err
 	}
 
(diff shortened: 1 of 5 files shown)


xoscar (approved):
LGTM! 🔥

jorgeepc (approved):

server/http/controller.go
```diff
@@ -569,7 +569,7 @@ func (c *controller) executeTest(ctx context.Context, test model.Test, runInfo o
 	testID := id.ID(resp.Body.(openapi.UpsertDefinitionResponse).Id)
 	// test ready, execute it
 	resp, err = c.RunTest(ctx, testID.String(), runInfo)
-	if err != nil {
+	if resp.Code != http.StatusOK {
```
danielbdias, line 572: Being paranoid: can we handle both status and error here? Like this:

```suggestion
	if resp.Code != http.StatusOK || err != nil {
```

server/http/controller.go
```diff
@@ -1006,7 +1006,7 @@ func (c *controller) executeTransaction(ctx context.Context, transaction model.T
 
 	// transaction ready, execute it
 	resp, err = c.RunTransaction(ctx, transactionID.String(), runInfo)
-	if err != nil {
+	if resp.Code != http.StatusOK {
```
danielbdias, line 1009: Same here:

```suggestion
	if resp.Code != http.StatusOK || err != nil {
```

danielbdias (approved):

Issue #1854, opened 2023-01-19
fix `--output json` option for `test run` cmd


mathnogueira:
**Describe the bug**
When providing `--output json` to the `test run` command, the output doesn't contain anything relevant for the test run result. It just shows the run and test objects. We need to make it print more relevant info.

PR #1891, opened 2023-01-25
fix: only print test and transaction result in json format

mathnogueira:
This PR changes the CLI to only display result information when the option `--output json` is defined. This makes it simpler for users to parse the result of the tests.
## Fixes

- #1854

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/actions/run_test_action.go
+++ b/cli/actions/run_test_action.go
@@ -259,7 +259,7 @@ func (a runTestAction) runDefinitionFile(ctx context.Context, f file.File, param
 		}).
 		Execute()
 
-	if resp.StatusCode == http.StatusUnprocessableEntity {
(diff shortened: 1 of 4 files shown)


cli/actions/run_test_action.go
```diff
@@ -259,7 +259,7 @@ func (a runTestAction) runDefinitionFile(ctx context.Context, f file.File, param
 		}).
 		Execute()
 
-	if resp.StatusCode == http.StatusUnprocessableEntity {
+	if resp != nil && resp.StatusCode == http.StatusUnprocessableEntity {
```
mathnogueira, line 262: prevents possible panics

danielbdias (approved):

PR #1907, opened 2023-01-27
chore(frontend): model types update

xoscar:
This PR updates the models to export the default type. In this case we gain some improvements like:
- Having types and factories defined in the same place
- Using the model types as props and parameters
- When looking for a definition the editor sends you directly to the model and not to the separate types file
- Decouples models from generated types

## Changes

- Moving model types out of the types folder into the models
- Updating every usage of `TModel` instances to use the `Model` directly

## Fixes

- Tech debt

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


web/src/models/DataStore.model.ts
```diff
@@ -1,4 +1,15 @@
-import {SupportedDataStores, TDataStore, TRawDataStore} from 'types/Config.types';
+import {SupportedDataStores} from 'types/Config.types';
+import {Model, TDataStoreSchemas} from 'types/Common.types';
+
+export type RawDataStore = TDataStoreSchemas['DataStore'];
+type DataStore = Model<
```
xoscar, line 5: Define the type as the same name as the default exported function but do not exported

web/src/providers/DataStore/DataStore.provider.tsx
```diff
@@ -10,7 +10,8 @@ import {
   useDeleteDataStoreMutation,
 } from 'redux/apis/TraceTest.api';
 import DataStoreService from 'services/DataStore.service';
-import {SupportedDataStores, TConnectionResult, TDataStore, TDraftDataStore} from 'types/Config.types';
+import {SupportedDataStores, TConnectionResult, TDraftDataStore} from 'types/Config.types';
+import DataStore from 'models/DataStore.model';
```
xoscar, line 14: Can be used latter on as type and as the factory function 🤩

jorgeepc (approved):

PR #1908, opened 2023-01-30
fix(backend): fixing the get test run query join condition

xoscar:
This PR fixes a bug we have in meta where all runs are matching at least one transaction because a missing condition as part of the join query.

## Changes

- Fixing the query at the testdb level

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/testdb/runs.go
+++ b/server/testdb/runs.go
@@ -337,7 +337,7 @@ SELECT
 FROM test_runs
 LEFT OUTER JOIN
 	transaction_run_steps transaction_run
-ON transaction_run.test_run_id = id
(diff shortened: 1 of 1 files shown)


danielbdias (approved):

jorgeepc (approved):

mathnogueira (approved):

mathnogueira:
This is a fix for a problem in the integration instance. I had to merge it with a failing pipeline cause it would be impossible for that to work without the changes in this PR

PR #1913, opened 2023-01-30
feat: allow cli to accept an `address` override param

schoren:
This PR allows users to override the target endpoint of cli requests through a flag. Example:

```
tracetest test list --address http://some-host.com:1234
# short version
tracetest test list -a http://some-host.com:1234
````

**UPDATE** renamed flag from `--endpoint|-e` to `--server-url|-s` to avoid conflict with `--environment|-e` flag

## Changes

-

## Fixes

- #1893 

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -19,7 +19,6 @@ var (
 
 	// overrides
 	overrideEndpoint string
-	overrideScheme   string
 )
 
 var rootCmd = &cobra.Command{
@@ -43,5 +42,4 @@ func init() {
 	rootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "display debug information")
 
 	rootCmd.PersistentFlags().StringVarP(&overrideEndpoint, "endpoint", "e", "", "endpoint")
-	rootCmd.PersistentFlags().StringVarP(&overrideScheme, "scheme", "s", "", "scheme")
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: cli/cmd/root.go [hunk 1]
```diff
@@ -37,4 +40,6 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&output, "output", "o", string(formatters.DefaultOutput), fmt.Sprintf("output format [%s]", outputFormatsString))
 	rootCmd.PersistentFlags().StringVarP(&configFile, "config", "c", "config.yml", "config file will be used by the CLI")
 	rootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "display debug information")
+
+	rootCmd.PersistentFlags().StringVarP(&overrideEndpoint, "endpoint", "e", "", "endpoint")
```
danielbdias, line 44: The alias `e` is used for the environment option, we will need to change that.
This is the error that had happened in our pipeline: `panic: unable to redefine 'e' shorthand in "run" flagset: it's already used for "environment" flag`

danielbdias (commented):
The overall code is fine! But there is one detail that we need to change to avoid colliding with the `environment` options.

cli/cmd/root.go
mathnogueira: Nice catch. Do we need a short name for this option? We could use just `--endpoint`. Or maybe use something else as `--server-url (-s)`

schoren: good catch. renamed the flag to overcome conflict

mathnogueira (approved):

danielbdias (approved):

PR #1949, opened 2023-02-06
fix: check if env id exists every time

mathnogueira:
This PR makes the CreateEnvironment check if the id exists every time.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


danielbdias (approved):

mathnogueira:
This will not pass because the integration instance will not be updated and tests are failing due to a problem in the integration instance. So I'm merging it anyway.

PR #1952, opened 2023-02-07
feat: show help page when user tries to run placeholder commands

mathnogueira:
If user tries to run the following commands, it will get a help page for the command:

* tracetest
* tracetest datastore
* tracetest environment
* tracetest server
* tracetest test

based on @danielbdias's idea.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/cmd/datastore_cmd.go
+++ b/cli/cmd/datastore_cmd.go
@@ -1,9 +1,6 @@
 package cmd
 
 import (
-	"fmt"
-
-	"github.com/kubeshop/tracetest/cli/analytics"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


jorgeepc (approved):

cli/cmd/datastore_cmd.go
```diff
@@ -13,9 +10,7 @@ var dataStoreCmd = &cobra.Command{
 	Long:   "Manage your tracetest data stores",
 	PreRun: setupCommand(),
 	Run: func(cmd *cobra.Command, args []string) {
-		analytics.Track("Datastore", "cmd", map[string]string{})
```
mathnogueira, line 16: Removed those analytics events because there were useless for us. Besides, we are going to remove analytics from the CLI in a near future.

danielbdias (approved):

Issue #1938, opened 2023-02-02
[DataStores] Research Connectivity issues with Grafana Cloud Tempo


xoscar:
While running a test to validate tempo connection
We found out that our settings for tempo don't support grafana cloud tempo yet

We need to investigate and fix the integration.

xoscar:
Seems that the Tempo cloud version doesn't support the GRPC client we use to connect from Tracetest, the options now are:
1. Wait for a potential update in the future to support GRPC
2. Add a different connection mechanism for Tempo Cloud (only) to use HTTP instead


CC: @schoren @mathnogueira @kdhamric @jorgeepc

kdhamric:
I think we should do this in the next few weeks. Adding it to our v0.10 big boulder board.

xoscar:
The team agreed to add an option to the Tempo connector to support both GRPC and HTTP

PR #1963, opened 2023-02-09
feature: adding grafana tempo cloud support

xoscar:
This PR adds support for the HTTP client for Grafana Tempo Cloud
It decouples the tempo config from just supporting gRPC and adds HTTP capabilities.

## Changes

- Updates the Tempo config to support the HTTP client
- Adds section for gRPC and HTTP
- Implements generic HTTP and gRPC client for the BE
- Fixes issue with the test connection not supporting URLs with paths

## Fixes

- #1938 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

https://www.loom.com/share/f17a9708c64341ab93e853cd5940c760


api/dataStores.yaml
```diff
@@ -29,6 +29,15 @@ components:
         createdAt:
           type: string
           format: date-time
+    BaseClient:
```
xoscar, line 32: Base client supports both http and grpc settings

examples/tracetest-k6/docker-compose.yml
```diff
@@ -149,3 +149,12 @@ services:
         condition: service_healthy
       queue:
         condition: service_healthy
+
+  tempo:
```
xoscar, line 153: Adding example including tempo

server/http/mappings/data_store.go
```diff
@@ -54,9 +58,13 @@ func (m Model) DataStore(in openapi.DataStore) model.DataStore {
 	deepcopy.DeepCopy(in.Jaeger, &dataStore.Values.Jaeger)
 	deepcopy.DeepCopy(in.Jaeger.Tls, &dataStore.Values.Jaeger.TLSSetting)
 	deepcopy.DeepCopy(in.Jaeger.Tls.Settings, &dataStore.Values.Jaeger.TLSSetting.TLSSetting)
+
 	deepcopy.DeepCopy(in.Tempo, &dataStore.Values.Tempo)
-	deepcopy.DeepCopy(in.Tempo.Tls, &dataStore.Values.Tempo.TLSSetting)
-	deepcopy.DeepCopy(in.Tempo.Tls.Settings, &dataStore.Values.Tempo.TLSSetting.TLSSetting)
+	deepcopy.DeepCopy(in.Tempo.Grpc.Tls, &dataStore.Values.Tempo.Grpc.TLSSetting)
+	deepcopy.DeepCopy(in.Tempo.Grpc.Tls.Settings, &dataStore.Values.Tempo.Grpc.TLSSetting.TLSSetting)
```
xoscar, line 64: We need to fix this at some point as its getting out of hand

server/tracedb/client/client.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/kubeshop/tracetest/server/config"
+	"github.com/kubeshop/tracetest/server/model"
+	"github.com/kubeshop/tracetest/server/tracedb/connection"
+	"google.golang.org/grpc"
+)
+
+type SupportedClients string
+
+var (
+	GRPC SupportedClients = "grpc"
```
xoscar, line 15: Main facade to use the http and grpc clients

server/tracedb/client/grpc.go
```diff
@@ -0,0 +1,80 @@
...
+		return errors.Wrap(connection.ErrInvalidConfiguration, err.Error())
+	}
+
+	conn, err := grpc.DialContext(ctx, client.config.Endpoint, opts...)
+	if err != nil {
+		return errors.Wrap(connection.ErrConnectionFailed, err.Error())
+	}
+
+	client.conn = conn
+	return nil
```
xoscar, line 39: Generic grpc client

server/tracedb/client/http.go
```diff
@@ -0,0 +1,123 @@
+package client
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io"
```
xoscar, line 10: Generic http client

server/tracedb/connection/connection.go
```diff
@@ -0,0 +1,64 @@
...
+	return r.OperationDescription != ""
+}
+
+func IsReachable(endpoint string) (bool, error) {
+	address, err := url.Parse(endpoint)
+	if err != nil {
+		return false, err
+	}
+
+	endpoint = strings.TrimPrefix(endpoint, "http://")
```
xoscar, line 47: Fixes an issue with urls like `https://tempo-us-central1.grafana.net/tempo`

server/tracedb/tempodb.go
```diff
@@ -158,10 +122,52 @@ func (ttd *tempoTraceDB) GetTraceByID(ctx context.Context, traceID string) (mode
 	return traces.FromOtel(trace), nil
 }
 
-func (ttd *tempoTraceDB) Close() error {
-	err := ttd.conn.Close()
+type HttpTempoTraceByIDResponse struct {
+	Batches []*traces.HttpResourceSpans `json:"batches"`
+}
+
+func httpGetTraceByID(ctx context.Context, traceID string, client client.HttpClient) (model.Trace, error) {
```
xoscar, line 129: Adding HTTP trace processor

web/src/components/Settings/DataStorePlugin/forms/BaseClient/BaseClient.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const BaseClient = () => {
+  const form = Form.useFormInstance<TDraftDataStore>();
+  const dataStoreType = form.getFieldValue('dataStoreType');
+  const baseName = ['dataStore', dataStoreType];
+  const type = (Form.useWatch([...baseName, 'type'], form) || SupportedClientTypes.GRPC) as SupportedClientTypes;
+  const Component = FieldsFormMap[type];
+
+  return (
+    <>
```
xoscar, line 19: Implements switching from HTTP to GRPC

web/src/components/Settings/DataStorePlugin/forms/HttpClient/HttpClient.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const HEADER_DEFAULT_VALUES = [{key: '', value: ''}];
+
+const HttpClient = () => {
+  const form = Form.useFormInstance<TDraftDataStore>();
+  const dataStoreType = form.getFieldValue('dataStoreType');
+  const baseName = ['dataStore', dataStoreType, 'http'];
+  const insecureValue = Form.useWatch([...baseName, 'tls', 'insecure'], form) ?? true;
+
+  return (
```
xoscar, line 14: new HTTP client

server/tracedb/client/client.go
```diff
@@ -0,0 +1,99 @@
...
+
+type HttpCallback func(ctx context.Context, traceID string, client HttpClient) (model.Trace, error)
+type GrpcCallback func(ctx context.Context, traceID string, connection *grpc.ClientConn) (model.Trace, error)
+
+type Callbacks struct {
+	GRPC GrpcCallback
+	HTTP HttpCallback
+}
+
+type Client struct {
```
danielbdias, line 27: This naming is difficult to understand because we are referencing as `client.Client`, which feels like "this is a client for what"? The idea behind it is here: https://github.com/golang/go/wiki/CodeReviewComments#package-names

At first glance, it seems to me something like `TraceDataSource`. Invoking other golang devs here: @schoren @mathnogueira , it makes sense folks?

server/traces/otel_http_converter.go
```diff
@@ -0,0 +1,112 @@
...
+		return value["stringValue"].(string)
+	}
+
+	if value["intValue"] != nil {
+		return fmt.Sprintf("%s", value["intValue"])
+	}
+
+	if value["doubleValue"] != nil {
+		val := value["doubleValue"].(float64)
+		if val != 0.0 {
```
danielbdias, line 97: I'm not sure if this comparison will work correctly every time, because of the floating point craziness that we have (like `0.1 * 10 != 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1 + 0.1` that we have in some languages like Python and JS).

To check if it is a double here we could do:
```go
val, ok := value["doubleValue"].(float64)
if ok {
  // do stuff of `if val != 0.0`
}
```

server/tracedb/client/client.go
xoscar: I like `TraceDataSource` or just `DataSource` I'll give it a go, thanks for pointing me to that documentation file ❤️

xoscar:
Hey @danielbdias thanks for the review my man, I just updated some of the things you mentioned!

server/tracedb/client/http.go
mathnogueira: Is it worth adding a generic http client? Every api is different and I can't see a benefit in having a generic client for it. I think we would have to make it extremely configurable, and I don't think that's worth the effort.

This might be applicable to the gRPC client too.

@schoren @danielbdias I want your thoughts here

danielbdias: The thing about having it extremely configurable is that the complexity of the generic client will explode in a little time (for instance, having a lot of `ifs` and `options`), that's correct?

To tame that, I was thinking if we can follow something like the [strategy design pattern](https://sourcemaking.com/design_patterns/strategy): we could have one client for each DataStore and the data store itself decides how to use communication strategy internally (HTTP and/or gRPC).

xoscar: @danielbdias the initial idea was to build something similar to the strategy pattern, but as this is a double layer (trace db + clients) I think a better way to describe it is as an [Abstract Factory](https://refactoring.guru/design-patterns/abstract-factory).

The situation here is that tracedbs are all similar until the `getTraceById` step, where each one has a different way of handling that. I can create a `data_store` interface so each of the clients has the same implementation, and a single callback type

xoscar: I updated the code based on the comments, let me know what you think

danielbdias (approved):

jorgeepc (approved):
Great job 🔥

PR #1977, opened 2023-02-10
feat(server): implement viper

schoren:
This PR implements Viper in the server, allowing configure server configurations from config file, flags or env.
There are some cleanups, like embedding migrations and simplifying integration testing.

As a bonus, I implemented Cobra commands, so we can have a `tracetest-server help` command.

## Checklist

- [x] tested locally
- [x] added new dependencies
- [x] updated the docs
- [x] added a test


server/config/pooling.go
```diff
@@ -1,41 +1,38 @@
...
 }
 
-func (c *Config) SetPoolingConfig(pc PoolingConfig) {
+func (c *Config) PoolingMaxWaitTimeForTraceDuration() time.Duration {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	c.config.PoolingConfig = pc
+	maxWaitTimeForTrace := c.vp.GetDuration("poolingConfig.maxWaitTimeForTrace")
+	if maxWaitTimeForTrace == 0 {
```
danielbdias, line 21: Can we print a Warning log here saying that this value is malformed and we are assuming a default value? It could be good to avoid implicit behaviors for our users.

server/config/server.go
```diff
@@ -1,47 +1,81 @@
...
+var serverOptions = options{
+	{"postgres.host", "postgres", "postgres DB host"},
+	{"postgres.user", "postgres", "postgres DB user"},
+	{"postgres.password", "postgres", "postgres DB password"},
+	{"postgres.dbname", "tracetest", "postgres DB dbname"},
+	{"postgres.port", 5432, "postgres DB port"},
+	{"postgres.params", "sslmode=disable", "postgres DB connection params"},
+
+	{"server.httpPort", 11633, "tracetest HTTP Port"},
+	{"server.pathPrefix", "", "tracetest HTTP Path prefix"},
```
danielbdias, line 16:
```suggestion
	{"postgres.host", "postgres", "Postgres DB host"},
	{"postgres.user", "postgres", "Postgres DB user"},
	{"postgres.password", "postgres", "Postgres DB password"},
	{"postgres.dbname", "tracetest", "Postgres DB database name"},
	{"postgres.port", 5432, "Postgres DB port"},
	{"postgres.params", "sslmode=disable", "Postgres DB connection parameters"},

	{"server.httpPort", 11633, "Tracetest server HTTP Port"},
	{"server.pathPrefix", "", "Tracetest server HTTP Path prefix"},
```

server/migrations/migrations.go
```diff
@@ -0,0 +1,6 @@
+package migrations
+
+import "embed"
+
+//go:embed *.sql
```
danielbdias, line 5: Makes sense to force here just to embed the files in the migrations folder? Is there a change of embedding files in the `examples` folder by mistake?

server/testdb/postgres.go
```diff
@@ -49,7 +48,12 @@ func (p *postgresDB) ensureLatestMigration() error {
 	if err != nil {
 		return fmt.Errorf("could not get driver from postgres connection: %w", err)
 	}
-	migrateClient, err := migrate.NewWithDatabaseInstance(p.migrationsFolder, "tracetest", driver)
+	sourceDriver, err := iofs.New(migrations.Migrations, ".")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	migrateClient, err := migrate.NewWithInstance("iofs", sourceDriver, "tracetest", driver)
```
danielbdias, line 56: 👏 👏 👏 👏

danielbdias (approved):
It is a great PR! I liked how we are dealing now with migrations, too. It became easier to deal with them.

I've added some comments about documentation and small things to improve user alerting.

mathnogueira (commented):
Just one comment on my part. LGTM after @danielbdias's comments are addressed

server/cmd/serve.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/kubeshop/tracetest/server/config"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(serveCmd)
+}
+
+var serveCmd = &cobra.Command{
+	Use:   "serve",
```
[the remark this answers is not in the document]
schoren: no particular reason. I'm open to change it if people like other verb better

server/migrations/migrations.go
schoren: this is embed during build time. this helps reduce the configurations needed to boot the app. We used to have an internal config for migrations path that was only used for testing, so this change makes the migrations "hardcoded" withinh the app so no configuration is needed

schoren:
@danielbdias @mathnogueira I added config validation on startup, so we don;'t need to check errors when accesing the values

Screenshot 2023-02-14 at 13 24 19

Issue #179, opened 2022-04-12
Support Amazon x-ray


povilasv:
1. Make an app that sends data to x-ray
2. Enable 1/10 client sampling
3. Test if tracetest  forces client to send the trace
4. Implement get trace by id for Amazon x-ray
5. Convert amazon trace model to otel.

Get trace by id API -> https://docs.aws.amazon.com/xray/latest/devguide/xray-api-gettingdata.html

kdhamric:
@mathnogueira Are you going to be working off this issue once you start on X-Ray? I want to tag @mik3h0 on which ever issue you do the work on. His company uses Amazon x-ray and he may be able to assist in trying it once we have it integrated.

PR #1987, opened 2023-02-13
feature: adding support for AWS X-RAY

xoscar:
This PR enables Amazon X-Ray as a data store for Tracetest.

## Changes

- Adds a Node.js example to use Tracetest with X-Ray
- Includes BE and FE support for the new Data Store

## Fixes

- #179 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Using AWS XRAY SDK to instrument plus the ADOT COllector
https://www.loom.com/share/5441578aef2442c78684916834335428

## Using AWX XRAY SDK to instrument plus AWS XRAY SDK to pull Traces
https://www.loom.com/share/4ca15c89ba48455b813d6c605c828063

## Using OpenTelemetry (Pokeshop) to instrument plus XRAY ADOT collector exporter plus AWS XRAY SDK to pull Traces
https://www.loom.com/share/3a4f3ad4955a4390ae475c18cad30f99


examples/tracetest-amazon-x-ray/tracetest-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+demo:
+  enabled: []
+
+experimentalFeatures: []
+
+telemetry:
+  dataStores:
+    otlp:
+      type: otlp
```
jorgeepc, line 18: should the type be `awsxray`?

server/tracedb/tracedb.go
```diff
@@ -67,6 +67,7 @@ func (f *traceDBFactory) New(ds model.DataStore) (tdb TraceDB, err error) {
 		tdb, err = newOpenSearchDB(ds.Values.OpenSearch)
 	case openapi.SIGNAL_FX:
 		tdb, err = newSignalFXDB(ds.Values.SignalFx)
+	case openapi.AWSXRAY:
```
mathnogueira, line 70: Shouldn't we get the trace from X-Ray's API instead of relying on the collector here?

https://docs.aws.amazon.com/sdk-for-go/api/service/xray/#XRay.BatchGetTraces

xoscar: It's better to use the ADOT collector as it's becoming the norm for AWS and OpenTelemetry, plus, as per one of the dev teams we talked with, they will be using X-RAY (step functions) + Honeycomb. The ADOT/OTEL collectors allow us to combine multiple receivers/exporters, instead of natively supporting various data stores now.

Also, the simplest way to integrate x-ray today is through the ADOT collector, if not you'll have to set up the x-ray [daemon instance](https://docs.aws.amazon.com/xray/latest/devguide/xray-daemon.html).

I'm open to ideas tho if we want to natively use the x-ray API directly, we would need to ask users to add credentials or roles to the Tracetest instance/container so it can access it.

examples/tracetest-amazon-x-ray/tracetest-config.yaml
xoscar: Fixed 😄

server/tracedb/tracedb.go
xoscar: Ok, so after talking to @kdhamric we are going to separate this into two things, the first one will be a recipe for Tracetest - X-Ray - ADOT collector 

And the actual settings page and BE will be the configuration to connect to the native X-Ray API

server/id/generator.go
```diff
@@ -41,9 +42,15 @@ func (g randGenerator) ID() ID {
 }
 
 func (g randGenerator) TraceID() trace.TraceID {
-	tid := trace.TraceID{}
-	g.rand.Read(tid[:])
-	return tid
+	var r [16]byte
```
xoscar, line 45: @mathnogueira @schoren there is quite an annoying rule by x-ray that requires traces to include a prefix to calculate the creation time, if the trace is older than 30 days the API returns an error, if the prefix is missing x-ray returns an error, I have this code here knowing that is wrong to put it in this place, should we also support a different way of creating traces based on data store? I'm open to ideas

server/tracedb/awsxray.go
```diff
@@ -0,0 +1,801 @@
+package tracedb
+
+import (
```
xoscar, line 3: Big chunky file that contains everything related to the x-ray connectivity and parsing of the data

server/tracedb/tracedb.go
xoscar: Done 🥴

server/id/generator.go
danielbdias: Brainstorming: Maybe we should have specializations of this generator (or about the traceid logic) and do something similar with the connectivity PR?

server/tracedb/awsxray.go
```diff
@@ -0,0 +1,767 @@
...
+	for _, s := range seg.Subsegments {
+		nestedSpans, err := segToSpans(s, traceID, &span)
+
+		if err != nil {
+			return spans, err
+		}
+
+		spans = append(spans, nestedSpans...)
+
+		// if seg.Cause != nil &&
```
danielbdias, line 168: Can we remove this code?

server/tracedb/awsxray.go
```diff
@@ -0,0 +1,767 @@
...
+}
+
+const (
+	traceIDLength    = 35 // fixed length of aws trace id
+	identifierOffset = 11 // offset of identifier within traceID
+)
+
+func convertToAmazonTraceID(traceID trace.TraceID) (string, error) {
+	const (
+		// maxAge of 28 days.  AWS has a 30 day limit, let's be conservative rather than
```
danielbdias, line 396:
```suggestion
		// maxAge of 28 days in seconds.  AWS has a 30-day limit, let's be conservative rather than
```

server/tracedb/awsxray.go
danielbdias: Makes sense to create a package for XRay inside of `tracedb` and break this file by responsibilities/concerns?

api/dataStores.yaml
```diff
@@ -62,6 +64,15 @@ components:
           type: string
         token:
           type: string
+    AwsXRay:
+      type: object
+      properties:
+        region:
+          type: string
+        accessKeyId:
```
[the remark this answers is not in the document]
xoscar: Will be addressed in a separate PR

PR #1994, opened 2023-02-14
refactor connectivity test to reuse components

mathnogueira:
This PR changes the internals of how we test the connection of each data store. Instead of having a long function with repetitive code, it uses a new `connection.Tester` which allow test steps to be chained and reused.

This is a required change to make it easier to expand our validations in the future.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/tracedb/connection/connection.go
+++ b/server/tracedb/connection/connection.go
@@ -45,11 +45,11 @@ func (r ConnectionTestStepResult) IsSet() bool {
 	return r.OperationDescription != ""
 }
 
-func IsReachable(endpoint string, protocol Protocol) (bool, error) {
+func IsReachable(endpoint string, protocol Protocol) error {
 	if protocol == ProtocolHTTP {
 		address, err := url.Parse(endpoint)
 		if err != nil {
-			return false, err
+			return err
 		}
 
 		endpoint = strings.TrimPrefix(endpoint, "http://")
@@ -66,8 +66,8 @@ func IsReachable(endpoint string, protocol Protocol) (bool, error) {
 
 	_, err := net.DialTimeout("tcp", endpoint, 5*time.Second)
 	if err != nil {
-		return false, err
+		return err
 	}
 
-	return true, nil
+	return nil
 }
--- a/server/tracedb/connection/connection.go
+++ b/server/tracedb/connection/connection.go
@@ -39,23 +39,25 @@ func (r ConnectionTestStepResult) IsSet() bool {
 }
 
 func IsReachable(endpoint string) (bool, error) {
-	address, err := url.Parse(endpoint)
-	if err != nil {
-		return false, err
+	if strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://") {
+		address, err := url.Parse(endpoint)
+		if err != nil {
+			return false, err
+		}
+
+		endpoint = strings.TrimPrefix(endpoint, "http://")
+		endpoint = strings.TrimPrefix(endpoint, "https://")
+
+		if address.Scheme == "https" && address.Port() == "" {
+			endpoint = fmt.Sprintf("%s:443", address.Hostname())
+		}
+
+		if address.Scheme == "http" && address.Port() == "" {
+			endpoint = fmt.Sprintf("%s:80", address.Hostname())
+		}
 	}
 
-	endpoint = strings.TrimPrefix(endpoint, "http://")
-	endpoint = strings.TrimPrefix(endpoint, "https://")
-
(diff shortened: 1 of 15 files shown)


server/tracedb/connection/connection.go
```diff
@@ -39,23 +39,25 @@ func (r ConnectionTestStepResult) IsSet() bool {
 }
 
 func IsReachable(endpoint string) (bool, error) {
-	address, err := url.Parse(endpoint)
-	if err != nil {
-		return false, err
-	}
+	if strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://") {
```
mathnogueira, line 42: By checking if it's HTTP or not. It prevents problems when an IP is used as an endpoint.

server/tracedb/connection/tester.go
```diff
@@ -0,0 +1,48 @@
...
+			ConnectivityTestResult:   connectivityTestResult,
+			AuthenticationTestResult: authTestResult,
+		}
+	}
+
+	return ConnectionTestResult{
+		ConnectivityTestResult:   t.connectivityTestStep.TestConnection(ctx),
+		AuthenticationTestResult: t.authenticationTestStep.TestConnection(ctx),
+		TraceRetrievalTestResult: t.pollingTestStep.TestConnection(ctx),
+	}
```
schoren, line 47: This code is running the tests twice, once to check the error and again for building the result. It can do both things at the same time:

```suggestion
func (t *Tester) TestConnection(ctx context.Context) res ConnectionTestResult {
	res.ConnectivityTestResult = t.connectivityTestStep.TestConnection(ctx)
	if res.ConnectivityTestResult.Error != nil {
		return 
	}

	res.AuthenticationTestResult = t.authenticationTestStep.TestConnection(ctx)
	if res.ConnectivityTestResult.Error != nil {
		return 
	}

	res.TraceRetrievalTestResult = t.pollingTestStep.TestConnection(ctx)

	return
}
```

server/tracedb/connection/connection.go
```diff
@@ -39,23 +39,25 @@ func (r ConnectionTestStepResult) IsSet() bool {
...
 
-	if address.Scheme == "http" && address.Port() == "" {
-		endpoint = fmt.Sprintf("%s:80", address.Hostname())
+		if address.Scheme == "http" && address.Port() == "" {
+			endpoint = fmt.Sprintf("%s:80", address.Hostname())
+		}
 	}
 
-	_, err = net.DialTimeout("tcp", endpoint, 5*time.Second)
+	_, err := net.DialTimeout("tcp", endpoint, 5*time.Second)
```
danielbdias, line 60: Can we move the `5*time.Second` timeout to a constant?

schoren (changes requested):
This looks great! I left a few suggestions, but there's one change I think needs to be done, regarding the duplicated connections testing.

server/tracedb/connection/options.go
```diff
@@ -0,0 +1,37 @@
+package connection
+
+import "context"
+
+func WithConnectivityTest(step TestStep) TesterOption {
```
xoscar, line 5: This pattern is very interesting, do you have documentation around it? I'd like to understand it better 😄

mathnogueira: It's called _option pattern_. You can read about it here: https://golang.cafe/blog/golang-functional-options-pattern.html

server/tracedb/connection/connection.go
```diff
@@ -38,11 +47,11 @@ func (r ConnectionTestStepResult) IsSet() bool {
 	return r.OperationDescription != ""
 }
 
-func IsReachable(endpoint string) (bool, error) {
-	if strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://") {
+func IsReachable(endpoint string, protocol Protocol) error {
```
danielbdias, line 50: Since we are not returning a `bool` anymore, it feels weird to call `IsReachable` to check an error. Makes sense to change this function to `CheckReachability` or something similar?

server/tracedb/connection/connection.go
```diff
@@ -9,6 +9,15 @@ import (
 	"time"
 )
 
+const timeoutIsReachable = 5 * time.Second
```
danielbdias, line 12:
```suggestion
const reachabilityTimeout = 5 * time.Second
```

danielbdias (approved):

schoren (approved):
LGTM!!!

PR #1999, opened 2023-02-15
config(server): allow provisioning from file or environent variable

schoren:
This PR adds the ability to provision tracetest from a config file passed on server start, or from an environment variable containing a base64 encoded provisioning file.

Provisioning file examples:
- elastic APM: https://github.com/kubeshop/tracetest/pull/1999/files#diff-f4f72665eddc70d1132cb4da2fc85c1c0d50a3453c1537fb8401bff7f13aec52
- jaeger https://github.com/kubeshop/tracetest/pull/1999/files#diff-085fdc1ffc07871140a8c272754a6f23947d986cf5ca57c00257f7e781717205
- opensearch https://github.com/kubeshop/tracetest/pull/1999/files#diff-079a37a807ffca8ab08663744a832ab30a61ce2076d4f55d4036e5b7ef91d4b0
- more examples available at the `server/provisioning/testdata/` dir in this PR.

Env var example: https://github.com/kubeshop/tracetest/pull/1999/files#diff-1cccd986a18cb0e5acf7d1d7e38133b13b0164318e7e152188ad2147f47dbb2aR28

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


danielbdias (approved):

examples/tracetest-provisioning-env/docker-compose.yml
```diff
@@ -0,0 +1,64 @@
+version: '3'
+services:
+
+    tracetest:
+        image: kubeshop/tracetest:${TAG:-latest}
+        platform: linux/amd64
+        volumes:
+            - type: bind
+              source: ./tracetest-config.yaml
+              target: /app/tracetest.yaml
```
xoscar, line 10: Can we remove this so everything comes from the env variables?

server/app/app.go
```diff
@@ -48,31 +50,68 @@ func New(config *config.Config) (*App, error) {
...
-	fmt.Println(a.Version())
+type appOption func(app *App)
+
+func WithProvisioningFile(path string) appOption {
+	return func(app *App) {
+		app.provisioningFile = path
+	}
+}
+
+func (app *App) provision(db model.Repository) {
```
xoscar, line 80: I'm a big fan of having more explicit naming like this!! 🔥

server/testdb/data_stores.go
```diff
@@ -225,36 +223,12 @@ func (td *postgresDB) readDataStoreRow(ctx context.Context, row scanner) (model.
...
 		}
 
-		return patchDataStoreValues(dataStore), nil
+		return dataStore, nil
 	default:
 		return model.DataStore{}, err
 	}
 }
 
-func patchDataStoreValues(dataStore model.DataStore) model.DataStore {
```
xoscar, line 234: I guess this is now covered by the `nil` checks from the models right?

xoscar (changes requested):
Amazing job @schoren I left a couple of comments, the important one to me is having the example without using the `tracetest.yaml` config at all

server/testdb/data_stores.go
schoren: Forgot to delete this, thanks

schoren: Yes, Otherwise it's just the same as using direct values instead of references

examples/tracetest-provisioning-env/docker-compose.yml
schoren: Great idea

schoren: Now that I checked it better, we still have some values that work through the "old" config file, like telemetry settings. this will be changed to a different approach later, so I didn't migrate them.

xoscar (approved):

PR #2001, opened 2023-02-15
feat: allow durations to have spaces in it in expressions

mathnogueira:
This PR makes it possible to declare duration values using spaces separating the number and the time unit.

## Changes

- `100 ms` is valid and it's converted into `100ms`

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/expression/parser_rules.go
```diff
@@ -107,25 +107,3 @@ func (term *Term) Type() TermType {
 
 	return ""
 }
-
-func (expr *Expr) GetTermsByType(termType TermType) []*Term {
```
mathnogueira, line 111: Unused function

danielbdias (approved):

jorgeepc (approved):

PR #2011, opened 2023-02-17
fix(server): removing MinSpanCount span logic from trace poller and adding unit tests

danielbdias:
This PR aims to update our Trace collection polling logic to remove an assumption about "Tracetest trigger" node and add unit tests to handle more cases on our poller.

I'll tackle the problem of using the "Tracetest trigger" node in another PR.

## Changes

- Updating polling logic to make the stop condition explicit
- Remove logic that considers "Tracetest trigger" span on stopping criterion

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [x] added a test

### DIFF
--- a/server/executor/poller_executor.go
+++ b/server/executor/poller_executor.go
@@ -159,7 +159,12 @@ func (pe DefaultPollerExecutor) donePollingTraces(job *PollingRequest, traceDB t
 		return false
 	}
 
-	if len(trace.Flat) > traceDB.MinSpanCount() && len(trace.Flat) == len(job.run.Trace.Flat) {
+	haveNotCollectedSpansSinceLastPoll := len(trace.Flat) == len(job.run.Trace.Flat)
+
+	// Today we consider that we finished collecting traces
+	// if we haven't collected any new spans since our last poll
+
+	if haveNotCollectedSpansSinceLastPoll {
 		log.Printf("[PollerExecutor] Test %s Run %d: Done polling. Condition met\n", job.test.ID, job.run.ID)
 		return true
 	}
--- a/server/executor/poller_executor.go
+++ b/server/executor/poller_executor.go
@@ -160,11 +160,13 @@ func (pe DefaultPollerExecutor) donePollingTraces(job *PollingRequest, traceDB t
 	}
 
 	haveNotCollectedSpansSinceLastPoll := len(trace.Flat) == len(job.run.Trace.Flat)
+	haveCollectedSpansInTestRun := len(trace.Flat) > 0
 
 	// Today we consider that we finished collecting traces
 	// if we haven't collected any new spans since our last poll
+	// and we have collected at least one span for this test run
 
-	if haveNotCollectedSpansSinceLastPoll {
+	if haveNotCollectedSpansSinceLastPoll && haveCollectedSpansInTestRun {
 		log.Printf("[PollerExecutor] Test %s Run %d: Done polling. Condition met\n", job.test.ID, job.run.ID)
 		return true
 	}
--- a/server/executor/poller_executor_test.go
+++ b/server/executor/poller_executor_test.go
@@ -2,6 +2,7 @@ package executor_test
 
 import (
 	"context"
+	"log"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


mathnogueira (approved):

server/executor/poller_executor.go
```diff
@@ -159,7 +159,14 @@ func (pe DefaultPollerExecutor) donePollingTraces(job *PollingRequest, traceDB t
...
 
-	if len(trace.Flat) > traceDB.MinSpanCount() && len(trace.Flat) == len(job.run.Trace.Flat) {
+	hasCollectedTriggeredSpans := len(trace.Flat) > traceDB.MinSpanCount()
+	haveNotCollectedSpansSinceLastPoll := len(trace.Flat) == len(job.run.Trace.Flat)
+
+	// Today we consider that we finished collecting traces
+	// when we collected at least one more span beyond "Tracetest trigger"
+	// and if we haven't collected any new spans since our last poll
+
+	if hasCollectedTriggeredSpans && haveNotCollectedSpansSinceLastPoll {
```
mathnogueira, line 169:
```suggestion
	hasCollectedTriggerSpans := len(trace.Flat) > traceDB.MinSpanCount()
	haveNotCollectedSpansSinceLastPoll := len(trace.Flat) == len(job.run.Trace.Flat)

	// Today we consider that we finished collecting traces
	// when we collected at least one more span beyond "Tracetest trigger"
	// and if we haven't collected any new spans since our last poll

	if hasCollectedTriggerSpans && haveNotCollectedSpansSinceLastPoll {
```

I think it's easier to understand "trigger spans" rather than "triggered spans" which might reference the spans that were triggered by the test (i.e., all trace spans)

jorgeepc (approved):

server/executor/poller_executor_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/kubeshop/tracetest/server/tracing"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_PollerExecutor_ExecuteRequest(t *testing.T) {
+	t.Parallel()
+
+	// Cases: tests with 0 traces
+	// Cases: tests with only 1 trace
+	// Cases: tests with more than 1 trace
```
jorgeepc, line 22: with 0 `spans`? right?

danielbdias: Yes! I'm still finishing this test, hopefully, I'll enable it for review at the end of the day

server/executor/trace_poller.go
```diff
@@ -75,6 +75,15 @@ type PollingRequest struct {
 	count int
 }
 
+func NewPollingRequest(ctx context.Context, test model.Test, run model.Run, count int) *PollingRequest {
```
mathnogueira, line 78: Can't you use this on the production code as well? Now It's only used for tests at the moment.

server/executor/poller_executor.go
```diff
@@ -159,7 +159,12 @@ func (pe DefaultPollerExecutor) donePollingTraces(job *PollingRequest, traceDB t
 		return false
 	}
 
-	if len(trace.Flat) > traceDB.MinSpanCount() && len(trace.Flat) == len(job.run.Trace.Flat) {
+	haveNotCollectedSpansSinceLastPoll := len(trace.Flat) == len(job.run.Trace.Flat)
+
+	// Today we consider that we finished collecting traces
+	// if we haven't collected any new spans since our last poll
+
+	if haveNotCollectedSpansSinceLastPoll {
```
mathnogueira, line 167: It's important to have the `len(trace.Flat) > 0` because we might not get any trace data for a long time. Usually, it takes more to receive the initial spans than to get the subsequent spans. That condition was important so we didn't stop polling before getting any span.

server/executor/poller_executor_test.go
```diff
@@ -0,0 +1,380 @@
...
+func Test_PollerExecutor_ExecuteRequest_NoRootSpan_NoSpanCase(t *testing.T) {
+	t.Parallel()
+
+	// Scenario: Trace without spans
+	// Given the trigger execution returns 0 spans
+	// And tracetest does not send the root span
+	// When the server do the polling process
+	// Then it should stop at the second iteration
+	// And it should have no error on the process
+	// And a root span should be added to it
```
mathnogueira, line 34: I think this should never happen. I think we should timeout the polling process instead of stopping it when we don't get any trace data.

mathnogueira (changes requested):
A couple of things. There's only one scenario that I don't think we should add

server/executor/trace_poller.go
danielbdias: Makes sense, adding it!

mathnogueira (approved):

Issue #1898, opened 2023-01-26
[Support] Work on a ECS Fargate Deployment for Tracetest


xoscar:
We got contacted by a team that is interested in using Tracetest for their Serverless setup, to allow for easier integration, we are going to be working on an ECS Fargate example deployment for Tracetest.

Acceptance criteria:
AC1:
As a user wanting to understand how to install tracetest to an AWS ECS Fargate setup
I should be able to look at the example part of the repo

xoscar:
To have a better way to support this use case we have to consider working in the following issues first:

1. https://github.com/kubeshop/tracetest/issues/1912
2. https://github.com/kubeshop/tracetest/issues/1911

PR #2028, opened 2023-02-22
feature(examples): adding AWS terraform example

xoscar:
This PR adds a terraform AWS example for tracetest making use of the new config changes

## Changes

- Introduces a new example for tracetest using terraform and AWS

## Fixes

- #1898 

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/e69ebd60863e429186775747959e861b


server/config/config.go
```diff
@@ -80,7 +80,8 @@ func configureConfigFile(vp *viper.Viper) {
 var ErrConfigFileNotFound = errors.New("config file not found")
 
 func readConfigFile(vp *viper.Viper) error {
-	if confFile := vp.GetString("config"); confFile != "" {
+	confFile := vp.GetString("config")
```
xoscar, line 83: There is a really weird bug that sometimes when you don't specify a `tracetest.yaml` config file the server can't start because of a weird issue trying to parse an empty file or something like that.

I added this but I think the initial bug needs to be fixed either way @schoren @mathnogueira

jorgeepc (approved):
🔥

PR #2039, opened 2023-02-23
Resources manager

schoren:
This PR implements a `resourcemanager` and provides a simple but useful *Config* resource (package `config/configresource`).

## Overview

The main idea behind the `resourcemanager` is to have a very tested and standardized way of doing the boring task of managing resources (a.k.a CRUD). We have, and will be adding more, resources that are just entities persisted in the DB that needs to be handled. They also have a yaml and json representation that needs to be handled.

The similarities in the management of these resources tends to create very copy/paste-y code, such as controllers, service layers, etc. The `resourcemanager` encapsulates these similarities.

A key feature is the normalization it provides. All resources are handled in a standard way, and follow a standard pattern. This applies to the "interface" (the API), meaning that all resources handled the same *from the perspective of the client*.
This clarification is important because while the API interface is standardized, the internal handling of each resource is left to each implementation.

## Testing 

In the same spirit of normalization, we have a `resourcemanager/testutil` package that provides a set of standard tests that a resource needs to pass. This has 2 benefits:
1. A developer doesn't need to spend a lot of time creating essentialy copy/pasted tests, and
2. It ensures compatibility and contract implementation

The API of these normalized tests consists of a single test entrypoint function, `rmtests.TestResourceType`. Internally it runs multiple operationTests (get success, list success, list with no results, etc), once for each supported content type (which at the moment is only json and yaml). The final test matrix is defined by:
1. resource Type
2. operation
3. content type

so the total tests run on a resource are `(number of operations) x (content types)`. In this incomplete implementation we have [10 operations](https://github.com/kubeshop/tracetest/pull/2039/files#diff-13aa5bec3a570833ae213d3d03b20b984d2fbb4b914569a2422f50389d735a51R13)  and [2 content types](https://github.com/kubeshop/tracetest/pull/2039/files#diff-4113c542a74918b6256e4f88046a3a8e35994cc244a0747a7c0f7193c8d53961R39), totaling in 20 tests per resource.

If we had to create 20 tests per resource, we would have, at the moment, 100 manually copy/pasted tests: 20 for each of the 5 existing compatible resources: tests, transactions, data stores, environments, configs.

On top of that, adding new tests expands the coverage equally fast. Adding 1 test operation results on 2 new tests applied to each resources (using the `rmtests.TestResourceType` utility). Adding a new content type (if we wanted to) would exponentially create 10 new tests, with the currently defined operations.

The testing is done through HTTP requests and `http/httptest.Server` so we ensure http compatibility. This way makes the tests for the resources that uses an actual DB connection for tests _almost_ e2e.


server/resourcemanager/encoding.go
```diff
@@ -0,0 +1,58 @@
...
+func (e jsonEncoder) Accepts(contentType string) bool {
+	return contentType == "application/json"
+}
+func (e jsonEncoder) Marshal(in interface{}) (out []byte, err error) {
+	return json.Marshal(in)
+}
+func (e jsonEncoder) Unmarshal(in []byte, out interface{}) (err error) {
+	return json.Unmarshal(in, out)
+}
+
```
mathnogueira, line 46: Why gofmt didn't format this file to add empty lines separating the functions?

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	targetResource := Resource[T]{}
+	err = mapstructure.Decode(values, &targetResource)
+	if err != nil {
+		writeResponse(w, http.StatusBadRequest, fmt.Sprintf("cannot unmarshal body values: %s", err.Error()))
+		return
+	}
+
+	// TODO: if resoruceType != values.resourceType return error
```
mathnogueira, line 67:
```suggestion
	// TODO: if resourceType != values.resourceType return error
```

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,132 @@
...
+		return
+	}
+
+	writeResponse(w, http.StatusCreated, string(bytes))
+
+}
+
+func writeResponse(w http.ResponseWriter, code int, msg string) {
+	w.WriteHeader(code)
+	w.Write([]byte(msg))
```
mathnogueira, line 92: You probably need to write the Content-Type header here as well

server/resourcemanager/testutil/test_resource.go
```diff
@@ -0,0 +1,114 @@
...
+			return string(j)
+		},
+	},
+}
+
+func TestResourceType(t *testing.T, rt ResourceTypeTest) {
+	t.Helper()
+
+	t.Run(rt.ResourceType, func(t *testing.T) {
+		rt := rt
```
mathnogueira, line 68: I'm almost sure you don't need this line because `rt` doesn`t change. We only need to do this thing when we are working with parallel tests inside a loop. Just like in did on line 73.

mathnogueira (commented):
Good job. Just a couple of small things

server/resourcemanager/encoding.go
schoren: not sure, but I like it :P

server/resourcemanager/testutil/test_resource.go
schoren: yeap, I forgot to `t.Parallel()` here

schoren: You're right. I'll clean this up and check if i didn't repeat anything

server/resourcemanager/resource_manager.go
schoren: not strictly neccesary, but love the consistency this change introduces. great catch!

Dockerfile
```diff
@@ -6,7 +6,6 @@ COPY ./tracetest-server /app/tracetest-server
 COPY ./tracetest /app/tracetest
 
 COPY ./web/build ./html
-COPY ./server/migrations/ /app/migrations/
```
schoren, line 9: this is a leftover from a previous change

server/config/configresource/config_resource_test.go
```diff
@@ -0,0 +1,63 @@
...
+	db := testmock.MustGetRawTestingDatabase()
+	sampleConfig := configresource.Config{
+		ID:               "1",
+		Name:             "test",
+		AnalyticsEnabled: true,
+	}
+
+	rmtests.TestResourceType(t, rmtests.ResourceTypeTest{
+		ResourceType: "Config",
+		RegisterManagerFn: func(router *mux.Router) any {
```
schoren, line 26: This function is run before each operation, so serves as a kind of `setup` function

server/config/configresource/config_resource_test.go
```diff
@@ -0,0 +1,63 @@
...
+			db := testmock.MustCreateRandomMigratedDatabase(db)
```
schoren, line 27: We create a new random, clean DB for each operation/content type combination

server/config/configresource/config_resource_test.go
```diff
@@ -0,0 +1,63 @@
...
+			configRepo := configresource.Repository(db)
+
+			manager := resourcemanager.New[configresource.Config]("Config", configRepo, id.GenerateID)
+			manager.RegisterRoutes(router)
+
+			return configRepo
+		},
+		Prepare: func(t *testing.T, op rmtests.Operation, bridge any) {
```
schoren, line 35: This func allows to configure the repository before each operation is run

server/config/configresource/config_resource_test.go
```diff
@@ -0,0 +1,63 @@
...
+			configRepo := bridge.(resourcemanager.ResourceHandler[configresource.Config])
+			switch op {
+			case rmtests.OperationGetSuccess,
+				rmtests.OperationUpdateSuccess,
+				rmtests.OperationDeleteSuccess,
+				rmtests.OperationListSuccess:
+				configRepo.Create(context.TODO(), sampleConfig)
```
schoren, line 42: In this example, we create a new example config for the operations that require a preexisting config in the DB. This approach allows to not depend on the order of the operations

server/resourcemanager/encoding.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+type encoder interface {
+	Marshal(in interface{}) (out []byte, err error)
+	Unmarshal(in []byte, out interface{}) (err error)
+	Accepts(contentType string) bool
+	ResponseContentType() string
+}
+
+var encoders = []encoder{
```
schoren, line 19: Registering encoders helps reduce coupling in the code by passing around specific instances instead of having `switch contentType`s everywhere

server/resourcemanager/encoding.go
```diff
@@ -0,0 +1,68 @@
...
+func (e basicEncoder) Marshal(in interface{}) (out []byte, err error) {
+	return e.marshalFn(in)
+}
+func (e basicEncoder) Unmarshal(in []byte, out interface{}) (err error) {
+	return e.unmarshalFn(in, out)
+}
+
+var errUnacceptableContentType = errors.New("unacceptable content type")
+
+func encoderFromRequest(r *http.Request) (encoder, error) {
```
schoren, line 59: this func is used to get the correct encoder instance for a given request based on the content type

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,347 @@
...
+	subrouter.HandleFunc("/", m.create).Methods(http.MethodPost).Name(fmt.Sprintf("%s.Create", m.resourceType))
+	subrouter.HandleFunc("/", m.update).Methods(http.MethodPut).Name(fmt.Sprintf("%s.Update", m.resourceType))
+
+	subrouter.HandleFunc("/{id}", m.get).Methods(http.MethodGet).Name(fmt.Sprintf("%s.Get", m.resourceType))
+	subrouter.HandleFunc("/{id}", m.delete).Methods(http.MethodDelete).Name(fmt.Sprintf("%s.Delete", m.resourceType))
+
+	return subrouter
+}
+
+func (m *manager[T]) create(w http.ResponseWriter, r *http.Request) {
```
schoren, line 71: These functions are the real essence of this PR. This code is very much improvable, but provides the basic functionality

server/resourcemanager/resource_manager_test.go
```diff
@@ -0,0 +1,197 @@
...
+		SomeValue: "the value",
+	}
+
+	sampleUpdated := sampleResource{
+		ID:        "1",
+		Name:      "the name updated",
+		SomeValue: "the value updated",
+	}
+
+	rmtests.TestResourceTypeWithErrorOperations(t, rmtests.ResourceTypeTest{
```
schoren, line 83: This test serves to test the actual resource manager. It relies on a mock instead of an actual DB so we can have better control of what happens at each step. 

For example, we can easily simulate DB errors to assert they are handled correctly and nicely presented to users. 

It uses the `TestResourceTypeWithErrorOperations` function (instead of `TestResourceType`) because it includes operations targeted at checking these kinds of internal error handling.

These generic/shared functionalities are tested here, leaving each resource implementation with the responsibility to test its internal workings by thenselves.

server/resourcemanager/testutil/test_resource.go
```diff
@@ -0,0 +1,135 @@
...
+			j, err := yaml.YAMLToJSON([]byte(yamlString))
+			if err != nil {
+				panic(err)
+			}
+			return string(j)
+		},
+	},
+}
+
+func TestResourceType(t *testing.T, rt ResourceTypeTest) {
```
schoren, line 67: This is the entry point test function for general resources

server/resourcemanager/testutil/test_resource.go
```diff
@@ -0,0 +1,135 @@
...
+	TestResourceTypeOperations(t, rt, defaultOperations)
+}
+
+func TestResourceTypeWithErrorOperations(t *testing.T, rt ResourceTypeTest) {
+	t.Helper()
+
+	TestResourceTypeOperations(t, rt, append(defaultOperations, errorOperations...))
+}
+
+func TestResourceTypeOperations(t *testing.T, rt ResourceTypeTest, operations []operationTester) {
```
schoren, line 79: Generic function for applying any operations. Could (should?) be unexported

server/resourcemanager/testutil/operations.go
```diff
@@ -0,0 +1,115 @@
...
+	"encoding/json"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	defaultOperations = []operationTester{
```
schoren, line 13: these are the operations intended for the "general public" resources implementation

server/testmock/database.go
```diff
@@ -67,8 +93,9 @@ func getMainDatabaseConnection(container *gnomock.Container) (*sql.DB, error) {
 }
 
 func createRandomDatabaseForTest(db *sql.DB, baseDatabase string) (*sql.DB, error) {
-	epoch := time.Now().UnixNano()
```
schoren, line 70: this approach for "randomization" of DB names was causing collisions when running multiple operations in parallel, so I changed it to more random approach

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,347 @@
...
+	Count int              `mapstructure:"count"`
+	Items []map[string]any `mapstructure:"items"`
+}
+
+type Resource[T ResourceSpec] struct {
+	Type string `mapstructure:"type"`
+	Spec T      `mapstructure:"spec"`
+}
+
+type ResourceHandler[T ResourceSpec] interface {
```
danielbdias, line 33: ❤️

server/go.mod
```diff
@@ -53,6 +53,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
 	gotest.tools/v3 v3.1.0
+	sigs.k8s.io/yaml v1.2.0
```
mathnogueira, line 56: You added this dependency, but you are actually using `gopkg.in/yaml.v2` instead. Can this be removed?

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,347 @@
...
+
+func (m *manager[T]) list(w http.ResponseWriter, r *http.Request) {
+	encoder, err := encoderFromRequest(r)
+	if err != nil {
+		writeResponse(w, http.StatusBadRequest, fmt.Sprintf("cannot process request: %s", err.Error()))
+		return
+	}
+	w.Header().Set("Content-Type", encoder.ResponseContentType())
+
+	ctx := r.Context()
```
mathnogueira, line 122: It's better to use the context from the request, so if there's any instrumentation, it gets propagated.

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,347 @@
...
+		resourceList.Items = append(resourceList.Items, values)
+	}
+
+	bytes, err := encodeValues(resourceList, encoder)
+	if err != nil {
+		writeError(w, encoder, http.StatusInternalServerError, fmt.Errorf("cannot marshal entity: %w", err))
+		return
+	}
+
+	writeResponse(w, 200, string(bytes))
```
mathnogueira, line 177:
```suggestion
	writeResponse(w, http.StatusOK, string(bytes))
```

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,347 @@
...
+		Spec: item,
+	}
+
+	bytes, err := encodeValues(newResource, encoder)
+	if err != nil {
+		writeError(w, encoder, http.StatusInternalServerError, fmt.Errorf("cannot marshal entity: %w", err))
+		return
+	}
+
+	writeResponse(w, 200, string(bytes))
```
mathnogueira, line 208:
```suggestion
	writeResponse(w, http.StatusOK, string(bytes))
```

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,347 @@
...
+	vars := mux.Vars(r)
+	id := id.ID(vars["id"])
+
+	err = m.handler.Delete(r.Context(), id)
+	if err != nil {
+		m.handleResourceHandlerError(w, "deleting", err, encoder)
+		return
+	}
+
+	w.WriteHeader(204)
```
mathnogueira, line 228:
```suggestion
	w.WriteHeader(http.StatusNoContent)
```

server/resourcemanager/resource_manager.go
```diff
@@ -0,0 +1,347 @@
...
+		writeError(w, encoder, http.StatusBadRequest, fmt.Errorf("cannot unmarshal body values: %w", err))
+		return
+	}
+
+	// TODO: if resourceType != values.resourceType return error
+
+	// TODO: check if this needs to be done per operation
+	if !targetResource.Spec.HasID() {
+		targetResource.Spec = m.handler.SetID(targetResource.Spec, m.idgen())
+	}
```
mathnogueira, line 269: question: This enables two different configurations to be saved into the server. In case that happens, which one would be picked and applied by the server?

mathnogueira (commented):
Couple of small things

server/go.mod
schoren: No, this is used in tests: https://github.com/kubeshop/tracetest/pull/2039/files#diff-4113c542a74918b6256e4f88046a3a8e35994cc244a0747a7c0f7193c8d53961R51

server/resourcemanager/resource_manager.go
schoren: `r` is the request actually

mathnogueira: I have no idea why, but I read it as `ctx := ctx.Background()`. I must be tired :grimacing:

server/resourcemanager/resource_manager.go
schoren: I didn't get to that part yet. These configs are also *not* being consumed by anyone, so it's just to insert/manage the entity. In a later PR, this can be enforced by last created, or always being an update, or other approaches.

mathnogueira (approved):
Looks good to me

danielbdias (approved):

server/config/configresource/config_resource.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	_, err = tx.ExecContext(ctx, updateQuery,
+		cfg.ID,
+		updated.Name,
+		updated.AnalyticsEnabled,
+	)
+
+	if err != nil {
+		tx.Rollback()
```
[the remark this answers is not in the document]
schoren: how would that look exactly? sounds interesting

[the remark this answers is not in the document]
danielbdias: sure! the main idea is this example: https://go.dev/doc/database/execute-transactions

we can do something like this on this function:

```go
func (r *repository) Update(ctx context.Context, updated Config) (Config, error) {
	cfg, err := r.Get(ctx, updated.ID)
	if err != nil {
		return Config{}, err
	}

	tx, err := r.db.BeginTx(ctx, nil)
	if err != nil {
		return Config{}, err
	}
  defer tx.Rollback() // defer Rollback, executing it at the end of the function if Commit wasn't called

	_, err = tx.ExecContext(ctx, updateQuery,
		cfg.ID,
		updated.Name,
		updated.AnalyticsEnabled,
	)

	if err != nil {
		return Config{}, fmt.Errorf("sql exec: %w", err)
	}

	err = tx.Commit()
	if err != nil {
		return Config{}, fmt.Errorf("commit: %w", err)
	}

	return updated, nil
}
```

jorgeepc (commented):
Great work. One question, we are not going to use the OpenAPI definition for these resources, right?

mathnogueira:
Can we merge this?

Issue #2019, opened 2023-02-21
Environments page improvement


olha23:
[Environments page improvement#1986](https://github.com/kubeshop/tracetest/issues/1986)

Issue #1986, opened 2023-02-13
Environments page improvement


olha23:
Link for design - https://www.figma.com/file/LBN4SKVPq3ykegrPKbHT2Y/0.8-0.9-Release-Tracetest?node-id=1686%3A28255&t=NTCvZlJORKNhyutP-1

xoscar:
@olha23 I think this is done right? as it has the link fo the designs?

olha23:
@xoscar yes

Issue #1478, opened 2022-11-14
Implement Notification system


xoscar:

olha23:
@xoscar can you add description here? i will help with this form design side

olha23:
mockups here https://www.figma.com/file/LBN4SKVPq3ykegrPKbHT2Y/0.8-0.9-Release-Tracetest?node-id=1884%3A29145&t=kppsclr4eq0O35dO-1

PR #2041, opened 2023-02-23
feature(frontend): environment ux improvements

xoscar:
This PR implements the Environments UX/UI changes described here [Environments page improvement#1986](https://github.com/kubeshop/tracetest/issues/1986)

## Changes

- Moves the environment modal to its own component folder
- Extends the environment provider to include the CRUD
- Decouples the environment modal from the API calls

## Fixes

- #2019 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/c6e25f981e1b4cabbce97ac5738bcc8e


web/src/providers/Environment/hooks/useEnvironmentCrud.ts
```diff
@@ -0,0 +1,46 @@
...
+        onConfirm: () => deleteEnvironment({environmentId: id}),
+      });
+    },
+    [deleteEnvironment, onOpen]
+  );
+
+  const edit = useCallback(
+    async (environmentId: string, environment: Environment) => {
+      await updateEnvironment({environmentId, environment});
+      message.success('Environment updated successfully');
```
jorgeepc, line 30: We should try to use an standard component for those confirmation messages. I think we have a different one for DataStores. Not a blocker we can handle that in a different task.

jorgeepc (approved):
Clean 🧼  Great job dude!

web/src/providers/Environment/hooks/useEnvironmentCrud.ts
olha23: agree with @jorgeepc comment

xoscar: I believe we added the messages to provide feedback for some of the FE actions. 

I created this issue a long time ago while thinking about adding a standardized notification system https://github.com/kubeshop/tracetest/issues/1478

Let's all think about that and how we can introduce it in more general way 😄

Issue #1996, opened 2023-02-15
[config] support `postgresConnString` for backwards compatibility


adnanrahic:
Make it possible to configure Tracetest's Postgres connection with:

- connection string
- parameters

mathnogueira:
Based on our conversation, it would be:

```yaml
postgresConnString: "my connection string"
```

PR #2043, opened 2023-02-23
fix: postgres config retrocompatibility 

mathnogueira:
This PR makes the new config retro compatible with the old configuration file format. This means that `postgresConnString` will be used if present in the config file.

## Fixes

- #1996 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


schoren (approved):
LGTM!

xoscar (approved):

PR #2055, opened 2023-02-28
feat: validate port and show warning if using a non-standard port

mathnogueira:
This PR adds a step to the data store test which validates the used port. If it's not a standard port, show a warning to the user and point to the correct port.

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

Loom: https://www.loom.com/share/7105d6d5268c411a84665d896eb3cf4e

### DIFF
--- a/api/config.yaml
+++ b/api/config.yaml
@@ -13,6 +13,8 @@ components:
     ConnectionResult:
       type: object
       properties:
+        endpointLinting:
+          $ref: "#/components/schemas/ConnectionTestStep"
         connectivity:
           $ref: "#/components/schemas/ConnectionTestStep"
         authentication:
@@ -24,6 +26,12 @@ components:
       properties:
         passed:
           type: boolean
+        status:
+          type: string
+          enum:
+            - passed
+            - passed_with_warning
+            - failed
         message:
           type: string
         error:
--- a/api/config.yaml
+++ b/api/config.yaml
@@ -13,7 +13,7 @@ components:
     ConnectionResult:
       type: object
       properties:
-        endpointLinting:
+        portCheck:
           $ref: "#/components/schemas/ConnectionTestStep"
         connectivity:
           $ref: "#/components/schemas/ConnectionTestStep"
@@ -30,7 +30,7 @@ components:
           type: string
           enum:
             - passed
-            - passed_with_warning
+            - warning
             - failed
         message:
           type: string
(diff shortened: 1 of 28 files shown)


api/config.yaml
```diff
@@ -13,6 +13,8 @@ components:
     ConnectionResult:
       type: object
       properties:
+        endpointLinting:
```
mathnogueira, line 16: I'm not sure about this name. I'm accepting suggestions

web/src/components/TestConnectionNotification/TestConnectionNotification.styled.ts
```diff
@@ -1,4 +1,4 @@
-import {CheckCircleFilled, MinusCircleFilled} from '@ant-design/icons';
+import {CheckCircleFilled, MinusCircleFilled, WarningFilled} from '@ant-design/icons';
```
mathnogueira, line 1: I used the warning icon, but it doesn't look great. Any suggestion of an icon that we can use? Especially because there's no `WarningCircleFilled` icon :disappointed:

web/src/components/TestConnectionNotification/TestConnectionNotification.styled.ts
```diff
@@ -12,6 +12,11 @@ export const FailedCheckIcon = styled(MinusCircleFilled)`
   margin-top: 3px;
 `;
 
+export const WarningCheckICon = styled(WarningFilled)`
```
jorgeepc, line 15:
```suggestion
export const WarningCheckIcon = styled(WarningFilled)`
```

api/config.yaml
jorgeepc: As we are only checking the port, maybe `portCheck` ? 🤔

jorgeepc (approved):
Great work @mathnogueira

api/config.yaml
```diff
@@ -13,6 +13,8 @@ components:
...
+          $ref: "#/components/schemas/ConnectionTestStep"
```
xoscar, line 17: I'm not a huge fan of having these be strict entries of the connection result object, I think this should be an array with different shapes of objects it can return, maybe having a type.

Not something for this PR tho 😄

xoscar (approved):
GJ my dudeee I left some comments let me know what you think

api/config.yaml
mathnogueira: used `portCheck` because it's way better than `endpointLinting`

danielbdias (approved):

web/src/components/TestConnectionNotification/TestConnectionNotification.styled.ts
olha23: @mathnogueira for keeping consistency we can use round warning icon: 
Screenshot 2023-02-28 at 1 41 38 PM
i will add here svg for it:
ExclamationCircle

Issue #1997, opened 2023-02-15
CLI showing wrong results for Datadog


danielbdias:
**Describe the bug**

When running Datadog example `./examples/tracetest-datadog` through the CLI we are receiving a failed message:
```sh
✘ OpenTelemetry Shop - List Products (http://localhost:11633/test/JBYAfKJ4R/run/5/test)
        ✔ span[tracetest.span.type="general" name="Tracetest trigger"]
                ✔ #385f2fc2cbb95311
                        ✔ attr:tracetest.response.status   =   200 (200)
                        ✔ attr:tracetest.span.duration  <  10ms (9ms)
        ✘ span[tracetest.span.type="rpc" name="grpc.hipstershop.ProductCatalogService/ListProducts"]
                ✘ #meta
                        ✘ attr:rpc.grpc.status_code = 0 (http://localhost:11633/test/JBYAfKJ4R/run/5/test?selectedAssertion=1)
        ✘ span[tracetest.span.type="rpc" name="hipstershop.ProductCatalogService/ListProducts" rpc.system="grpc" rpc.method="ListProducts" rpc.service="hipstershop.ProductCatalogService"]
                ✘ #meta
                        ✘ attr:rpc.grpc.status_code = 0 (http://localhost:11633/test/JBYAfKJ4R/run/5/test?selectedAssertion=2)
```

However, when accessing this test on the Web UI we see that all tests passed:
<img width="1762" alt="image" src="https://user-images.githubusercontent.com/12397879/219049385-f657187b-c69c-4805-b65c-69e0e2b87eb6.png">

**Expected behavior**
Both CLI and Web UI should show the same results.

danielbdias:
Team, this happened because our TracePoller stopped before receiving all trace spans. I've added another stopping condition on https://github.com/kubeshop/tracetest/pull/2059 to avoid this situation.

Issue #1904, opened 2023-01-27
[bug] distributed trace is not loaded correctly but shows up on hard page reload


adnanrahic:
**Describe the bug**
When I run a test it does not load any traces… but when I hard reload the page, the traces load and my assertions pass. However the prior empty state is saved in the test list. Very strange.

https://www.loom.com/share/bb1d23bbdb734274b04d6439739f96e0

**To Reproduce**
1. Go to a test run
2. Click on run test
3. Wait for page to refresh and display distributed trace
4. Distributed trace is not loaded
5. On reloading the page manually it shows up

**Expected behavior**
Distributed trace should load initially.

adnanrahic:
@xoscar FYI, I just ran this test again with a few different versions that work as expected without this bug:

- `v0.9.3`
- `v0.9.2`
- `v0.9.1`
- `v0.9.0`


This issue happens in versions:

- `v0.9.4`
- `v0.9.5` as in my initial test


Seems it started with `v0.9.4`.

xoscar:
@adnanrahic what setup are you using here? is this a different instrumented app than the pokeshop/otel demo? what data store and config are you using? 🤔

mathnogueira:
@adnanrahic what was the application you were using to reproduce it and what was the data store?

jorgeepc:
@mathnogueira in the video I see a `books service`, maybe Adnan is using this example https://github.com/kubeshop/tracetest/tree/main/examples/quick-start-nodejs-manual-instrumentation

adnanrahic:
@jorgeepc is right. That's the one I used.

/cc @mathnogueira @xoscar

adnanrahic:
@jorgeepc is right. That's the one I used.

/cc @mathnogueira @xoscar

adnanrahic:
@danielbdias - Will your PR https://github.com/kubeshop/tracetest/pull/2011 close this issue?

danielbdias:
Team, this happened because our TracePoller stopped before receiving all trace spans. I've added another stopping condition on https://github.com/kubeshop/tracetest/pull/2059 to avoid this situation.

Issue #1563, opened 2022-11-22
Single page app has to be refreshed to show finished test


kdhamric:
A number of time when testing in beta I have to refresh the page to see that the test has actually finished. Maybe 1 out of 4 times - not real difficult to reproduce. Here is an example of the screen before refreshing:

<img width="1124" alt="image" src="https://user-images.githubusercontent.com/1385894/203437001-129ec7c8-77f2-4daf-b9b4-e435d548d874.png">

xoscar:
@mathnogueira @danielbdias do you mind taking a look a the websocket configuration? we might be having some issues causing this bug

xoscar:
CCing @schoren too 😄

danielbdias:
Team, this happened because our TracePoller stopped before receiving all trace spans. I've added another stopping condition on #2059 to avoid this situation.

PR #2059, opened 2023-03-01
fix(server): update tracepoller stopping condition to consider root node

danielbdias:
This PR aims to fix the problem of our TracePoller stopping prematurely with only the root node. 

This is important because we don't have control on how OpenTelemetry is configured on client applications. If a client API emits telemetry way slower than us, a test can be stopped with only the "Tracetest trigger" node.

It depends on #2011 to be merged first.

## Changes

- Updated logs on Tracepoller
- Changed stop condition on Tracepoller
- Added test to validate a scenario where a root node arrives way earlier than the client spans

## Fixes

- #1563 
- #1904 
- #1997 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


jorgeepc (approved):
🔥 LGTM!

xoscar (approved):

PR #2064, opened 2023-03-01
feature(examples): adding aws step functions example

xoscar:
This PR provides a complete example of using Tracetest to test AWS Step Functions instrumentation.

It uses this repository as a base https://github.com/aws-samples/aws-step-functions-plagiarism-demo-dotnetcore

Which includes:
- Step functions example
- .NET Lambda functions
- Async processing w/ Sendgrid email notifications
- DynamoDB
- XRay auto instrumentation

## Changes

- Adds new recipe example

## Fixes

- #2045 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

https://www.loom.com/share/1deb1c05bab842069e0268499fe71501


server/config/config.go
```diff
@@ -100,7 +100,7 @@ func readConfigFile(vp *viper.Viper) error {
 	}
 
 	_, fileNotFound := err.(viper.ConfigFileNotFoundError)
-	if fileNotFound {
+	if fileNotFound || vp.GetString("config") == "" {
```
xoscar, line 103: There is a bug with the current implementation that if you don't provide a `config.yaml` file the server fails @schoren I'm going to create an issue in our board with the specs

server/executor/trigger/http.go
```diff
@@ -73,7 +73,7 @@ func (te *httpTriggerer) Trigger(ctx context.Context, test model.Test, opts *Tri
 	client := httpClient()
 
 	ctx = trace.ContextWithSpanContext(ctx, newSpanContext(ctx))
-	ctx, cncl := context.WithTimeout(ctx, 5*time.Second)
+	ctx, cncl := context.WithTimeout(ctx, 30*time.Second)
```
xoscar, line 76: Some services take more than 5 seconds, I increased this in the meantime but we should consider having this part of config or something like that

server/tracedb/awsxray.go
```diff
@@ -129,11 +129,16 @@ func parseXRayTrace(traceID string, rawTrace *xray.Trace) (model.Trace, error) {
 		return model.Trace{}, nil
 	}
 
-	segment := rawTrace.Segments[0]
-	spans, err := parseSegmentToSpans([]byte(*segment.Document), traceID)
+	spans := []model.Span{}
 
-	if err != nil {
-		return model.Trace{}, err
```
xoscar, line 136: This is fixing an issue with the x-ray integration caused by us just looking at the initial segment instead of the full list

danielbdias (approved):
Good job! The PR sounds fine!

I left one comment about a concern with the `Token` variable. Do you think it makes sense?

And one question thinking in more examples in the future: should we split PRs like this into three parts: one for documentation, one for the code itself, and a last one for the infra code (like terraform)?

examples/tracetest-aws-step-functions/src/SubmitExamTask.Tests/FunctionTests.cs
```diff
@@ -0,0 +1,200 @@
...
+namespace SubmitExamTask.Tests
+{
+    public class FunctionTests : IDisposable
+    {
+        private readonly ITestOutputHelper _testOutputHelper;
+        private readonly IAmazonDynamoDB _dynamoDbClient;
+        private const string TablePrefix = "IncidentsTestTable-";
+
+        private const string Token =
+            "AAAAKgAAAAIAAAAAAAAAAbdvA5UnsPbXk2HGkayUMygJK8eFJq3pnwBV/xTTDwiIbXvk246zL6Y1+UxXRWzPnbLD0mex2AEUEwMfjxjOj0lW0g+6AwFv6gA0MW/gU2SAdkHZl7tQQ1o3uBL2eOlSSYakcvPvF35BJdXCFkhhKaoqB8CzpnzkJPr7KVSXumjMouy/C4KwJJMqcVpeIW2Xhjyxq6FFT8+GRfNspJUaGE3aId15q/dK94xRTPG/Gidez7iuINk6Y7JpbA4/sj3T2hpUuDKyi4CcCkI8A4z93Hn2Tw2OMqWwhmserDGNfI3UgW3Um6pHRYNvL1prARZ9DkGHHftGaaXXBU8IO1mxYij4TciyP2Cky4b/Dk6ImioM0s+xdIeFOfMprMg73KG5WPK0XAWF+coMC7zBKJTtHZmudk9wKzTPdiSEZrwmPgeD3hVeWTQXwi7GF9hVbpS8wz/QrtI78HGPcbUdMi0Y79YihuGDo6iN4booO/5Tek3prcfDKhU3JtqqqVFRp9ugqQlOxhnkGmKaajp5miRFDcgrghxvP8Fp4D1DDY+/5vUxHFS+tOqvrp24YpSfO51xQxp7GWeg0k9qSnSWntOKdJRjmE7gyvIhKC9XMnlLktJEeBpCQa/B3pqzIr31sPB9ooDTS7m97REIl6Gf0VOtOx4=";
```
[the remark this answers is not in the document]
xoscar: Yeah, its the same as the one from the original repo https://github.com/aws-samples/aws-step-functions-plagiarism-demo-dotnetcore/blob/develop/src/backend/SubmitExamTask.Tests/FunctionTests.cs

xoscar:
> Good job! The PR sounds fine!
> 
> I left one comment about a concern with the `Token` variable. Do you think it makes sense?
> 
> And one question thinking in more examples in the future: should we split PRs like this into three parts: one for documentation, one for the code itself, and a last one for the infra code (like terraform)?

Dude thanks for your review, you are right about splitting the changes into multiple PRs!

server/config/config.go
xoscar: Bug ticket https://github.com/kubeshop/tracetest/issues/2071

server/executor/trigger/http.go
danielbdias: Created ticket to add this later to the trigger: https://github.com/kubeshop/tracetest/issues/2079

xoscar: Thanks!

PR #2078, opened 2023-03-01
Adding defer logic to config resource

danielbdias:
This PR aims to add a `defer` logic into every transaction that we open on `database/sql` package, following this: https://go.dev/doc/database/execute-transactions

## Changes

- Added `defer` logic to every `tx.Rollback()`

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
LGTM!

Issue #2071, opened 2023-03-01
Tracetest empty config file error


xoscar:
Hey team, I found a bug while working on the Fargate provisioning for Tracetest, what happens is that the server always expects a config file to be available, and it throws an error and kills the server if it's not found.

Here's a video about the bug

https://www.loom.com/share/340a43cdfac44bad903b2c0560a10070

![Image](https://user-images.githubusercontent.com/11051031/222236264-6064a6af-a169-49e0-9cec-305e6f8b4dc9.png)

schoren:
This was actually a very unexpected bug. we have the `tracetest` cli binary saved in the `/app/tracetest`. When tracetest-server starts, using `/app/` as the CWD, viper tries to find a config file in that directory. Apparently, setting the required config type by calling `vp.SetConfigType("yaml")` is not enough to enforce the file extension, so viper thought `/app/tracetest` was a valid config file path.

When `docker-compose.yaml` had the config file volume enabled, the `/app/tracetest.yaml` took precedence and the error didn't trigger.

This is fixed by https://github.com/kubeshop/tracetest/pull/2080. As an extra measure, I added a log to show the config file in use, if any. Might be useful for debugging similar errors in the future

xoscar:
@schoren WTH haha makes sense, thanks for the fix!

PR #2080, opened 2023-03-01
fix(server): force extension on config file lookup

schoren:
The server was accepting a config file with a name matching the expected config name but without the extension as a valid config file. This caused that in docker, where we have the cli binary in `/app/tracetest`, the server was trying to read said binary as a config file, obviously failing.

## Changes

- enforce config file with extension to avoid collisions.

## Fixes

- #2071

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -76,8 +76,7 @@ func (opts options) registerFlags(flags *pflag.FlagSet) {
 var configOptions options
 
 func configureConfigFile(vp *viper.Viper) {
-	vp.SetConfigName("tracetest")
-	vp.SetConfigType("yaml")
+	vp.SetConfigName("tracetest.yaml")
 	vp.AddConfigPath("/etc/tracetest")
 	vp.AddConfigPath("$HOME/.tracetest")
 	vp.AddConfigPath(".")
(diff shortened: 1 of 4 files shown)


server/config/config.go
```diff
@@ -128,9 +128,7 @@ func New(flags *pflag.FlagSet, logger logger) (*Config, error) {
 
 	configOptions.registerDefaults(vp)
 
-	warnAboutDeprecatedFields(vp, logger)
```
danielbdias, line 131: Do we want to remove these deprecation warnings?

mathnogueira (approved):

server/config/config.go
danielbdias: Or is it related to this PR: https://github.com/kubeshop/tracetest/pull/2082?

danielbdias (approved):

server/config/config.go
schoren: no, it's a mistake

schoren: @danielbdias i was trying to fix that on this pr, but decided to make another pr for that, and mistakenly removed this line. It's readded noe

xoscar (approved):

PR #2092, opened 2023-03-02
fix: test that checks if endpoint is reachable

mathnogueira:
This PR fixes the bug that @xoscar found. When testing an endpoint with a schema, the reachability would fail with "too many colons" error.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/tracedb/connection/connectivity_step.go
+++ b/server/tracedb/connection/connectivity_step.go
@@ -55,5 +55,6 @@ func (s *connectivityTestStep) TestConnection(_ context.Context) ConnectionTestS
 func ConnectivityStep(protocol Protocol, endpoints ...string) TestStep {
 	return &connectivityTestStep{
 		endpoints: endpoints,
+		protocol:  protocol,
 	}
 }


jorgeepc (approved):

schoren (changes requested):
can we add a test to cover this case?

mathnogueira:
> can we add a test to cover this case?

we can, but which endpoint would we try? Google? What about gRPC?

schoren:
@mathnogueira I was thinking more like a unit test to cover the parsing, not actual connectivy test

mathnogueira:
> @mathnogueira I was thinking more like a unit test to cover the parsing, not actual connectivy test

that would not catch the problem we had here. This error wasn't caused by the `url.Parse` method, it was returned by the`net.DialTcp` method. So, without a connection test, we would not cover this bug. I think it's plausible to have a real connection test here, I just don't know how we can simulate a gRPC endpoint.

schoren (approved):
as per talks with @mathnogueira we're not adding tests right now

PR #2098, opened 2023-03-02
fix: time conversion from grafana cloud

mathnogueira:
This PR fixes the time conversion from grafana cloud

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/traces/otel_http_converter.go
+++ b/server/traces/otel_http_converter.go
@@ -4,6 +4,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math"
+	"strconv"
 	"time"
 
 	"github.com/kubeshop/tracetest/server/model"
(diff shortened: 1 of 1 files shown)


server/traces/otel_http_converter.go
```diff
@@ -23,10 +24,17 @@ type httpScopeSpans struct {
...
+	ParentSpanId      string               `json:"parentSpanId"`
+	Kind              string               `json:"kind"`
+	StartTimeUnixNano string               `json:"startTimeUnixNano"`
+	EndTimeUnixNano   string               `json:"endTimeUnixNano"`
+	Attributes        []*httpSpanAttribute `json:"attributes"`
+	Status            *httpSpanStatus      `json:"status"`
+}
+
+type httpSpanStatus struct {
+	Code string `json:"code"`
```
mathnogueira, line 37: We were using the v1.Span Status Code field which is an `int`. Thus it would fail to convert the status "STATUS_OK" into an integer.

server/traces/otel_http_converter.go
```diff
@@ -61,12 +69,14 @@ func convertHttpOtelSpanIntoSpan(span *httpSpan) *model.Span {
 
 	var startTime, endTime time.Time
 
-	if span.GetStartTimeUnixNano() != 0 {
-		startTime = time.Unix(0, int64(span.GetStartTimeUnixNano()))
+	if span.StartTimeUnixNano != "" {
+		startTimeNs, _ := strconv.ParseInt(span.StartTimeUnixNano, 10, 64)
+		startTime = time.Unix(0, startTimeNs)
```
mathnogueira, line 74: Same problem, the value for StartTimeUnixNano is a string instead of int, so we have to get it as a string and convert it by hand here.

jorgeepc (approved):

xoscar (approved):
GJ my man!

PR #2102, opened 2023-03-03
chore(server): accept functional options in config

schoren:
The `config` object is needs to hold reference to different repositories to gather configs from DB, so it's much easier to handle the growing amount of dependencies by using the functional options patter.

This tiny pr just updates that internally without affecting any functionality.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"log"
 	"os"
 	"strings"
 	"sync"
@@ -29,6 +30,7 @@ type Config struct {
 	config *oldConfig
 	vp     *viper.Viper
 	mu     sync.Mutex
+	logger logger
 }
 
 type logger interface {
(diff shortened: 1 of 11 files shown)


server/testmock/database.go
```diff
@@ -92,10 +92,15 @@ func getMainDatabaseConnection(container *gnomock.Container) (*sql.DB, error) {
...
 	rand.Seed(time.Now().UnixNano())
-	randomInt := rand.Int()
-	newDatabaseName := fmt.Sprintf("%s_%d", baseDatabase, randomInt)
+	min := 1
+	max := 1000000
+	return rand.Intn(max-min) + min
+}
+
+func createRandomDatabaseForTest(db *sql.DB, baseDatabase string) (*sql.DB, error) {
+	newDatabaseName := fmt.Sprintf("%s_%d%d%d", baseDatabase, randomInt(), randomInt(), randomInt())
```
schoren, line 103: this change is kind of unrelated, but I got a few DB name collisions, so I figure this might make the tests more robust

server/config/config.go
```diff
@@ -117,44 +119,60 @@ func SetupFlags(flags *pflag.FlagSet) {
...
+func New(flags *pflag.FlagSet, confOpts ...configOption) (*Config, error) {
+	cfg := Config{
+		vp: viper.New(),
+	}
+
+	for _, coFn := range confOpts {
+		coFn(&cfg)
+	}
+
+	cfg.defaults()
```
schoren, line 139: the new approach makes it easier to set defaults, which is great for things like test setups

server/config/config.go
```diff
@@ -117,44 +119,60 @@ func SetupFlags(flags *pflag.FlagSet) {
...
 		return nil, fmt.Errorf("invalid configuration: %w", err)
 	}
 
-	return cfg, nil
+	return &cfg, nil
+}
+
+func (c *Config) defaults() {
+	if c.logger == nil {
+		c.logger = log.Default()
```
schoren, line 174: not sure if this should be the default, or we can create a `noopLogger` that does nothing. I'm open to suggestions

danielbdias: I believe this is ok for now. `zap` does something similar with the [`NewDevelopmentConfig`](https://pkg.go.dev/go.uber.org/zap#NewDevelopmentConfig) function.

danielbdias (approved):

mathnogueira (approved):

PR #2105, opened 2023-03-03
feat(server): consume new config

schoren:
This PR updates the config object to rely on the "config resource" repository to get the analytics config value, and drops support for the old `googleAnalytics.enabled` config value

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/config/analytics.go
```diff
@@ -1,14 +1,16 @@
...
+	"os"
+)
 
-func (c *Config) AnalyticsEnabled() bool {
+func (cfg *Config) AnalyticsEnabled() bool {
 	if os.Getenv("TRACETEST_DEV") != "" {
 		return false
 	}
 
-	c.mu.Lock()
```
schoren, line 10: the lock is only required when accessing values handled by viper

server/config/analytics_test.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		return config.WithConfigResource(repo)
+	}
+	t.Run("DefaultValues", func(t *testing.T) {
+		cleanEnv()
+		opt := setup()
+		cfg, _ := config.New(opt)
+
+		assert.True(t, cfg.AnalyticsEnabled())
```
schoren, line 40: The default used to be `false`, meaning disabled. I thought it might be useful to make it enabled by default. Thoughts @kdhamric?

server/config/configresource/config_resource.go
```diff
@@ -32,18 +32,32 @@ type repository struct {
...
+	defaultConfig := Config{
+		AnalyticsEnabled: true,
+	}
+	list, err := r.List(ctx, 1, 0, "", "", "")
+	if err != nil || len(list) != 1 {
+		// TODO: log error
+		return defaultConfig
+	}
+
+	return list[0]
```
jorgeepc, line 45: in case of multiple records in the DB you are using the first one, right?

mathnogueira: If that will be fixed in another PR, this one looks good to me

mathnogueira (approved):
LGTM

mathnogueira (approved):

PR #2111, opened 2023-03-06
Add pagination tests to resource manager

danielbdias:
This PR aims to add pagination tests to Resource Manager and update the tests for the first resource to consider that

## Changes

- Refactor resource manager tests to simplify the readability of each operation
- Added tests for resource manager
- Added pagination capabilities for config_resource
- Added pagination tests for config_resource

## Fixes

- https://github.com/kubeshop/tracetest/issues/2060

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/resourcemanager/resource_manager.go
+++ b/server/resourcemanager/resource_manager.go
@@ -32,9 +32,13 @@ type Resource[T ResourceSpec] struct {
 	Spec T      `mapstructure:"spec"`
 }
 
-type ResourceHandler[T ResourceSpec] interface {
-	SetID(T, id.ID) T
+type SortableHandler interface {
 	SortingFields() []string
+}
+
+type ResourceHandler[T ResourceSpec] interface {
+	SortableHandler
+	SetID(T, id.ID) T
 	List(_ context.Context, take, skip int, query, sortBy, sortDirection string) ([]T, error)
 	Count(_ context.Context, query string) (int, error)
 	Create(context.Context, T) (T, error)
--- a/server/resourcemanager/resource_manager.go
+++ b/server/resourcemanager/resource_manager.go
@@ -143,9 +143,6 @@ func (m *manager[T]) list(w http.ResponseWriter, r *http.Request) {
 		sortDirection,
 	)
 	if err != nil {
-		fmt.Println("---------------------------------------------------------------------------------------------------")
-		fmt.Println(err)
-		fmt.Println("---------------------------------------------------------------------------------------------------")
 		m.handleResourceHandlerError(w, "listing", err, encoder)
 		return
 	}
--- a/server/resourcemanager/resource_manager.go
+++ b/server/resourcemanager/resource_manager.go
@@ -147,6 +147,9 @@ func (m *manager[T]) list(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// TODO: the name "count" can be misleading when using pagination.
+	//       an user can paginate the request and see a different number
+	//       of records inside "item"
 	resourceList := ResourceList[T]{
 		Count: count,
 		Items: []map[string]any{},
--- a/server/resourcemanager/resource_manager.go
+++ b/server/resourcemanager/resource_manager.go
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"strings"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/gorilla/mux"
 	"github.com/kubeshop/tracetest/server/id"
 	"github.com/mitchellh/mapstructure"
@@ -32,6 +34,7 @@ type Resource[T ResourceSpec] struct {
 
 type ResourceHandler[T ResourceSpec] interface {
 	SetID(T, id.ID) T
+	SortingFields() []string
 	List(_ context.Context, take, skip int, query, sortBy, sortDirection string) ([]T, error)
 	Count(_ context.Context, query string) (int, error)
 	Create(context.Context, T) (T, error)
@@ -90,7 +93,7 @@ func getIntFromQuery(r *http.Request, key string) (int, error) {
 	return val, nil
 }
 
-func paginationParams(r *http.Request) (take, skip int, query, sortBy, sortDirection string, err error) {
+func paginationParams(r *http.Request, sortingFields []string) (take, skip int, query, sortBy, sortDirection string, err error) {
 	take, err = getIntFromQuery(r, "take")
 	if err != nil {
 		err = fmt.Errorf("error reading take param: %w", err)
@@ -104,6 +107,11 @@ func paginationParams(r *http.Request) (take, skip int, query, sortBy, sortDirec
 	}
 
 	sortBy = r.URL.Query().Get("sortBy")
+	if sortBy != "" && !slices.Contains(sortingFields, sortBy) {
+		err = fmt.Errorf("invalid sort field: %s", sortBy)
+		return
+	}
+
 	sortDirection = r.URL.Query().Get("sortDirection")
 
 	query = r.URL.Query().Get("query")
@@ -122,7 +130,7 @@ func (m *manager[T]) list(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	take, skip,
 		query, sortBy,
-		sortDirection, err := paginationParams(r)
+		sortDirection, err := paginationParams(r, m.handler.SortingFields())
 	if err != nil {
 		writeResponse(w, http.StatusBadRequest, fmt.Sprintf("cannot process request: %s", err.Error()))
 		return
--- a/server/resourcemanager/resource_manager.go
+++ b/server/resourcemanager/resource_manager.go
@@ -143,6 +143,9 @@ func (m *manager[T]) list(w http.ResponseWriter, r *http.Request) {
 		sortDirection,
 	)
 	if err != nil {
+		fmt.Println("---------------------------------------------------------------------------------------------------")
+		fmt.Println(err)
+		fmt.Println("---------------------------------------------------------------------------------------------------")
 		m.handleResourceHandlerError(w, "listing", err, encoder)
 		return
 	}
--- a/server/resourcemanager/testutil/operations.go
+++ b/server/resourcemanager/testutil/operations.go
@@ -31,6 +31,8 @@ var (
 
 		listNoResultsOperation,
 		listSuccessOperation,
+		listPaginatedAscendingSuccessOperation,
+		listPaginatedDescendingSuccessOperation,
 		// TODO: add tests for pagination etc
 	}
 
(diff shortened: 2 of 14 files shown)


server/resourcemanager/testutil/operations.go
```diff
@@ -1,115 +1,46 @@
...
 	"net/http"
+	"net/http/httptest"
 	"testing"
-
-	"github.com/stretchr/testify/require"
 )
 
+type Operation string
+
+type operationTester struct {
```
danielbdias, line 11: Folks, here is a polemic refactor that I would like to talk about with you later.

On one side, adding the operationTester as an interface is suitable for defining polymorphism and type-checking on golang. Still, for another, it generated tons of boilerplate code on our tests, making it hard to understand at first glance. 

One example was defining `postAssert` for each operation when we used it only 10% of the time. Seeing that operation was stateless, and I changed it for structs (in a js-way, you can kill me later 😛 ), omitting unused methods and reducing some declarations, like the `name` function.

Does it make sense to you? If needed, I can rollback this change and only add the pagination tests.

server/resourcemanager/testutil/operations.go
```diff
@@ -1,115 +1,46 @@
...
-		listInteralErrorOperation{},
+		createInternalErrorOperation,
+		updateInternalErrorOperation,
+		getInternalErrorOperation,
+		deleteInternalErrorOperation,
+		listInternalErrorOperation,
 	}
 )
-
-func removeID(input map[string]any) map[string]any {
```
danielbdias, line 40: As these functions are helpers, not related directly to operations, I've moved them to`common_functions.go`.

server/resourcemanager/testutil/operations_list.go
```diff
@@ -10,100 +10,166 @@ import (
...
+	name: OperationListPaginatedAscendingSuccess,
+	buildRequest: func(t *testing.T, testServer *httptest.Server, ct contentTypeConverter, rt ResourceTypeTest) *http.Request {
+		return buildListRequest(
+			rt.ResourceType,
+			map[string]string{
+				"take":          "2",
+				"skip":          "1",
+				"sortBy":        "id",
+				"sortDirection": "asc",
+				// TODO: think how to use "query" param
```
danielbdias, line 105: Probably I'll add a new test to test this parameter. But this is another point to discuss, do we want to pass queries using SQL strings "as is" to each endpoint (which is easier, but opens a breach to SQL injection), or do we want to use another way, like passing a stringified JSON with key and values for instance.

server/resourcemanager/testutil/test_resource.go
```diff
@@ -8,60 +8,17 @@ import (
...
-}
-
-type operationTester interface {
-	buildRequest(*testing.T, *httptest.Server, contentType, ResourceTypeTest) *http.Request
-	assertResponse(*testing.T, *http.Response, contentType, ResourceTypeTest)
-	name() Operation
-	postAssert(t *testing.T, ct contentType, rt ResourceTypeTest, testServer *httptest.Server)
-}
-
-var contentTypes = []contentType{
```
danielbdias, line 39: I've created a dedicated file just for these conventions: `content_type.go`

server/resourcemanager/resource_manager.go
```diff
@@ -147,6 +147,9 @@ func (m *manager[T]) list(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// TODO: the name "count" can be misleading when using pagination.
```
schoren, line 150: this field hold the `total count`, not the page count. I see how this can be confusing. feel free to rename the field

server/resourcemanager/testutil/operations_list.go
```diff
@@ -1,109 +1,251 @@
...
 
-func (op listNoResultsOperation) buildRequest(t *testing.T, testServer *httptest.Server, ct contentType, rt ResourceTypeTest) *http.Request {
-	return buildListRequest(
-		rt.ResourceType,
-		ct,
-		testServer,
-		t,
-	)
-}
+func getSafeSortFields(sortFields []string) []string {
```
schoren, line 31: a nil slice in is safe to use, no need for this func: https://go.dev/play/p/ULVYg48l1R-

server/resourcemanager/testutil/operations_list.go
```diff
@@ -1,109 +1,251 @@
...
+
+var listPaginatedAscendingSuccessOperation = operationTester{
+	name: OperationListPaginatedAscendingSuccess,
+	buildRequest: func(t *testing.T, testServer *httptest.Server, ct contentTypeConverter, rt ResourceTypeTest) *http.Request {
+		return buildListRequest(
+			rt.ResourceType,
+			map[string]string{
+				"take":          "2",
+				"skip":          "1",
+				"sortBy":        getSortFieldsString(rt.PaginationSortFields),
```
schoren, line 147: I think we should make one request per sorting field. I don't think we support sorting by multiple fields, so this request would actually fail

server/resourcemanager/testutil/operations_list.go
```diff
@@ -1,109 +1,251 @@
...
+
+		sortFields := getSafeSortFields(rt.PaginationSortFields)
+
+		for _, field := range sortFields {
+			for _, item := range parsedJsonBody.Items {
+				if prevVal == nil {
+					prevVal = item[field]
+					continue
+				}
+				assert.True(t, compareFieldsGreaterThan(prevVal, item[field]))
```
schoren, line 227: you can use `assert.GreaterOrEqual` so we don't have to maintain the type assertions

server/resourcemanager/testutil/operations_list.go
```diff
@@ -1,109 +1,251 @@
...
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func buildListRequest(rt string, ct contentType, testServer *httptest.Server, t *testing.T) *http.Request {
-	url := fmt.Sprintf("%s/%s/", testServer.URL, strings.ToLower(rt))
+func buildQueryString(params map[string]string) string {
```
schoren, line 15:
```go
func buildQueryString(params map[string]string) string {
  if len(params) == 0 {
    return ""
  }

  vals := url.Values{}
  for k, v := range params {
    vals.Add(k, v)
  }

  return "?"+vals.Encode()
}

```

schoren (changes requested):
There are a few needless functions that can be removed, but mainly we need to handle each sort field as a separated request instead of sending them all joined in the same request

server/resourcemanager/testutil/operations_list.go
danielbdias: After our discussion yesterday, it not makes sense for this function anymore, I deleted it.

server/resourcemanager/testutil/operations_list.go
danielbdias: Changed

server/resourcemanager/testutil/operations_list.go
```diff
@@ -1,109 +1,251 @@
...
+		}
+	},
 }
 
-func (listInteralErrorOperation) name() Operation {
-	return OperationListInteralError
+const OperationListPaginatedDescendingSuccess Operation = "ListPaginatedDescendingSuccess"
+
+var listPaginatedDescendingSuccessOperation = operationTester{
+	name: OperationListPaginatedDescendingSuccess,
```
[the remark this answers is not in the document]
danielbdias: I agree with that, but to do this change, we will need to change our operation structure to deal with more requests (today, it is fixed with only one request as part of the operation structure).

Can we unify this behavior in other PR? (to avoid increasing the scope of this PR)

server/resourcemanager/testutil/operations_list.go
danielbdias: It worked, thanks!

server/resourcemanager/testutil/operations_list.go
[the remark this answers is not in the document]
danielbdias: Never mind, after thinking for a while, our structure can have two `operationTester` with the same name. By doing that we could add both `asc` and `desc` in the same operation. 

It does make sense?

server/resourcemanager/testutil/operations.go
```diff
@@ -1,115 +1,47 @@
...
-		deleteNotFoundOperation{},
-		deleteSuccessOperation{},
+		deleteNotFoundOperation,
+		deleteSuccessOperation,
 
-		listNoResultsOperation{},
-		listSuccessOperation{},
+		listNoResultsOperation,
+		listSuccessOperation,
+		listWithInvalidSortFieldOperation,
```
schoren, line 34: you can move this to `errorOperations` since the test is actually testing how the resourcemanager itself handles an error, and is not directly related to each of the implementing resources

server/resourcemanager/resource_manager.go
danielbdias: No problem, I'll do this in another PR!

server/resourcemanager/testutil/operations.go
danielbdias: Ok!

server/resourcemanager/testutil/operations.go
```diff
@@ -1,115 +1,72 @@
...
+	assertResponse func(*testing.T, *http.Response, contentTypeConverter, ResourceTypeTest)
+	postAssert     func(*testing.T, contentTypeConverter, ResourceTypeTest, *httptest.Server)
 }
 
-func extractID(input string) string {
-	parsed := parseJSON(input)
-	id := parsed["spec"].(map[string]any)["id"]
-	if id == nil {
-		return ""
+func buildSingleStepOperation(operation singleStepOperationTester) operationTester {
```
schoren, line 29: nice!

server/resourcemanager/testutil/test_resource.go
```diff
@@ -81,53 +41,62 @@ func TestResourceTypeOperations(t *testing.T, rt ResourceTypeTest, operations []
...
 	if rt.Prepare != nil {
-		rt.Prepare(t, op.name(), testBridge)
+		rt.Prepare(t, op.name, testBridge)
 	}
 
-	req := op.buildRequest(t, testServer, ct, rt)
-	resp := doRequest(t, req, ct, testServer)
+	// minor hack to pass "sortFields" to each operation
+	// without needing to propagate sortableHandler for each one
+	sortableHandler := testBridge.(rm.SortableHandler)
```
schoren, line 81: Let people know what is missing

```suggestion
	sortableHandler, ok := testBridge.(rm.SortableHandler)
	if !ok {
		panic("provided testBridge does not implement `resourcemanager.SortableHandler` interface")
	}
```

schoren (approved):
Looking awesome!!! thanks for bearing with my requests. I left an optional suggestion just to be informative, but feel free to merge as is

server/resourcemanager/testutil/test_resource.go
danielbdias: makes sense! I'll open another PR to add that

Issue #2118, opened 2023-03-07
OTLP implementations using Datadog, Lightstep and New Relic as a Data Store instead of OTLP crashes when trying to run test


danielbdias:
**Describe the bug**

With Datadog, Lightstep, and New Relic, we can use them as a Data Store provisioning the OTLP datastore or a specific Data Store for them. However, setting up a specific Data Store crashes Tracetest on a Test run.

**To Reproduce**

1. Go to 'tracetest-datadog', 'tracetest-new-relic' or 'lightstep' example
2. Change the collector file adding the API Key
3. Change tracetest-provision to use `datadog`, `newrelic` or `lightstep` instead of `otlp`
4. Run a test through the console
5. See the following error:
<img width="1250" alt="image" src="https://user-images.githubusercontent.com/12397879/223433944-ba3dc13b-83fa-4b2e-af95-025955847cc3.png">

**Expected behavior**
It should work as the `otlp` provider

PR #2121, opened 2023-03-07
Fixing bug on TraceDB instantiation

danielbdias:
This PR fixes a bug found on [#2118](https://github.com/kubeshop/tracetest/issues/2118): 
With Datadog, Lightstep, and New Relic, we can use them as a Data Store provisioning the OTLP datastore or a specific Data Store for them. However, setting up a specific Data Store crashes Tracetest on a Test run.

## Changes

- Fix Data Store instantiation

## Fixes

- #2118 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/tracedb/tracedb.go
+++ b/server/tracedb/tracedb.go
@@ -60,7 +60,15 @@ func Factory(repo model.Repository) func(ds model.DataStore) (TraceDB, error) {
 	return f.New
 }
 
-func (f *traceDBFactory) New(ds model.DataStore) (tdb TraceDB, err error) {
+func (f *traceDBFactory) getTraceDBInstance(ds model.DataStore) (TraceDB, error) {
+	var tdb TraceDB
+	var err error
+
+	if ds.IsOTLPBasedProvider() {
+		tdb, err = newCollectorDB(f.repo)
+		return tdb, err
+	}
+
 	switch ds.Type {
 	case model.DataStoreTypeJaeger:
 		tdb, err = newJaegerDB(ds.Values.Jaeger)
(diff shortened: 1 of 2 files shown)


server/tracedb/tracedb.go
```diff
@@ -60,7 +60,15 @@ func Factory(repo model.Repository) func(ds model.DataStore) (TraceDB, error) {
 	return f.New
 }
 
-func (f *traceDBFactory) New(ds model.DataStore) (tdb TraceDB, err error) {
+func (f *traceDBFactory) getTraceDBInstance(ds model.DataStore) (TraceDB, error) {
+	var tdb TraceDB
+	var err error
+
+	if ds.IsOTLPBasedProvider() {
+		tdb, err = newCollectorDB(f.repo)
```
xoscar, line 68: Where exactly was the bug?

xoscar (approved):
LGTM!

server/tracedb/tracedb.go
danielbdias: It was on our `switch/case` statement. Today we do this:
```go
switch ds.Type {
//...
    case model.DataStoreTypeNewRelic:
    case model.DataStoreTypeLighStep:
    case model.DataStoreTypeDataDog:
    case model.DataStoreTypeOTLP:
        tdb, err = newCollectorDB(f.repo)
//...
}
```

Instead of this:
```go
switch ds.Type {
//...
    case model.DataStoreTypeNewRelic,
         model.DataStoreTypeLighStep,
         model.DataStoreTypeDataDog,
         model.DataStoreTypeOTLP:
        tdb, err = newCollectorDB(f.repo)
//...
}
```

Both are valid statements in golang, but the first one only executes `newCollectorDB` for the type `model.DataStoreTypeOTLP` and assumes that for the other cases we don't have statements.

On the second, as we are aggregating four types on the same `case`, they all execute `newCollectorDB`.

Since we now have the function `IsOTLPBasedProvider` to encapsulate the logic of which providers are OTLP-based, I preferred to create this if and reduce one place to update when we need to add other providers.

PR #2126, opened 2023-03-07
feature(cli): adding support for config resource

xoscar:
This PR adds support for the config resource and includes the base work for the rest of the resources to come.

## Changes

- Adds a new apply, list, and delete commands
- Implements the apply command for the config resource

## Fixes

- #2114

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/actions/resource.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http"
+
+	"go.uber.org/zap"
+)
+
+type ApplyArgs struct {
+	File string
+}
+
+type ResourceActions interface {
```
xoscar, line 15: Resource interface which is going to establish the methods for each supported resource

cli/cmd/apply_cmd.go
```diff
@@ -0,0 +1,60 @@
...
+
+var definitionFile string
+
+var applyCmd = &cobra.Command{
+	Use:    "apply [resource type]",
+	Short:  "Apply (create/update) resources to your Tracetest server",
+	Long:   "Apply resources",
+	PreRun: setupCommand(),
+	Args:   cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
```
xoscar, line 23: In sequence for this command, we do:
1. grab the resource type
2. get the resource client
3. get the resource actions instance
4. apply the command

cli/cmd/config.go
```diff
@@ -140,3 +144,27 @@ func getAPIClient() *openapi.APIClient {
 	}
 	return openapi.NewAPIClient(config)
 }
+
+func getResourceClient(resourceType string) (http.Client, http.Request) {
```
xoscar, line 148: As we don't have the open api specs I added this basic HTTP client to push YAML files to the backend services

cli/cmd/list_cmd.go
```diff
@@ -0,0 +1,21 @@
+package cmd
+
```
xoscar, line 2: WIPS

server/model/yaml/encoding.go
```diff
@@ -50,6 +51,13 @@ func Decode(contents []byte) (File, error) {
 			return File{}, fmt.Errorf("cannot decode datastore: %w", err)
 		}
 		f.Spec = dataStore
+	case FileTypeConfig:
```
xoscar, line 54: We now support config resource files

cli/actions/config.go
```diff
@@ -0,0 +1,73 @@
...
+	if fileContent.Definition().Type != "Config" {
+		return fmt.Errorf(`file must be of type "Config"`)
+	}
+
+	body := bytes.NewBufferString(fileContent.Contents())
+
+	config.request.Body = io.NopCloser(body)
+	config.request.Method = http.MethodPost
+
+	res, err := config.client.Do(&config.request)
```
schoren, line 54: you should create a new request every time the func is called: 

```suggestion
req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(fileContent.Contents())
if err != nil {
  // handle error, but this should never happen
}

resp, err := config.client.Do(req)
```

cli/actions/resource.go
```diff
@@ -0,0 +1,47 @@
...
+	request http.Request
+}
+
+type supportedResources string
+
+var (
+	Config supportedResources = "config"
+)
+
+func NewResourceActions(resourceType string, logger *zap.Logger, client http.Client, request http.Request) (ResourceActions, error) {
```
schoren, line 34: this looks like a great place to use functional options 😋

cli/cmd/config.go
schoren: Instead of returning a request, you can return a custom type with a reference to the client and the url or any other info needed. Then each consumer can build their own request.

go.work.sum
```diff
@@ -182,6 +180,7 @@ github.com/googleapis/gax-go/v2 v2.2.0/go.mod h1:as02EH8zWkzwUoLbBaFeQ+arQaj/Oth
 github.com/googleapis/gax-go/v2 v2.3.0/go.mod h1:b8LNqSzNabLiUpXKkY7HAR5jr6bIT99EXz9pXxye9YM=
 github.com/googleapis/gax-go/v2 v2.4.0/go.mod h1:XOTVJ59hdnfJLIP/dh8n5CGryZR2LxK9wbMD5+iXC6c=
 github.com/googleapis/go-type-adapters v1.0.0/go.mod h1:zHW75FOG2aur7gAO2B+MLby+cLsWGBF62rFAi7WjWO4=
+github.com/gophercloud/gophercloud v0.23.0 h1:I3P10oKlGu3DHP9PrEWMr1ya+/+3Rc9uRHNkRZ9wO7g=
```
schoren, line 183: what's this dep?

cli/actions/resource.go
```diff
@@ -0,0 +1,47 @@
...
+		client:  client,
+		request: request,
+	}
+
+	switch resourceType {
+	case string(Config):
+		return NewConfigActions(args), nil
+	default:
+		return nil, fmt.Errorf("unsupported resource type: %s", resourceType)
+	}
```
schoren, line 46: I think we can skip this switch, and use a more polymorphyc approach. Using a registry of some kind, you can have each resource register itself against a centralized registry, and have that registry get the requirec `ResourceActions` for you.

This will allow to easily implement new resource types, reduce the chance of "forgetting to add to the switch here or there" kind of errors, etc

schoren (changes requested):

cli/cmd/apply_cmd.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+)
+
+var definitionFile string
+
+var applyCmd = &cobra.Command{
+	Use:    "apply [resource type]",
+	Short:  "Apply (create/update) resources to your Tracetest server",
+	Long:   "Apply resources",
```
danielbdias, line 20: The long version is shorter than the short version, hehehehe

Maybe we can set both as the same text for now:

```suggestion
	Short:  "Create or update a resource to your Tracetest server",
	Long:   "Create or update a resource to your Tracetest server",
```

cli/actions/resource.go
```diff
@@ -0,0 +1,67 @@
...
+
+type resourceArgs struct {
+	logger *zap.Logger
+	client *openapi.APIClient
+}
+
+type resourceArgsOption = func(*resourceArgs)
+
+type ResourceRegistry map[string]ResourceActions
+type supportedResources string
```
danielbdias, line 30: Since we are typing resource names as `supportedResources`, we could turn this type public and define the `ResourceRegistry` as `map[supportedResources]ResourceActions`, does it make sense?

A suggestion of refactor:

```suggestion
type ResourceRegistry map[SupportedResource]ResourceActions
type SupportedResource string
```

cli/cmd/list_cmd.go
danielbdias: Still WIP?

cli/cmd/list_cmd.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+)
+
+var listCmd = &cobra.Command{
+	Use:    "list [resource type]",
+	Long:   "List resources from your Tracetest server",
+	Short:  "List resources",
+	PreRun: setupCommand(),
+	Args:   cobra.MinimumNArgs(1),
```
danielbdias, line 18: To think in the future: should we add optional parameters to sort outputs and filter results?

cli/utils/api.go
```diff
@@ -0,0 +1,69 @@
...
+	transport := &http.Transport{
+		DialContext: (&net.Dialer{
+			Timeout: time.Second,
+		}).DialContext,
+	}
+
+	client := http.Client{
+		Transport: transport,
+		Timeout:   30 * time.Second,
+	}
```
danielbdias, line 43: Both timeouts could be constants to make them explicit and to allow us to refactor and expose them as a CLI configuration of the future.

server/model/yaml/encoding.go
danielbdias: @xoscar can you add a test for this structure on `encoding_test.go`, please?

cli/cmd/list_cmd.go
xoscar: I think those can be handled using flags, that will be part of the next pr when introducing polling profiles

danielbdias (approved):

server/model/yaml/encoding_test.go
```diff
@@ -77,6 +78,23 @@ func TestDecode(t *testing.T) {
...
+					Name:             "config",
+					AnalyticsEnabled: true,
+				},
+			},
+			testSpecFn: func(t *testing.T, expected, actual yaml.File) {
+				config, err := actual.Config()
+				require.NoError(t, err)
+				assert.Equal(t, expected.Spec.(configresource.Config), config)
+			},
+		},
```
danielbdias, line 97: ❤️

PR #2127, opened 2023-03-07
feat(server): make config unique

schoren:
This PR makes the `config` resource unique. This means that we won't have many rows in a table, just a single record that hold the latest config.

To this end, we are "hardcoding" the ID and name of the resource so we don't create conflict with random IDs and original names, that are actually ignored in the DB.

This PR also modifies the resource manager to allow defining resources that don't use the entire CRUD interface. For example the config resource only has the `Get` and `Update` methods. 

As a side project, it fixes some incorrectness in the pagination tests.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/resourcemanager/operations.go
```diff
@@ -0,0 +1,163 @@
...
+	OperationGet,
+	OperationDelete,
+}
+
+type SortableHandler interface {
+	SortingFields() []string
+}
+
+type List[T ResourceSpec] interface {
+	SortableHandler
```
xoscar, line 34: This looks a little bit complex to me, why do we need this file?

server/resourcemanager/resource_manager.go
```diff
@@ -32,55 +32,109 @@ type Resource[T ResourceSpec] struct {
...
 	// prefix is /{resourceType | lowercase}/
 	subrouter := r.PathPrefix("/" + strings.ToLower(m.resourceType)).Subrouter()
 
-	subrouter.HandleFunc("/", m.list).Methods(http.MethodGet).Name(fmt.Sprintf("%s.List", m.resourceType))
-	subrouter.HandleFunc("/", m.create).Methods(http.MethodPost).Name(fmt.Sprintf("%s.Create", m.resourceType))
-	subrouter.HandleFunc("/", m.update).Methods(http.MethodPut).Name(fmt.Sprintf("%s.Update", m.resourceType))
+	enabledOps := m.EnabledOperations()
 
-	subrouter.HandleFunc("/{id}", m.get).Methods(http.MethodGet).Name(fmt.Sprintf("%s.Get", m.resourceType))
-	subrouter.HandleFunc("/{id}", m.delete).Methods(http.MethodDelete).Name(fmt.Sprintf("%s.Delete", m.resourceType))
```
xoscar, line 73: I like how dynamic and flexible this is!

xoscar (approved):
Hey dude I left some non-blocking questions but this looks good to me!

server/resourcemanager/testutil/operations_list.go
```diff
@@ -155,29 +163,28 @@ func buildPaginationOperationStep(sortDirection, sortField string) operationTest
...
 			json.Unmarshal([]byte(jsonBody), &parsedJsonBody)
 
 			require.Equal(t, 3, parsedJsonBody.Count)
+			require.Greater(t, len(parsedJsonBody.Items), 1)
 
 			var prevVal any
-			field := rt.sortFields[0]
-
 			for _, item := range parsedJsonBody.Items {
+				itemSpec := item["spec"].(map[string]any)
```
schoren, line 176: this test was never doing anything because `item[field]` was always nil, because the structure was incorrectly parsed.

server/resourcemanager/testutil/operations_list.go
```diff
@@ -150,43 +158,49 @@ func buildPaginationOperationStep(sortDirection, sortField string) operationTest
...
+					asserted++
 				} else {
-					assert.GreaterOrEqual(t, prevVal, item[field])
+					assert.GreaterOrEqual(t, prevVal, itemSpec[sortField])
+					asserted++
 				}
-
-				prevVal = item[field]
 			}
+			assert.Equal(t, len(parsedJsonBody.Items), asserted)
```
schoren, line 194: with this check we make sure that we're checking all the items in the list

server/config/configresource/config_resource.go
```diff
@@ -8,8 +8,14 @@ import (
 	"os"
 
 	"github.com/kubeshop/tracetest/server/id"
+	"github.com/kubeshop/tracetest/server/resourcemanager"
 )
 
+var Operations = []resourcemanager.Operation{
```
schoren, line 14: I don't like having the resource depending on `resourcemanager` just to provide a glorified list of strings, but putting this here makes it easy to share the config everywhere (app.go, tests, etc), and if it generates circular deps at somepoint (not sure how), we can move it to some other place

server/config/configresource/config_resource.go
```diff
@@ -54,245 +60,102 @@ func Repository(db *sql.DB, opts ...option) *repository {
...
 
-func (r *repository) Current(ctx context.Context) Config {
+func (r *Repository) Current(ctx context.Context) Config {
 	defaultConfig := Config{
 		AnalyticsEnabled: true,
 	}
 
-	list, err := r.List(ctx, 1, 0, "", "", "")
-	if err != nil || len(list) != 1 {
+	cfg, err := r.Get(ctx, id.ID("current"))
```
schoren, line 90: the id doesn't really exists, so I just picked `current` so we can have a non empty id that makes sense, for consistency

server/config/configresource/config_resource.go
```diff
@@ -54,245 +60,100 @@ func Repository(db *sql.DB, opts ...option) *repository {
...
+		QueryRowContext(ctx, selectQuery).
 		Scan(
-			&cfg.ID,
-			&cfg.Name,
 			&cfg.AnalyticsEnabled,
 		)
 
 	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return cfg, nil
```
schoren, line 114: since `config` must always have a value, `not found` actually means, `defaults not overriden`

server/config/configresource/config_resource.go
```diff
@@ -54,245 +60,100 @@ func Repository(db *sql.DB, opts ...option) *repository {
...
+const (
+	deleteQuery = "DELETE FROM config"
+	insertQuery = `INSERT INTO config ("analytics_enabled") VALUES ($1)`
+)
 
-func (r *repository) Delete(ctx context.Context, id id.ID) error {
-	cfg, err := r.Get(ctx, id)
-	if err != nil {
-		return err
+func (r *Repository) Update(ctx context.Context, updated Config) (Config, error) {
```
schoren, line 127: This func is hard to review, but essentialy it dele any existing config and creates a new one with the updated values

server/resourcemanager/operations.go
schoren: We need to be able to selectivly enable endpoints. the code in this file enables that behaviour

server/resourcemanager/operations.go
```diff
@@ -0,0 +1,163 @@
...
+
+type Get[T ResourceSpec] interface {
+	Get(context.Context, id.ID) (T, error)
+}
+
+type Delete[T ResourceSpec] interface {
+	Delete(context.Context, id.ID) error
+}
+
+type resourceHandler[T ResourceSpec] struct {
```
schoren, line 56: The idea here was to replace the interface with a struct that can have "optional" functions. This allows for partial resources, like config, to only implement the required methods.

Note that this is an unexported struct. It's mapped when creating a new resourcemanager, so clients don't need to know how this works. Always keep in mind that the idea of this package is to be very flexible, and require as little coupling as possible

server/resourcemanager/operations.go
```diff
@@ -0,0 +1,163 @@
...
+	List          func(_ context.Context, take, skip int, query, sortBy, sortDirection string) ([]T, error)
+	Count         func(_ context.Context, query string) (int, error)
+	SortingFields func() []string
+	Create        func(context.Context, T) (T, error)
+	Update        func(context.Context, T) (T, error)
+	Get           func(context.Context, id.ID) (T, error)
+	Delete        func(context.Context, id.ID) error
+}
+
+func (rh *resourceHandler[T]) bindOperations(enabledOperations []Operation, handler any) error {
```
schoren, line 67: this func validates and configures the actual functions being used

server/resourcemanager/resource_manager.go
```diff
@@ -32,55 +32,109 @@ type Resource[T ResourceSpec] struct {
...
-type ResourceHandler[T ResourceSpec] interface {
-	SortableHandler
-	SetID(T, id.ID) T
-	List(_ context.Context, take, skip int, query, sortBy, sortDirection string) ([]T, error)
-	Count(_ context.Context, query string) (int, error)
-	Create(context.Context, T) (T, error)
-	Update(context.Context, T) (T, error)
-	Get(context.Context, id.ID) (T, error)
-	Delete(context.Context, id.ID) error
+type Manager interface {
```
schoren, line 35: this interface is mostly useful for passing managers in tests without needing to know the generic concrete type

server/resourcemanager/resource_manager.go
```diff
@@ -32,55 +32,109 @@ type Resource[T ResourceSpec] struct {
...
+
+	cfg := config{
+		enabledOperations: availableOperations,
+		idgen:             func() id.ID { return id.ID("") },
+	}
+
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+
```
schoren, line 78: functional options FTW!!

server/resourcemanager/resource_manager_test.go
```diff
@@ -20,38 +20,76 @@ func TestSampleResource(t *testing.T) {
...
-		SomeValue: "the value 3",
-	}
-
 	sampleUpdated := sampleResource{
 		ID:        "1",
 		Name:      "the name updated",
 		SomeValue: "the value updated",
 	}
 
+	prepareSortByID := func(m *sampleResourceManager) {
```
schoren, line 29: this code is unrelated to this PR directly, but I realized the behaviour of the sort test was incorrect. the order could be coming from any field, not the specific one being tested. While this is not very relevant for this specific tests, 
this test serves as a "meta" test for the test suit itself. This guarantees that the test works as expected when actual resources are implemented.

server/resourcemanager/resource_manager_test.go
```diff
@@ -164,6 +200,89 @@ func TestSampleResource(t *testing.T) {
 	})
 }
 
+func TestRestrictedResource(t *testing.T) {
```
schoren, line 203: This test validates the behaviour for "partial" resources (meaning, ones that don't implement the entire CRUD interface)

server/resourcemanager/resource_manager.go
```diff
@@ -32,55 +32,109 @@ type Resource[T ResourceSpec] struct {
...
+	err := rh.bindOperations(cfg.enabledOperations, handler)
```
schoren, line 79: You may be wandering: Why do we need to provide an explicit list of `enabledOperations` if we can infer them from the given `handler`?
You might notice that the `bindOperations` actually checks tha the `handler` has the requried methods and "binds" them to the internal `resourceHandler`. Why don't just rely on that?

The answer is, to double check. This applies mainly to the case when a developer is creating a new, full CRUD resource. If we just rely on the information we can infer, it could happen that a dev forgets to implement the `create` method in the repository, and that would make the ResourceManager to *not* "activate" the create endpoint.

So, by forcing developers to be explicit about resources not being fully CRUD, we have an extra validation that the implemented behaviour is the expected one.

server/resourcemanager/testutil/test_resource.go
```diff
@@ -70,16 +90,20 @@ func testOperationForContentType(t *testing.T, op operationTester, ct contentTyp
...
 	if rt.Prepare != nil {
-		rt.Prepare(t, op.name, testBridge)
+		rt.Prepare(t, op.name, manager)
 	}
 
-	// minor hack to pass "sortFields" to each operation
-	// without needing to propagate sortableHandler for each one
-	sortableHandler := testBridge.(rm.SortableHandler)
-	rt.sortFields = sortableHandler.SortingFields()
+	if !op.needsToRun(manager.EnabledOperations()) {
```
schoren, line 104: this again is a double check to ensure that we are running the tests that we need to run based on the explicit definition and not from infered information. See [this comment](https://github.com/kubeshop/tracetest/pull/2127/files#diff-73567862de3c0b65102c45d3df1f98a49f2649fe588140f55a56fa693068bf51R79)

server/resourcemanager/resource_manager_test.go
danielbdias: Thinking about it that make sense to change the test name to something like `TestResourceWithPartialOperations`?
By looking at its name the first thing that I associated is that a `RestrictedResource` is a `private` one.

server/resourcemanager/testutil/operations_list.go
danielbdias: Wondering if it makes sense: 
We are assuming the structure of the `Resource[T ResourceSpec]` here, but usually, we want to work only with the `ResourceSpec`. Makes sense to have "wrap/unwrap" function in our tests to make this behavior explicit? 

Why? We can use it here to "unwrap" resource and test it and also use it on `SampleJSON` and `SampleJSONUpdated` to pass the "raw" resource as a go object, and later convert it to JSON and wrap it on a `Resource` json.

Does it make sense?

danielbdias (approved):

server/resourcemanager/testutil/operations_list.go
schoren: THe idea of these tests is to "black box" test as much as possible of a resource. The `Resource[T]` is not something we assume, but a standard format for the resourcemanager. all api calls/responses wrap around that. The tests must be as "from the user perspective" as possible, so we don't hide any behaviours

Issue #2134, opened 2023-03-09
bind source path does not exist


fangzongzhou:
**Describe the bug**
when i follow the step from https://docs.tracetest.io/getting-started/installation/  

the result is ` bind source path does not exist`



**To Reproduce**
exec 
```
docker compose -f tracetest/docker-compose.yaml  up -d
```


**Screenshots**
output is 
```
➜  workspace docker compose -f tracetest/docker-compose.yaml  up -d
[+] Running 3/3
 ⠿ demo-worker Pulled                                                                                                                                                                                                                                                                                                                                                  4.1s
 ⠿ demo-api Pulled                                                                                                                                                                                                                                                                                                                                                     3.7s
 ⠿ demo-rpc Pulled                                                                                                                                                                                                                                                                                                                                                     3.7s
[+] Running 0/0
 ⠋ Container tracetest-tracetest-1  Creating                                                                                                                                                                                                                                                                                                                           0.1s
Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /host_mnt/Users/fangzongzhou/workspace/tracetest/tracetest-provision.yaml
```




**Additional context**
my device is mac M1 

docker desktop 

```
Version 4.17.0 (99724)

Engine: 20.10.23

Compose: v2.15.1

Credential Helper: v0.7.0

Kubernetes: v1.25.4
```

danielbdias:
Thanks for reporting it @fangzongzhou . We are checking this bug and fixing it this week. As soon as release a version we will let you know!

danielbdias:
hi, @fangzongzhou! We released version `v0.10.1` today with a bug fix for this problem. We you have time, can you update `tracetest` CLI through `brew upgrade tracetest` and try again?

danielbdias:
hi folks! I'll close this issue now since it is fixed on `v0.10.1`.  Feel free to ping us in case of any doubt!

PR #2138, opened 2023-03-09
Added fixes to docker compose installer

danielbdias:
This PR aims to fix a bug added on `v0.10.0`, where when a user asks to install tracetest and choose a `docker compose` installation with a backend, we install an incomplete version of tracetest, without a provisioning file.

**Note**: this branch was created directly from tag `v0.10.0` and this PR should merged with `hotfix-v0.10.1-rc`. Later we can cherry-pick this commit and add it to `main`.

## Changes

- Installer to consider provisioning changes

## Fixes

- https://github.com/kubeshop/tracetest/issues/2134

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/installer/docker_compose.go
```diff
@@ -106,12 +111,19 @@ func getDockerComposeFileContents(ui cliUI.UI, config configuration) []byte {
 		include = append(include, "cache", "queue", "demo-api", "demo-worker", "demo-rpc", "otel-collector")
 	}
 
-	filterContainers(ui, project, include)
+	// filter and update project
+	filterAndFixContainers(ui, project, include)
```
danielbdias, line 115: Tech debt to the future: even changing the container to consider the demo projects, we didn't consider the initialization order of the containers.

Today Tracetest can start before `PokeshopAPI`, which can make our users try to run a test and receive an error because the API wasn't started yet.

cli/installer/templates/config.yaml.tpl
```diff
@@ -0,0 +1,32 @@
...
+    otelProductCatalog: {{ .otelProductCatalog }}
+    otelCart: {{ .otelCart }}
+    otelCheckout: {{ .otelCheckout }}
+
+experimentalFeatures: []
+
+{{ if eq .installBackend "true" }}
+server:
+  telemetry:
+    dataStore: {{ .backendType }}
```
jorgeepc, line 31: Do we need this line in the config file?

jorgeepc (commented):
Looks good. Added one question.

cli/installer/templates/config.yaml.tpl
mathnogueira: I don't think we need it because we have the provision file

mathnogueira (approved):
Left one comment. But overall, great job

cli/installer/templates/config.yaml.tpl
danielbdias: You are right folks! I've removed this part from the config file.

PR #2141, opened 2023-03-10
Added fixes to docker compose installer (#2138)

jorgeepc:
* Added fixes to docker compose installer

* Removing unused section on config

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


danielbdias (approved):

danielbdias:
Due to a problem with our chocolatey distribution and having a tag associated with the branch, Github will block this merge. 😢 

I'm thinking if we don't need to make a "copy" of the b51bc98d31f349c0316887f7e8f1ef442daa9b5f commit and merge it on main instead of merging `hotfix-v0.10.1` directly.

@mathnogueira @xoscar as two Git wizards what do you think we can do? It makes sense to merging `hotfix-v0.10.1` or creating a copy of  b51bc98d31f349c0316887f7e8f1ef442daa9b5f just for main?

We could try to rebase and put b51bc98d31f349c0316887f7e8f1ef442daa9b5f just after the `v0.10.0` commit, but maybe doing a `git push --force` on main could be bad.

jorgeepc:
> Due to a problem with our chocolatey distribution and having a tag associated with the branch, Github will block this merge. 😢
> 
> I'm thinking if we don't need to make a "copy" of the [b51bc98](https://github.com/kubeshop/tracetest/commit/b51bc98d31f349c0316887f7e8f1ef442daa9b5f) commit and merge it on main instead of merging `hotfix-v0.10.1` directly.
> 
> @mathnogueira @xoscar as two Git wizards what do you think we can do? It makes sense to merging `hotfix-v0.10.1` or creating a copy of [b51bc98](https://github.com/kubeshop/tracetest/commit/b51bc98d31f349c0316887f7e8f1ef442daa9b5f) just for main?
> 
> We could try to rebase and put [b51bc98](https://github.com/kubeshop/tracetest/commit/b51bc98d31f349c0316887f7e8f1ef442daa9b5f) just after the `v0.10.0` commit, but maybe doing a `git push --force` on main could be bad.

I would prefer to merge the hotfix branch. If our plan is to use gitflow, this is the recommended approach.

danielbdias:
> I would prefer to merge the hotfix branch. If our plan is to use gitflow, this is the recommended approach.

I believe that we can bypass the chocolatey check for now.

Issue #1937, opened 2023-02-02
[Config] Enable flexibility and simplicity in configuring  Tracetest 


kdhamric:
Goals:

- minimize the complexity of configuring Tracetest
- enable single command automated install & configuration
- minimize complexity in our documentation / blog posts
- enable configuration via the UI & CLI after installation
- make the tracetest.yaml file so it is minimized to the bare minimum

Tasks:
- [ ] Implement Viper
- [ ] Move analytics into UI
- [ ] Enable setting analytics via CLI
- [ ] Move polling settings into the UI
- [ ] Enable setting polling via the CLI
- [ ] Move demo configurations into the UI
- [ ] Enable setting demo configuration via the CLI
- [ ] Ability to provision Tracetest server configuration automatically
- [ ] Update examples to use new install method / config
- [ ] Update docs to use new install method / config
- [ ] Discuss how to handle exporter section in config file

schoren:
# Proposal

We can split the config into 3 main categories: 
- `server` config includes values that are **required** to start a server instance,
- `user` config includes values more dynamic in nature, and very user oriented.
- `initialization` used to load initial configs when first installing. 

In detail:

### Server Config
Included values:
- postgres:
	- host
	- user
	- pasword
	- database
	- params
- server:
	- port
	- pathPrefix
- experimental features (list of enabled, temporary, feature flags)
- internal telemetry: (required for internal testing)
	- enabled
	- otelcollector address

### User config

- Data Stores
- analytics enabled
- Trace Pull Strategies:
	- Timeout:
		- max wait time for trace
		- retry delay
- demo:
	- PokeShop
		- enabled
		- http endpoint
		- grpc endpoint
	- otel store
		- enabled
		- Frontend endpoint
		- ProductCatalog endpoint
		- Cart endpoint
		- Checkout endpoint

### Initialization
must be kept equal to  `user` config.

## Where each config should live

- `server` config can be stored in ENV vars, config file, or passed via flags on server startup
- `user` settings must be kept in DB, so they can easily be updated via UI.
- `initialization`  should be a yaml file matching the `user` settings structure. It can be read on first run and inserted into DB, and then ignored for the rest of the runs. This is useful for fleet or automated initialization.
[...]

PR #2144, opened 2023-03-10
feature(cli): adding support for polling profile resources

xoscar:
This PR adds support for the polling profiles resource to be handled by the CLI

## Changes

- Adds the delete, get list and apply logic for polling profiles

## Fixes

- #2115 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/93b1b86b0e6c436295382ddbedd5dc10


cli/formatters/pollingprofile.go
```diff
@@ -0,0 +1,66 @@
+package formatters
+
```
xoscar, line 2: We need to refactor the way we handle formatters in the future

cli/cmd/config.go
```diff
@@ -48,8 +48,11 @@ func setupCommand(options ...setupOption) func(cmd *cobra.Command, args []string
 
 		options := []actions.ResourceArgsOption{actions.WithClient(apiClient), actions.WithLogger(cliLogger), actions.WithConfig(cliConfig)}
 		configActions := actions.NewConfigActions(options...)
+		pollingActions := actions.NewPollingActions(options...)
 		resourceRegistry.Register(actions.SupportedResourceConfig, configActions)
 
+		resourceRegistry.Register(actions.SupportedResourcePollingProfile, pollingActions)
```
schoren, line 54: Having constants for supported resources defined in the `actions` package makes this package a central dependency. You can just use regular strings defined as constants within each resource package (or file, whatever applies).

Keep in mind the usecase. no code is actually consuming these constants, only when the resource registers itself. There is no other direct reference to this constant, so it's only creating a hardcoded value that is not actually consumed

cli/cmd/delete_cmd.go
```diff
@@ -1,21 +1,55 @@
...
+			return
+		}
+
+		cmd.Println(pterm.FgGreen.Sprintf(fmt.Sprintf("✔ Resource deleted successfully for resource type: %s", resourceType)))
 	},
 	PostRun: teardownCommand,
 }
 
 func init() {
+	deleteCmd.PersistentFlags().StringVarP(&deletedResourceID, "identifier", "i", "", "id of the resource to delete")
```
schoren, line 53: are you sure this needs to be a persistent flag? If i'm not mistaken, persistent flags are "inherited" by subcommands, while "not-persistent" flags are specific to the command only.

schoren: also, I think we defined the full flag to be `--id`, not `--identifier` (definition here: https://github.com/kubeshop/tracetest/issues/1937#issuecomment-1458493552). We probablly don't need a short version for the ID

cli/cmd/export_cmd.go
```diff
@@ -0,0 +1,53 @@
...
+			return
+		}
+
+		cmd.Println(pterm.FgGreen.Sprintf(fmt.Sprintf("✔  Definition exported successfully for resource type: %s", resourceType)))
+	},
+	PostRun: teardownCommand,
+}
+
+func init() {
+	exportCmd.PersistentFlags().StringVarP(&exportResourceID, "identifier", "i", "", "id of the resource to export")
```
schoren, line 51: I think we defined the full flag to be `--id`, not `--identifier` (definition here: https://github.com/kubeshop/tracetest/issues/1937#issuecomment-1458493552). We probablly don't need a short version for the ID

cli/cmd/list_cmd.go
```diff
@@ -45,5 +59,9 @@ var listCmd = &cobra.Command{
 }
 
 func init() {
+	listCmd.PersistentFlags().Int32VarP(&listTake, "take", "t", 20, "Take number")
```
schoren, line 62: Again, should these flags be persistent or not? Also, we can probablly skip the short version

cli/formatters/pollingprofile.go
schoren: +1. This is a very hardcoded approach and won't scale to match the dynamic nature of the resource manager

server/model/yaml/encoding.go
```diff
@@ -58,6 +59,13 @@ func Decode(contents []byte) (File, error) {
 			return File{}, fmt.Errorf("cannot decode config: %w", err)
 		}
 		f.Spec = config
+	case FileTypePollingProfile:
```
schoren, line 62: Is this consumed somewhere other than the CLI? If so, let's remove this and try to keep the new things decoupled from this old approach

schoren (changes requested):
Nice work @xoscar!! The issue I see with this PR is that it's duplicating a lof of effor by making the client do all the type conversion, when the functionaliy is provided by the resourcemanager.

Some of the architectural goals of the resource manager are to make clients as thin as possible, and also reduce coupling between packages as much as possible.

server/resourcemanager/encoding.go
```diff
@@ -65,5 +67,9 @@ func encoderFromRequest(r *http.Request) (encoder, error) {
 		}
 	}
 
+	if accept == "" && contentType == "" {
+		return defaultEncoder, nil
+	}
+
```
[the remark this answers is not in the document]
xoscar: Let's talk about the whole content-type header thing, the autogenerated open API client for the CLI is causing some annoying things

cli/cmd/config.go
xoscar: Can I update this in the following PR? it should be about adding support for the demo resource

schoren (approved):
Looking great now. Thanks!

PR #2154, opened 2023-03-13
feat: demo resource

mathnogueira:
This PR adds demos as a resource

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/demo/demoresource/demo_resource.go
```diff
@@ -0,0 +1,344 @@
+package demoresource
```
schoren, line 1: it's not neccesary to have a package for each resource. This could live in `config/demoresource`. Not neccesary but maybe it's not great to have a bunch of almost empty packages? More of a question than a suggestion

schoren (approved):
left a question/pseud-suggestion, but not really critical. other than that looks great

xoscar (approved):
GJ mr Matheus

server/demo/demoresource/demo_resource.go
mathnogueira: makes sense to have it on the config package.

PR #2163, opened 2023-03-14
feat: limit polling profiles to one

mathnogueira:
This PR makes sure that there's only one polling profile for now.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/executor/pollingprofile/polling_profile_resource.go
+++ b/server/executor/pollingprofile/polling_profile_resource.go
@@ -176,7 +176,7 @@ func (r *Repository) Update(ctx context.Context, profile PollingProfile) (Pollin
 	profile.ID = id.ID("current")
 
 	err := r.Delete(ctx, profile.ID)
-	if err != nil {
+	if err != nil && errors.Unwrap(err) != sql.ErrNoRows {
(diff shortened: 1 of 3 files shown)


server/executor/pollingprofile/polling_profile_resource.go
```diff
@@ -40,13 +50,31 @@ type PeriodicPollingConfig struct {
...
+	if _, err := time.ParseDuration(ppc.RetryDelay); err != nil {
+		return fmt.Errorf("retry delay configuration is invalid: %w", err)
+	}
+
+	if _, err := time.ParseDuration(ppc.Timeout); err != nil {
+		return fmt.Errorf("timeout configuration is invalid: %w", err)
+	}
+
+	return nil
+}
```
mathnogueira, line 67: Just noticed we weren't validating if the strings were actually durations.

danielbdias: Question: do we intend to allow users to write durations more relaxedly as we do on assertions, like "1 d" instead of "1d"?

danielbdias (approved):
It's ok! I left one question about the duration validation, but we can tackle it in another PR if needed.

Issue #2125, opened 2023-03-07
[config] Subscribe to analytics enabled config change


jorgeepc:
**AC1**
As a user in the web app,
I want to receive automatic updates if the `analyticsEnabled` config changed in the server,
so the web app can start/stop sending analytics events without the necessity of reloading the page.

**Notes:**
We are going to use the same approach we have for automatic test/transaction runs updates. The required websocket payload is:
```js
{type: 'subscribe', resource: '/app/config/update'}
```

PR #2172, opened 2023-03-15
feature(frontend): adding live updates for analytics changes

xoscar:
This PR adds live updates for the config resource in the front, allowing the FE to switch analytics on an off based on user input

## Changes

- Setting endpoint now listens to config live updates
- FE enables/disabled analytics based on the backend WebSocket

## Fixes

- #2125 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/models/Config.model.ts
+++ b/web/src/models/Config.model.ts
@@ -16,4 +16,18 @@ function Config({
   };
 }
 
+export type TRawLiveConfig = {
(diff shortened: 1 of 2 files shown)


mathnogueira (approved):
that was quicker than I thought it would take to implement this. Good job

PR #2175, opened 2023-03-15
chore(cicd): split RC pipeline

schoren:
This PR splits the RC pipeline so we have more control of what happens for this special case release. In particular at this time we don't want to deploy brew/apt/rpm/helm repositories.

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.github/workflows/release-candidate.yml
+++ b/.github/workflows/release-candidate.yml
@@ -1,0 +1,91 @@
+name: Release Tracetest
+
+permissions:
+  contents: write
+
+concurrency:
+  group: "release-server"
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+on:
+  push:
+    tags:
+      # this pipeline supports RC pre releases
+      - "v[0-9]+.[0-9]+.[0-9]+-rc.[0-9]+"
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-candidate.yml
```diff
@@ -0,0 +1,91 @@
+name: Release Tracetest
```
danielbdias, line 1:
```suggestion
name: Release Tracetest (RC version)
```

danielbdias (approved):

.github/workflows/release-candidate.yml
```diff
@@ -0,0 +1,91 @@
...
+          name: tracetest-web
+          path: web/build/
+
+      # release
+      - uses: goreleaser/goreleaser-action@v4
+        if: steps.cache.outputs.cache-hit != 'true'
+        with:
+          distribution: goreleaser-pro
+          version: v1.15.2
+          args: release --clean --nightly
```
danielbdias, line 86: It makes sense to add a comment here saying that we need the `--nightly` flag to avoid overriding `latest`?

PR #2178, opened 2023-03-15
chore(examples): #2 updating examples to match new config/provisioning

xoscar:
This PR updates the examples folder to match the new config

## Fixes

- (partially) #2177 

## Checklist

- [] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


examples/tracetest-lightstep-otel-demo/tracetest/tracetest-provision.yaml
```diff
@@ -1,2 +1,27 @@
-dataStore:
+---
+type: PollingProfile
+spec:
+  name: Default
+  strategy: periodic
+  default: true
+  periodic:
+    retryDelay: 500ms
+    timeout: 30s
```
danielbdias, line 9: This profile is stricter than the [default profile](https://github.com/kubeshop/tracetest/blob/main/server/executor/pollingprofile/polling_profile_resource.go#L29). It makes sense to use the default instead?

examples/tracetest-lightstep/tracetest/tracetest-provision.yaml
```diff
@@ -1,2 +1,27 @@
-dataStore:
+---
+type: PollingProfile
+spec:
+  name: Default
+  strategy: periodic
+  default: true
+  periodic:
+    retryDelay: 500ms
+    timeout: 30s
```
danielbdias, line 9: Same thing here. This profile is stricter than the [default profile](https://github.com/kubeshop/tracetest/blob/main/server/executor/pollingprofile/polling_profile_resource.go#L29). It makes sense to use the default instead?

PR #2228, opened 2023-03-21
docs: automate CLI reference generation

schoren:
This PR updates the  docs CI pipeline to include the CLI reference doc autogeneration as a step. This way we don't have to remember to manually do this.

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


.github/workflows/docusaurus.yml
```diff
@@ -19,16 +19,37 @@ jobs:
...
+          cache: true
+          cache-dependency-path: go.work
       # Node is required for npm
       - name: Set up Node
         uses: actions/setup-node@v2
         with:
           node-version: "18"
+
+      # build CLI docs
+      - name: Generate tracetest CLI reference docs
```
xoscar, line 37: if we were to have this, does it mean that the docs might be updated with CLI changes even if they are not released?

schoren: this means that every time we release docs, it will include the cli docs that results from building the version for the commit being built

mathnogueira (approved):

.github/workflows/docusaurus.yml
danielbdias: However, these files are not committed in our repo in this way, making the need of creating a PR to update them into the code. If we want to see them locally on docusaurus we need to run `make docgen` on CLI and later run docusarus manually, right?

Thinking about this, does it makes sense to have the autogen file committed in our repo?

PR #2230, opened 2023-03-21
fix: server openapi model generation

schoren:
This PR fixes the makefile target that generates the openapi model for the server, and updates the server with the latest openapi changes.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


Makefile
```diff
@@ -80,7 +80,6 @@ server/openapi: $(OPENAPI_SRC_FILES)
 	$(OPENAPI_GENERATOR_CLI) generate \
 		-i api/openapi.yaml \
 		-g go-server \
-		--global-property apis="api" \
```
xoscar, line 83: @danielbdias I remember you added this to avoid having the controller.go requiring us to add the missing route controllers right?

schoren: this is the cause for the server generation not working. Removing this doesn't break the server

xoscar (approved):

Issue #2040, opened 2023-02-23
[EPIC][Error Handling] Test Run Page Error Handling Improvements


olha23:
As we come across different user sessions, the team has identified multiple areas of opportunity regarding the error handling messaging when executing a test run.

Currently, a test run goes through three significant steps:
1. Trigger execution
2. Trace fetching
3. Test spec execution

Each step has its own set of success and failure scenarios that need to be appropriately displayed to the user.
Today, Tracetest uses only two fields from the test run to validate possible errors.
- `lastErrorState` which contains the string info for the last known error.
- `state` which controls the status of the test.

This was a good starting but now is not sufficient for the clients (CLI/UI) to display enough information so the user can understand how to fix potential problems. As well as providing good user feedback on what is the serverside executing at any given time.

In this case, we have identified a matrix of possible scenarios depending on the test run state, results, and what we should be displaying to the user.

### Test Run Flow Chart

```mermaid
flowchart TD
    A[Run] --> B[Created]
    B --> C[Resolve Trigger Vars]
    C --> D[Execute Trigger]
    D --> ET{Is Successful Trigger}
    ET -->|Yes| E[Queue Polling]
    ET -->|NO| ES[Set State to Failed]
    ES --> Q
    E --> F[Execute Polling Job]
    F --> G[Fetch Trace from Data Store]
    G --> H{Trace Exists}
    H -->|No| I{timed out config reached}
    H -->|Yes| J{Has the span # changed}
    J -->|Yes| G
    J -->|No| K[Trace is ready]
    I -->|No| G
    I -->|Yes| L[Trace fetch failed]
    K --> O[Generating Outputs]
    O --> P[Running Test Specs]
    P --> Q[Finish]
    L --> ES
```

### State Matrix for Test Runs

|             | CREATED       | TRIGGERING                                                                           | CONNECTING_TO_DATA_STORE                                         | POLLING_TRACE                                                                                | GENERATING_OUTPUTS                                        | RUNNING_TEST_SPECS | FINISHED            |
|-------------|---------------|--------------------------------------------------------------------------------------|------------------------------------------------------------------|----------------------------------------------------------------------------------------------|-----------------------------------------------------------|--------------------|---------------------|
| Successful  | Run Page      | Trigger response data  - body - timing - headers                                     | Signal of successful connection to data store                    | Trace                                                                                        | Outputs                                                   | Test Specs results | Trigger/Trace/Test  |
[...]

Issue #2183, opened 2023-03-16
[Error Handling] - Backend - Stopping a Test Run


xoscar:
We need a way for the BE to stop in-progress test runs

### Acceptance Criteria

**AC1**
As a client triggering a test run
I should be able to trigger an endpoint to stop the test run
The polling service should halt the execution,
The test run state should be switched to `STOPPED`
The partial trace (if any) should be stored
And the test specs execution should be skipped

PR #2255, opened 2023-03-27
add endpoint spec to get events from a test run

mathnogueira:
This PR adds the openAPI spec to get events generated from a test run. The model is based on diagram on #2040

```mermaid
classDiagram
    Event <|-- DataStoreConnection
    Event <|-- Polling
    Event <|-- Output
    Event : string type
    Event : enum[trigger, trace, test] stage
    Event : string description
    Event : date created_at
    Event : string test_id
    Event : string run_id

    class DataStoreConnection{
      DataStoreTestConnection info
    }

    class Polling{
      int number_of_spans
      int iteration_number
      string reason_of_next_iteration
      boolean is_complete
    }

    class Output{
      string warning
      string error
      string output_name
    }
```

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


api/testEvents.yaml
```diff
@@ -0,0 +1,56 @@
...
+          $ref: "#/components/schemas/PollingInfo"
+        outputs:
+          type: array
+          items:
+            $ref: "#/components/schemas/OutputInfo"
+
+    PollingInfo:
+      type: object
+      properties:
+        numberSpans:
```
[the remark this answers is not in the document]
mathnogueira: we could add a `type` field and then add objects to represent each type, would that be enough?

api/testEvents.yaml
```diff
@@ -0,0 +1,56 @@
...
+          type: boolean
+
+    OutputInfo:
+      type: object
+      properties:
+        logLevel:
+          type: string
+          enum:
+            - warning
+            - error
```
jorgeepc, line 52: Should we include a `info` type?

api/testEvents.yaml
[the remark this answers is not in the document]
schoren: makes sense, that's the approach we took on all similar situations

api/testEvents.yaml
mathnogueira: I don't think it's worth unless we want to show in the event log what are the values for each output after they were resolved. @xoscar what do you think?

xoscar:
Actually, can we consider adding the specs for this as well? https://github.com/kubeshop/tracetest/issues/2183

api/testEvents.yaml
xoscar: I think the info type is not something we'll be using in the near future so I think its fine to not have it

jorgeepc: Yeah, makes sense @mathnogueira

api/testEvents.yaml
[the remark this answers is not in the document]
mathnogueira: Done

jorgeepc (approved):

mathnogueira:
> Actually, can we consider adding the specs for this as well? #2183

added it

schoren (approved):

danielbdias (approved):

PR #2259, opened 2023-03-28
replace connection test result with structure from model package

mathnogueira:
This PR adds the model for the test run event and replace the connection test structures with the one located in the model package

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/model/test_run_event.go
+++ b/server/model/test_run_event.go
@@ -1,0 +1,49 @@
+package model
+
+import "time"
+
+type (
+	Protocol string
+	Status   string
+)
+
+var (
+	ProtocolHTTP Protocol = "http"
+	ProtocolGRPC Protocol = "grpc"
+)
+
+var (
+	StatusPassed  Status = "passed"
+	StatusWarning Status = "warning"
+	StatusFailed  Status = "failed"
+)
+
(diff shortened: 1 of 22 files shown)


server/model/test_run_event.go
```diff
@@ -0,0 +1,49 @@
...
+var (
+	ProtocolHTTP Protocol = "http"
+	ProtocolGRPC Protocol = "grpc"
+)
+
+var (
+	StatusPassed  Status = "passed"
+	StatusWarning Status = "warning"
+	StatusFailed  Status = "failed"
+)
```
jorgeepc, line 19: Looks good, one minor question, does it make sense to move this to the `connection_test_result.go` model? Or do you think the `test_run_event` model is a better place for this?

jorgeepc (approved):

server/model/test_run_event.go
mathnogueira: Initially, I add them in the `test_run_event`. I moved them from there because it's used by connection package and it doesn't have any relation to the `test_run_event`file. That's why they are not together.

xoscar (approved):
looking gud 🚀

PR #2272, opened 2023-03-28
increase timeout when checking if port is available

mathnogueira:
This PR increases the timeout from 30s to 60s when checking if a port is ready.

This was done because our pipeline would fail quite frequently because the Postgres instance would take some time to start on our cluster. I expect the number of false negative fails to drop after this is merged.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/scripts/wait-for-port.sh
+++ b/scripts/wait-for-port.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 PORT=$1
 
-TIME_OUT=30s
+TIME_OUT=60s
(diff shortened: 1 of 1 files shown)


schoren (changes requested):
This doesn't seem to fix the issue. The problem happens after this script has finished running, so it's not a `not enough wait time` issue. The connection at some point becomes available and then becomes unavailable later

Issue #2268, opened 2023-03-28
Add calls to the Event Publisher on Test spec execution (TEST_SPECS_RUN events)


danielbdias:
On this task, we should handle the following examples:

| Stage | Event Type                | Description                                               | Definition |
|-------|---------------------------|-----------------------------------------------------------|------------|
| Test  | TEST_SPECS_RUN_SUCCESS    | Test Specs were successfully executed                     |            |
| Test  | TEST_SPECS_RUN_ERROR      | Test specs execution error                                |            |
| Test  | TEST_SPECS_RUN_START      | Test specs execution start                                |            |

PR #2280, opened 2023-03-30
Add TestSpec events

danielbdias:
This PR aims to add TestSpec events when executing the assertion engine.

## Fixes

- #2268

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/executor/assertion_runner.go
+++ b/server/executor/assertion_runner.go
@@ -36,6 +36,7 @@ type defaultAssertionRunner struct {
 	inputChannel        chan AssertionRequest
 	exitChannel         chan bool
 	subscriptionManager *subscription.Manager
+	eventEmitter        EventEmitter
 }
 
 var _ WorkerPool = &defaultAssertionRunner{}
@@ -54,6 +55,7 @@ func NewAssertionRunner(
 		assertionExecutor:   assertionExecutor,
 		inputChannel:        make(chan AssertionRequest, 1),
 		subscriptionManager: subscriptionManager,
+		eventEmitter:        eventEmitter,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


server/model/test_run_event.go
```diff
@@ -59,6 +129,60 @@ func (e TestRunEvent) ResourceID() string {
...
+	e.Outputs = outputs
+	return e
+}
+
+func NewTestRunEvent(stage TestRunEventStage, eventType TestRunEventType, testID id.ID, runID int) TestRunEvent {
+	return NewTestRunEventWithArgs(stage, eventType, testID, runID, map[string]string{})
+}
+
+func NewTestRunEventWithArgs(stage TestRunEventStage, eventType TestRunEventType, testID id.ID, runID int, titleAndDescriptionArgs map[string]string) TestRunEvent {
+	eventTypeTable, found := eventTable[stage]
```
xoscar, line 152: Is this for events that have extra information attached to them?

server/executor/assertion_runner.go
```diff
@@ -103,19 +105,26 @@ func (e *defaultAssertionRunner) startWorker() {
 
 func (e *defaultAssertionRunner) runAssertionsAndUpdateResult(ctx context.Context, request AssertionRequest) (model.Run, error) {
 	log.Printf("[AssertionRunner] Test %s Run %d: Starting\n", request.Test.ID, request.Run.ID)
+
+	e.eventEmitter.Emit(ctx, model.NewTestRunEvent(model.StageTest, model.TestEventType_TestSpecsRunStart, request.Test.ID, request.Run.ID))
```
mathnogueira, line 109: There are some pieces of information missing here. An event should also contain a title and a description. Maybe, this is not the best way of emitting an event.

Instead of just declaring the types as constants in a file, I think it would be nice to have a package to centralize the creation of those events, so instead of:

```go
e.eventEmitter.Emit(ctx, model.NewTestRunEvent(model.StageTest, model.TestEventType_TestSpecsRunStart, request.Test.ID, request.Run.ID))
```

We could have:

```go
e.eventEmitter.Emit(ctx, events.TriggerCreated(request.Test.ID, request.Run.ID)
```

Each constructor would have a different set of parameters (testID and RunID are required in all of them) to match the description and title we want to provide.

danielbdias: Make sense! Changing it!

server/model/test_run_event.go
danielbdias: Yes, by following @mathnogueira we have a less complex way to do that, for instance here: https://github.com/kubeshop/tracetest/pull/2280/files#diff-cc7adc54283b5ac608363703f8b26f6b0c084cf35c520e7d06ed20da9dd0fadeR206

mathnogueira (approved):
LGTM

PR #2281, opened 2023-03-31
feat: add trigger resolve and execution events

mathnogueira:
This PR adds events when triggering tests

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/executor/runner.go
+++ b/server/executor/runner.go
@@ -221,7 +221,7 @@ func (r persistentRunner) processExecQueue(job execReq) {
 		}
 	}
 
-	err = r.eventEmitter.Emit(job.ctx, events.TriggerCreatedInfo(job.run.TestID, job.run.ID))
+	err = r.eventEmitter.Emit(job.ctx, events.TriggerExecutionStart(job.run.TestID, job.run.ID))
 	if err != nil {
 		r.handleError(job.run, err)
 	}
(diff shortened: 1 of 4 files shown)


server/executor/runner.go
```diff
@@ -192,6 +221,11 @@ func (r persistentRunner) processExecQueue(job execReq) {
 		}
 	}
 
+	err = r.eventEmitter.Emit(job.ctx, events.TriggerCreatedInfo(job.run.TestID, job.run.ID))
```
danielbdias, line 224: It makes sense to trigger the `TriggerExecutionStart`, right?

```suggestion
	err = r.eventEmitter.Emit(job.ctx, events.TriggerExecutionStart(job.run.TestID, job.run.ID))
```

server/executor/runner.go
```diff
@@ -81,6 +85,12 @@ func (r persistentRunner) handleDBError(run model.Run, err error) {
 	}
 }
 
+func (r persistentRunner) handleError(run model.Run, err error) {
+	if err != nil {
+		fmt.Printf("test %s run #%d trigger DB error: %s\n", run.TestID, run.ID, err.Error())
```
danielbdias, line 90: Philosophical question: should we put this error handling on `EventEmitter` and make `Emit` return nothing?

server/executor/runner.go
```diff
@@ -192,6 +221,11 @@ func (r persistentRunner) processExecQueue(job execReq) {
...
 	}
 
+	err = r.eventEmitter.Emit(job.ctx, events.TriggerCreatedInfo(job.run.TestID, job.run.ID))
+	if err != nil {
+		r.handleError(job.run, err)
+	}
+
 	response, err := triggerer.Trigger(job.ctx, resolvedTest, triggerOptions)
 	run = r.handleExecutionResult(run, response, err)
 	if err != nil {
```
danielbdias, line 231: Should we create a `TriggerExecutionFail` event and send it here?

danielbdias (approved):

server/executor/runner.go
mathnogueira: I prefer to handle it on the runner level. Mainly because this is a temporary thing, we should use a real logger here instead.

Issue #2267, opened 2023-03-28
Add calls to the EventEmitter on Trace fetching (all events, except STOPPED_INFO)


danielbdias:
On this task, we should handle the following examples:

| Stage | Event Type                | Description                                                                               | 
|-------|---------------------------|-------------------------------------------------------------------------------------------|
| Trace | POLLING_START             | Starting the trace polling process                                                        |
| Trace | POLLING_ITERATION_INFO    | A polling iteration has been executed,# of spans - iteration # - reason of next iteration |
| Trace | POLLING_SUCCESS           | The polling strategy has succeeded in fetching the trace from the Data Store              |
| Trace | POLLING_ERROR             | The polling strategy has failed to fetch the trace                                        |
| Trace | FETCHING_START          | Starting the trace fetching process                                       |
| Trace | FETCHING_SUCCESS          | The trace was successfully processed by the backend                                       |
| Trace | FETCHING_ERROR            | The trace was not able to be fetched                                                      |

PR #2291, opened 2023-04-01
Adding PollingEvents

danielbdias:
This PR aims to add polling events to a test run.

## Fixes

- https://github.com/kubeshop/tracetest/issues/2267

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/model/events/events.go
```diff
@@ -158,51 +158,6 @@ func TriggergRPCUnreachableHostError(testID id.ID, runID int, err error) model.T
...
-		Stage:               model.StageTrace,
-		Type:                "QUEUED_INFO",
-		Title:               "Trace Run has been queued to start the fetching process",
-		Description:         "Trace Run has been queued to start the fetching process",
-		CreatedAt:           time.Now(),
-		DataStoreConnection: model.ConnectionResult{},
-		Polling:             model.PollingInfo{},
-		Outputs:             []model.OutputInfo{},
-	}
-}
```
danielbdias, line 189: This event can be captured by the `TracePollingStart`, because they are the same. Thinking about that, I've removed this event and focused on send the `TracePollingStart` event.

server/model/events/events.go
```diff
@@ -158,51 +158,6 @@ func TriggergRPCUnreachableHostError(testID id.ID, runID int, err error) model.T
...
-		Stage:               model.StageTrace,
-		Type:                "DATA_STORE_CONNECTION_INFO",
-		Title:               "A Data store connection request has been executed,test connection result information",
-		Description:         "A Data store connection request has been executed,test connection result information",
-		CreatedAt:           time.Now(),
-		DataStoreConnection: model.ConnectionResult{},
-		Polling:             model.PollingInfo{},
-		Outputs:             []model.OutputInfo{},
-	}
-}
```
danielbdias, line 204: This event is done at the same time we are fetching a trace from the DataStore. Thinking about that, I've focused on emitting the `TraceFetchingStart` event.

api/testEvents.yaml
```diff
@@ -39,8 +39,6 @@ components:
           type: string
           enum:
             - periodic
-        reasonNextIteration:
-          type: string
```
mathnogueira, line 43: Why are we removing this?

server/model/test_run_event.go
```diff
@@ -61,10 +61,9 @@ func (e TestRunEvent) ResourceID() string {
 }
 
 type PollingInfo struct {
-	Type                PollingType
-	ReasonNextIteration string
```
danielbdias, line 65: We need to change this field to the `TracePollingSuccess` event, because we have reason to stop the iterations but not to run a new one.
For instance: we can stop a polling process because the trace hasn't changed for a while or due to a timeout. But we cannot say that about the iteration. Does it make sense?

danielbdias:
> Looks great @danielbdias I just want to understand if we are going to execute a `Data Store Connection Test` at any point (I see you are removing the `TraceDataStoreConnectionInfo` event) Not a blocker for this PR :)

@jorgeepc Thinking for a while and seeing the specs, it makes sense to have a `Data Store Connection Test` at the beginning of the polling process. I've added this event again and added the code to test the connection.

server/executor/poller_executor.go
```diff
@@ -98,17 +102,46 @@ func (pe DefaultPollerExecutor) ExecuteRequest(request *PollingRequest) (bool, m
 		return false, model.Run{}, err
 	}
 
+	if request.IsFirstRequest() {
+		connectionResult := traceDB.TestConnection(request.ctx)
```
danielbdias, line 106: Team, a philosophical question here: it makes sense to stop the polling process here if we receive a connection error on this test?

server/executor/poller_executor.go
```diff
@@ -98,17 +102,46 @@ func (pe DefaultPollerExecutor) ExecuteRequest(request *PollingRequest) (bool, m
...
+	}
+
 	traceID := run.TraceID.String()
 	trace, err := traceDB.GetTraceByID(request.ctx, traceID)
 	if err != nil {
+		anotherErr := pe.eventEmitter.Emit(request.ctx, events.TraceFetchingError(request.test.ID, request.run.ID, err))
+		if anotherErr != nil {
+			log.Printf("[PollerExecutor] Test %s Run %d: failed to emit TraceFetchingError event: error: %s\n", request.test.ID, request.run.ID, anotherErr.Error())
+		}
+
```
mathnogueira, line 126: A would expect a connection test to happen here, so if it fails, we at least know if the data store connection is working. We might want to filter out the `ErrTraceNotFound` error though.

api/testEvents.yaml
mathnogueira: I saw your comment, so you already answered it

server/executor/poller_executor.go
danielbdias: Have you considered this connection a complement of the `TraceFetchingError` event? Something like this:
```go
if err != nil {
		connectionResult := traceDB.TestConnection(request.ctx)

		anotherErr := pe.eventEmitter.Emit(request.ctx, events.TraceFetchingError(request.test.ID, request.run.ID,  connectionResult, err))
		if anotherErr != nil {
			log.Printf("[PollerExecutor] Test %s Run %d: failed to emit TraceFetchingError event: error: %s\n", request.test.ID, request.run.ID, anotherErr.Error())
		}
}
```

mathnogueira (approved):

PR #2292, opened 2023-04-03
feat: make expression package return ErrExpressionResolution when resolution fails

mathnogueira:
This PR makes the expression package return a ErrExpressionResolution when an expression or statement is executed and any resolution step fails and returns some context of why it failed.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/expression/data_store.go
+++ b/server/expression/data_store.go
@@ -1,6 +1,7 @@
 package expression
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 
@@ -21,7 +22,12 @@ func (ds AttributeDataStore) Source() string {
 }
 
 func (ds AttributeDataStore) Get(name string) (string, error) {
-	return ds.Span.Attributes.Get(name), nil
+	value := ds.Span.Attributes.Get(name)
+	if value == "" {
+		return "", errors.New("attribute not found")
+	}
+
+	return value, nil
 }
 
 type MetaAttributesDataStore struct {
@@ -74,5 +80,5 @@ func (ds EnvironmentDataStore) Get(name string) (string, error) {
 		}
 	}
 
-	return "", nil
+	return "", fmt.Errorf("environment variable not found")
 }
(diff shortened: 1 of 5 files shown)


server/expression/data_store.go
```diff
@@ -21,7 +22,12 @@ func (ds AttributeDataStore) Source() string {
 }
 
 func (ds AttributeDataStore) Get(name string) (string, error) {
-	return ds.Span.Attributes.Get(name), nil
+	value := ds.Span.Attributes.Get(name)
+	if value == "" {
+		return "", errors.New("attribute not found")
```
danielbdias, line 27: Does it make sense to type this error and send the attribute name as metadata? To handle later on the events or the API?

Something like this:

```go
// somewhere in the code
type ErrAttributeNoFound struct {
    attributeName string
}

func (e *ErrAttributeNoFound) Error() string {
    return fmt.Sprintf("attribute %s not found", e.attributeName)
} 

func (e *ErrAttributeNoFound) AttributeName() string {
    return e.attributeName
} 

// here
return "", ErrAttributeNoFound{attributeName: name}

```

server/expression/errors.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func resolutionError(typ ResourceType, name string, innerErr error) error {
+	newError := fmt.Errorf(`%w: %s "%s"`, ErrExpressionResolution, typ, name)
+
+	if innerErr != nil {
+		newError = fmt.Errorf("%w: %s", newError, innerErr.Error())
+	}
+
+	return newError
```
danielbdias, line 35: Or typing it here?

danielbdias (approved):

server/expression/data_store.go
mathnogueira: I added part of your idea to the `ResolutionError` object. While reading the errors.Is() source code, I saw that you can declare the `Is` and `Unwrap` functions to a struct and the `errors` package will be able to use those custom-defined functions as part of the standard `Is` and `Unwrap` functions.

PR #2309, opened 2023-04-04
feat: emit assertion errors

mathnogueira:
This PR checks if the assertion engine returns a resolution error, if so, it emits a `TEST_SPECS_ASSERTION_ERROR` with details of which assertion and which span caused the issue.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/assertion_executor.go
```diff
@@ -17,7 +17,73 @@ type AssertionExecutor interface {
 type defaultAssertionExecutor struct{}
 
 func (e defaultAssertionExecutor) Assert(ctx context.Context, defs model.OrderedMap[model.SpanQuery, model.NamedAssertions], trace model.Trace, ds []expression.DataStore) (model.OrderedMap[model.SpanQuery, []model.AssertionResult], bool) {
-	return assertions.Assert(defs, trace, ds)
+	testResult := model.OrderedMap[model.SpanQuery, []model.AssertionResult]{}
```
mathnogueira, line 20: I basically moved the content from the `assertions` package to here.

danielbdias (approved):

Issue #2285, opened 2023-03-31
Differentiate between a trigger failed state vs a trace failed state in a test run


jorgeepc:
Our current test run `state` implementation only supports one single failed state (`"FAILED"`). We need a way to differentiate between a `trigger failed state` vs a `trace failed state`.

**AC1**
As a user running a test that failed in the trigger stage,
I should receive a `"TRIGGER_FAILED"` state,
so I can understand that the trigger stage has a problem.

**AC2**
As a user running a test that failed in the trace stage,
I should receive a `"TRACE_FAILED"` state,
so I can understand that the trace stage has a problem.

PR #2310, opened 2023-04-04
feat(sever): for test runs, split the failed state into more descriptive statuses

schoren:
This PR makes the TestRun state more accurate by providing a new set of statuses: `TRACE_FAILED`, `TRIGGER_FAILED` and `ASSERTION_FAILED`.

## Changes

- add new states and implement where applicable

## Fixes

- #2285

## Checklist

- [X] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/model/tests.go
+++ b/server/model/tests.go
@@ -180,13 +180,20 @@ const (
 	RunStateCreated             RunState = "CREATED"
 	RunStateExecuting           RunState = "EXECUTING"
 	RunStateAwaitingTrace       RunState = "AWAITING_TRACE"
-	RunStateFailed              RunState = "FAILED"
+	RunStateTriggerFailed       RunState = "TRIGGER_FAILED"
+	RunStateTraceFailed         RunState = "TRACE_FAILED"
+	RunStateAssertionFailed     RunState = "ASSERTION_FAILED"
 	RunStateFinished            RunState = "FINISHED"
 	RunStateAwaitingTestResults RunState = "AWAITING_TEST_RESULTS"
 )
 
 func (rs RunState) IsFinal() bool {
-	return rs == RunStateFailed || rs == RunStateFinished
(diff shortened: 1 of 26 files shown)


mathnogueira (approved):
Looks good!

jorgeepc (commented):
Looks good. One question: with this approach, is it safe to assume that if I get a `TRACE_FAILED` state, the trigger stage was correct?

mathnogueira:
> Looks good. One question: with this approach, is it safe to assume that if I get a `TRACE_FAILED` state, the trigger stage was correct?

yes. If the trigger fails, the test execution stops

schoren:
@jorgeepc exactly. The failed stage means that all the previous steps succedded

jorgeepc (commented):
I think we need to update the OpenAPI specs.

schoren:
@jorgeepc good catch. updated

server/model/tests.go
```diff
@@ -180,13 +180,20 @@ const (
 	RunStateCreated             RunState = "CREATED"
 	RunStateExecuting           RunState = "EXECUTING"
 	RunStateAwaitingTrace       RunState = "AWAITING_TRACE"
-	RunStateFailed              RunState = "FAILED"
```
danielbdias, line 183: Today we save this value on our database, right? How will we keep the retro compatibility with older Tracetest versions?
Should we map the old `FAILED` string as always `ASSERTION_FAILED`?

danielbdias: Or even writing a migration to fix it on the database?

danielbdias (commented):
It sounds ok! Before merging it I have one comment about retro compatibility: should we write a migration or handle in separate the old status to avoid problems with early versions of Tracetest?

PR #2325, opened 2023-04-06
feature(backend): add temporary buffer to subscription manager

mathnogueira:
This PR adds a temporary buffer to the subscription manager. This makes sure that if someone subscribes to a topic after an event is received, it will still receive those events.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/subscription/buffer.go
+++ b/server/subscription/buffer.go
@@ -1,0 +1,114 @@
+package subscription
+
+import (
+	"sync"
+	"time"
+)
+
+type TemporaryBuffer struct {
+	messages             []Message
+	lastMessageTimestamp time.Time
+}
+
+func NewTemporaryBuffer() *TemporaryBuffer {
+	return &TemporaryBuffer{
+		messages:             make([]Message, 0),
+		lastMessageTimestamp: time.Now(),
+	}
+}
+
+func (b *TemporaryBuffer) AddMessage(message Message) {
+	b.messages = append(b.messages, message)
+	b.lastMessageTimestamp = time.Now()
+}
+
+func (b *TemporaryBuffer) GetMessages() []Message {
+	return b.messages
+}
+
+func (b *TemporaryBuffer) CleanUp() {
+	if len(b.messages) > 0 && time.Since(b.lastMessageTimestamp) > 10*time.Second {
+		b.messages = make([]Message, 0)
+		b.lastMessageTimestamp = time.Now()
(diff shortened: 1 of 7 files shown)


server/subscription/buffer.go
```diff
@@ -0,0 +1,114 @@
...
+	b.messages = append(b.messages, message)
+	b.lastMessageTimestamp = time.Now()
+}
+
+func (b *TemporaryBuffer) GetMessages() []Message {
+	return b.messages
+}
+
+func (b *TemporaryBuffer) CleanUp() {
+	if len(b.messages) > 0 && time.Since(b.lastMessageTimestamp) > 10*time.Second {
```
danielbdias, line 30: Can we move this magic number `10 * time.Second` to a constant?

server/subscription/buffer.go
```diff
@@ -0,0 +1,114 @@
+package subscription
```
danielbdias, line 1: Brainstorm: does it make sense in the future to start to using queue engines to deal with event messaging?
I was thinking if we could use something like https://github.com/adjust/rmq or similar.

server/subscription/manager.go
```diff
@@ -1,67 +1,45 @@
...
+	cleanupPeriod time.Duration
 }
 
 func NewManager() *Manager {
-	return &Manager{
-		subscriptions: make(map[string][]Subscriber),
-		mutex:         sync.Mutex{},
+	manager := &Manager{
+		subscriptions: NewBufferedSubscriptions(),
+		cleanupPeriod: 10 * time.Second,
```
danielbdias, line 13: Can we move this config to constant?

danielbdias (approved):

jorgeepc (commented):
Tested `tests` and `transactions` (they use the web socket too). Noticed a couple of things:
- It seems that the `clean up` process is not correct. I keep getting all the messages in the web socket even if I refresh the page.
- Transactions are failing in my local env. If I run a transaction the server stops sending web socket updates and the transactions stays in the `Executing` state forever.

server/subscription/buffer.go
mathnogueira: I think it makes sense to close this PR and to work with some queue engine.

danielbdias: ok!

Issue #2174, opened 2023-03-15
Add test snippets user can use


olha23:
We need  to add key assertions that provide value out-of-the-box.
Possible scenario:
Multiple of the examples Postman has would apply:
Response status code is 200
Response body contains string
Response body JSON value check
Response time is less than 200ms
but there are additional ones we could show:
all database spans return in less than 100ms
all http spans return status code of 200
all gRPC spans return status code of 0

xoscar:
This is still future work, right? if there is anything we can do to speed this up let us know 😄

olha23:
@xoscar @jorgeepc need you feedback on this https://www.figma.com/file/LBN4SKVPq3ykegrPKbHT2Y/0.8-0.9-Release-Tracetest?node-id=2163%3A33256&t=P9kNAyXKV3bbliy7-1

kdhamric:
Copying @xoscar questions from Figma here so we can discuss them:

Should the examples match the spans that exist in the trace? In the first iteration, lets just keep it simple and have a defined list and have all of them show.
Should we breakdown the examples by types? db/http/faas, etc? I don't think we need to
Should we have a list of examples by type documented somewhere? Perhaps, but I don't think we need to include that in this scope.

xoscar:
@kdhamric

PR #2366, opened 2023-04-11
feature(frontend): adding test spec snippets

xoscar:
This PR adds a way for users to select a predefined test spec snippet so they have a quick start when adding checks for a trace

## Changes

- Updating add test spec button
- Adding source for snippets

## Fixes

- #2174 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/dee7d821ab7d4bb1b4f53d81659e02c3

### DIFF
--- a/web/src/components/TestResults/AddTestSpecButton.tsx
+++ b/web/src/components/TestResults/AddTestSpecButton.tsx
@@ -1,18 +1,20 @@
-import React, {useCallback, useMemo} from 'react';
+import React, {useCallback, useEffect, useMemo, useRef} from 'react';
 import {CaretDownOutlined} from '@ant-design/icons';
 import {Dropdown, Menu} from 'antd';
 import SpanService from 'services/Span.service';
 import Span from 'models/Span.model';
+import {TEST_SPEC_SNIPPETS, TSnippet} from 'constants/TestSpecs.constants';
 import * as S from './TestResults.styled';
 import {useTestSpecForm} from '../TestSpecForm/TestSpecForm.provider';
-import {TEST_SPEC_SNIPPETS, TSnippet} from '../../constants/TestSpecs.constants';
 
 interface IProps {
   selectedSpan: Span;
+  visibleByDefault?: boolean;
 }
 
-const AddTestSpecButton = ({selectedSpan}: IProps) => {
+const AddTestSpecButton = ({selectedSpan, visibleByDefault = false}: IProps) => {
   const {open} = useTestSpecForm();
+  const caretRef = useRef<HTMLElement>(null);
   const handleEmptyTestSpec = useCallback(() => {
     const selector = SpanService.getSelectorInformation(selectedSpan);
 
@@ -36,6 +38,12 @@ const AddTestSpecButton = ({selectedSpan}: IProps) => {
     [open]
   );
 
+  useEffect(() => {
+    if (visibleByDefault && caretRef.current) {
+      caretRef.current?.click();
+    }
+  }, [visibleByDefault]);
+
   const menu = useMemo(
     () => (
       <Menu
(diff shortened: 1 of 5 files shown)


web/src/constants/TestSpecs.constants.ts
```diff
@@ -0,0 +1,45 @@
+import {IValues} from 'components/TestSpecForm/TestSpecForm';
+
+export type TSnippet = Required<IValues>;
+
+export const HTTP_SPANS_STATUS_CODE: TSnippet = {
```
xoscar, line 5: Here's where we can continue adding snippets, you can add it manually using the UI and then copy the details in this file @kdhamric

web/src/constants/TestSpecs.constants.ts
```diff
@@ -0,0 +1,45 @@
...
+  assertions: [
+    {
+      left: 'attr:tracetest.span.duration',
+      comparator: '<',
+      right: '100ms',
+    },
+  ],
+};
+
+export const TEST_SPEC_SNIPPETS: TSnippet[] = [
```
xoscar, line 41: Dont forget to add it to this array

web/src/components/TestResults/AddTestSpecButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+        items={[
+          {
+            label: 'Try these snippets for quick testing:',
+            key: 'test',
+            type: 'group',
+            children: TEST_SPEC_SNIPPETS.map(snippet => ({
+              label: snippet.name,
+              key: snippet.name,
+              onClick: () => onSnippetClick(snippet),
+            })),
```
jorgeepc, line 51: I was thinking if we should include an option to create an empty test spec. I know you can click the left part of the button but maybe users won't get it.

Screenshot 2023-04-11 at 14 08 33

jorgeepc (approved):
🔥 Looks great @xoscar Added one question, please take a look.

web/src/components/TestResults/AddTestSpecButton.tsx
xoscar: Would it behave differently than clicking the left button? I think we can gather some feedback on the current UX and then we can think of updating it.

We also need @kdhamric feedback on what other snippets he'd like to add

jorgeepc: It should have the same behavior as clicking the left button. It could be the last item in the list (with a visual separator).

danielbdias (approved):

web/src/components/TestResults/AddTestSpecButton.tsx
kdhamric: Here are 3 more 'prompted test specs'

export const TRIGGER_SPAN_RESPONSE_BODY_CONTAINS: TSnippet = {
    name: 'Trigger Span: Response body contains "this string"',
    selector: 'span[tracetest.span.type="general" name="Tracetest trigger"]',
    assertions: [
      {
        left: 'attr:tracetest.response.body',
        comparator: 'contains',
        right: '"this string"',
      },
    ],
  };


  export const GRPC_SPANS_STATUS_CODE: TSnippet = {
    name: 'All gRPC Spans: Status is Ok',
    selector: 'span[tracetest.span.type="rpc" rpc.system="grpc"]',
    assertions: [
      {
        left: 'attr:grpc.status_code',
        comparator: '=',
        right: '0',
      },
    ],
  };


  export const DB_SPANS_QUALITY_DB_STATEMENT_PRESENT: TSnippet = {
    name: 'All Database Spans: db.statement should always be defined (QUALITY)',
    selector: 'span[tracetest.span.type="database"]',
    assertions: [
      {
        left: 'attr:db.system',
        comparator: '!=',
        right: '""',
      },
    ],
  };

PR #2367, opened 2023-04-11
feat: allow users to stop a running test

schoren:
This PR implements the stop running request feature, already mocked.

## Fixes

- #2183

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


api/openapi.yaml
```diff
@@ -825,7 +825,7 @@ paths:
             application/yaml:
               schema:
                 type: string
-  /test/{testId}/run/{runId}/stop:
+  /tests/{testId}/run/{runId}/stop:
```
schoren, line 828: all test endpoints are actually `/tests` with a trailing `s`

local-config/collector.config.yaml
```diff
@@ -13,7 +13,7 @@ exporters:
     loglevel: debug
 
   otlp/1:
-    endpoint: host.docker.internal:21321
+    endpoint: tracetest:21321
```
schoren, line 16: this was probablly a leftover from someone doing docker tests, it should point to the internal url

local-config/tracetest.provision.yaml
```diff
@@ -4,3 +4,12 @@ spec:
   name: OpenTelemetry Collector
   type: otlp
   isdefault: true
+---
+type: Demo
```
schoren, line 8: addin a default demo locally eased the testing process

server/executor/runner.go
```diff
@@ -75,12 +75,11 @@ type persistentRunner struct {
 }
 
 type execReq struct {
-	ctx                 context.Context
-	test                model.Test
-	run                 model.Run
-	subscriptionManager *subscription.Manager
```
schoren, line 81: this var was unused

server/executor/runner.go
```diff
@@ -132,14 +131,18 @@ func getNewCtx(ctx context.Context) context.Context {
 }
 
 func (r persistentRunner) Run(ctx context.Context, test model.Test, metadata model.RunMetadata, environment model.Environment) model.Run {
-	ctx = getNewCtx(ctx)
+	ctx, cancelCtx := context.WithCancel(
```
schoren, line 134: we want to make the context cancellable, since it isn't by default

server/executor/trace_poller.go
```diff
@@ -132,6 +132,12 @@ func (tp tracePoller) enqueueJob(job PollingRequest) {
 }
 
 func (tp tracePoller) processJob(job PollingRequest) {
+	select {
+	default:
+	case <-job.ctx.Done():
+		fmt.Println("trace poller. Context cancelled.")
+	}
+
```
schoren, line 140: Handle context cancelation

server/executor/trace_poller.go
```diff
@@ -190,7 +196,13 @@ func (tp tracePoller) runAssertions(job PollingRequest) {
 func (tp tracePoller) handleTraceDBError(job PollingRequest, err error) (bool, string) {
 	run := job.run
 
-	pp := *tp.ppGetter.GetDefault(job.ctx).Periodic
+	profile := tp.ppGetter.GetDefault(job.ctx)
```
schoren, line 199: sometimes the context can get cancelled in the middle of this func call, causing a nil pointer panic. This new approach handles the case more gracefully

schoren:
@jorgeepc fixed!

local-config/collector.config.yaml
danielbdias: Question: is this file intended just for local testing, or do we use it in some pipelines?

schoren: local testing. It's used when you run the `./run.sh` script for example, that is mainly targetted to run an env similar to CI, so you can run the tracetesting tests locally easily.

server/executor/run_stop.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (r persistentRunner) listenForStopRequests(ctx context.Context, cancelCtx context.CancelFunc, run model.Run) {
+	sfn := subscription.NewSubscriberFunction(func(m subscription.Message) error {
+		stopRequest, ok := m.Content.(StopRequest)
+		if !ok {
+			return nil
+		}
+
+		ctx, _ := r.tracer.Start(ctx, "User Requested Stop Run")
+		// refresh data from DB to make sure we have the altest possible data before updating
```
danielbdias, line 31: Minor nit:
```suggestion
		// refresh data from DB to make sure we have the latest possible data before updating
```

server/executor/run_stop.go
```diff
@@ -0,0 +1,54 @@
...
+		run, err := r.runs.GetRun(ctx, stopRequest.TestID, stopRequest.RunID)
+		if err != nil {
+			log.Printf("[TracePoller] Test %s Run %d: fail to emit TracePollingStart event: %s \n", stopRequest.TestID, stopRequest.RunID, err.Error())
```
danielbdias, line 34:
```suggestion
			log.Printf("[TracePoller] Test %s Run %d: fail to get test run data: %s \n", stopRequest.TestID, stopRequest.RunID, err.Error())
```

server/executor/runner.go
danielbdias, line 1: Something to think about in another PR: is there a way to add a unit test that checks if the runner behaved correctly on a cancel case?

danielbdias (approved):

server/subscription/manager.go
```diff
@@ -4,13 +4,13 @@ import "sync"
 
 type Manager struct {
 	subscriptions map[string][]Subscriber
-	mutex         sync.Mutex
+	mutex         sync.RWMutex
```
schoren, line 7: All changes in this file are to fix the race condition making some tests fail

server/executor/eventemitter.go
```diff
@@ -34,9 +32,6 @@ func (em *internalEventEmitter) Emit(ctx context.Context, event model.TestRunEve
 		return err
 	}
 
-	em.mutex.Lock()
-	defer em.mutex.Unlock()
-
```
schoren, line 39: this lock is moved to the publisher now

server/executor/transaction_runner_test.go
```diff
@@ -32,26 +32,27 @@ func (r *fakeTestRunner) Run(ctx context.Context, test model.Test, metadata mode
 	}
 
 	go func() {
+		run := newRun                      // make a local copy to avoid race conditions
```
schoren, line 35: by using a copy instead of the original var, we avoid race conditions

server/executor/transaction_runner_test.go
```diff
@@ -159,20 +160,26 @@ func runTransactionRunnerTest(t *testing.T, withErrors bool, assert func(t *test
...
-			}
-
-			return nil
-		}),
-	)
-	// TODO: this will block indefinitely. we need to set a timeout or something
-	<-done
-
-	assert(t, transactionRun)
+	done := make(chan model.TransactionRun, 1)
```
schoren, line 163: instead of using the channel as a `done` signal, we use it to return the relevant value. by not sharing a variable between 2 goroutines, we eliminate a race condition

server/executor/transaction_runner_test.go
```diff
@@ -159,20 +160,26 @@ func runTransactionRunnerTest(t *testing.T, withErrors bool, assert func(t *test
...
+		return nil
+	})
+	subscriptionManager.Subscribe(transactionRun.ResourceID(), sf)
+
+	var finalRun model.TransactionRun
+	select {
+	case finalRun = <-done:
+		subscriptionManager.Unsubscribe(transactionRun.ResourceID(), sf.ID()) //cleanup to avoid race conditions
+		fmt.Println("run completed")
+	case <-time.After(10 * time.Second):
```
schoren, line 179: also it allows to set a timeout instead of waiting forever

server/subscription/manager.go
```diff
@@ -48,6 +48,8 @@ func (m *Manager) Unsubscribe(resourceID string, subscriptionID string) {
 }
 
 func (m *Manager) PublishUpdate(message Message) {
+	m.mutex.RLock()
+	defer m.mutex.RUnlock()
```
danielbdias, line 52: ❤️

Issue #2334, opened 2023-04-10
[CLI Improvements] Migrate DataStores to resource manager architecture


danielbdias:
To adapt the DataStore resource to the resource manager architecture, we should: 

1. Update the OpenAPI specs considering the new endpoints
2. Create new endpoints using the Resource Manager structure on the Backend
3. Update the code that uses the resource on the backend to use the new resource manager repository
4. Adapt the CLI to have commands that use this new structure and make the old ones deprecated
5. Adapt the frontend calls to consider the new endpoints
6. Update the documentation related to this resource

Our reference should be the [`DataStore` specification](https://github.com/kubeshop/tracetest/issues/2182#issuecomment-1490732763) on #2182.

### Tasks to do

- [x] #2343
- [x] #2341
- [x] #2342
- [x] #2344
- [x] #2345

danielbdias:
Added branch [feature/datastore-as-resource](https://github.com/kubeshop/tracetest/tree/feature/datastore-as-resource) to control this feature.

PR #2368, opened 2023-04-11
Updating DataStore API definition to consider ResourcesAPI structure

danielbdias:
This PR updates the OpenAPI definition to consider the DataStore API on the resources API architecture.

**Note:** We are only running `make generate` to change the types on the backend PRs (#2385), just to avoid breaking things on this PR.

**Loom:** https://www.loom.com/share/0870ffdc991b4f37b8e35faf281a6605

## Changes

- Update `api/openapi.yaml` definitions

## Fixes

- https://github.com/kubeshop/tracetest/issues/2343
- (partially) https://github.com/kubeshop/tracetest/issues/2334

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1429,89 +1429,17 @@ paths:
           description: "problem deleting a demo"
 
   # Data Stores
-  /datastores:
-    get:
-      tags:
-        - api
-      summary: "Get all Data Stores"
-      description: "Get all Data Stores"
-      operationId: getDataStores
-      parameters:
-        - in: query
-          name: take
-          description: "indicates how many data stores can be returned by each page"
-          schema:
-            type: integer
-            default: 20
-        - in: query
-          name: skip
-          description: "indicates how many data stores will be skipped when paginating"
-          schema:
-            type: integer
-            default: 0
-        - in: query
-          name: query
-          description: "query to search data stores, based on data store name"
-          schema:
-            type: string
-        - in: query
-          name: sortBy
-          description: "indicates the sort field for the data stores"
-          schema:
-            type: string
-            enum: [created, name]
-        - in: query
-          name: sortDirection
-          description: "indicates the sort direction for the data stores"
-          schema:
-            type: string
-            enum: [asc, desc]
-      responses:
-        200:
-          description: successful operation
-          headers:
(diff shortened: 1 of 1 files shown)


api/openapi.yaml
```diff
@@ -1449,17 +1449,12 @@ paths:
           schema:
             type: integer
             default: 0
-        - in: query
```
danielbdias, line 1452: Right now, we are not explicitly putting a `query` parameter on our endpoints because they are injected as it is on the queries, making our endpoints unsafe. Later we need to improve this.

api/openapi.yaml
```diff
@@ -1429,10 +1429,10 @@ paths:
           description: "problem deleting a demo"
 
   # Data Stores
-  /datastores:
+  /datastore:
```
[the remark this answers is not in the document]
schoren: so if we add a plural form, it would leave the api almost unchanged, right? it's not hard to add that to the resource manager, do you think that would work?

api/openapi.yaml
schoren: IMO we should put it here and fix the implementation

api/openapi.yaml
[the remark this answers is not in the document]
danielbdias: It makes sense. I'm finishing the first draft of the DataStore, and I'll add this functionality.

api/openapi.yaml
danielbdias: Agreed. We can open a PR to solve that.

api/openapi.yaml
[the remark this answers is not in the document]
danielbdias: Issue created: https://github.com/kubeshop/tracetest/issues/2387

api/openapi.yaml
danielbdias: Issue created: https://github.com/kubeshop/tracetest/issues/2388

kdhamric:
Is there anything to 'show' out of this via a Loom video?

danielbdias:
@kdhamric it is possible! I've updated the PR description with a Loom video.

Issue #2173, opened 2023-03-15
Fix navigation logic


olha23:
Check the video please:
https://www.loom.com/share/3101ee5311a344aa9818e169524545da

xoscar:
@jorgeepc what do you think?

olha23:
@xoscar not sure that i understand your comment. the idea was to make the back arrow behave as a regular back action

jorgeepc:
Ok, so the main issue @olha23  is trying to solve is: 
- Persist the paginated list "page" when the user navigates back.
(I think it's doable and provides value)

Now, I agree with @xoscar  about the strange behavior related to having the `Test Detail` page in the middle.

|   |  |
| ------------- | ------------- |
| <img width="1624" alt="Screenshot 2023-03-23 at 11 08 22" src="https://user-images.githubusercontent.com/3879892/227266111-a549a6c9-a150-42d4-ab87-e07d5141c2c7.png">  | <img width="1624" alt="Screenshot 2023-03-23 at 11 08 26" src="https://user-images.githubusercontent.com/3879892/227266137-6ffa62db-47ed-406b-ab1a-9755aa0feca2.png"> |

A couple of thoughts:

- Do we really need to implement a back button in the app? If you think about it, the back button in the `Test Run` page it's not even doing a back action, it's redirecting the user to a totally undesired page. IMO we should remove the back button there and offer the user the option to navigate to the `Test Detail` page using a link from the `Test Run` page.
- If we go with the path of removing all the back buttons we have in the app, we could rely on the native "back" functionality implemented in browsers.

xoscar:
I agree with @jorgeepc on removing the back buttons to relly on the browser's default behavior. What do you think @olha23 ?

olha23:
I’m not sure about this browser idea, because how we can be ensure the browser's integrated Back and Forward buttons always work for your website?
I prefer to leave our back button, but to use the right direction of transition for parent and child objects.  For example, if user open a specific version of test from test detail page we should back user on this page, when user open a test from main page we should back user on the main page when he click on back butoon

xoscar:
- Custom back button should behave the same way as the browsers
- Store the page number as part of the query parameters so users can stay at the current page

PR #2372, opened 2023-04-12
feature(frontend): adding persistent page number for list

xoscar:
This PR updates the behavior of the navigation logic when navigating across different pages, it also keeps track of the page number the list was for so users can come back to what exactly they were looking at.

## Changes

- Updates back buttons for headers to use the `goback` method
- Pagination component now updates the search params on page change
- `usePagination` hook now checks for a `defaultPage` from the search params

## Fixes

- #2173 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/7fd3bfc0781d4bfb9e9102dc4e47e0b9


jorgeepc (approved):
💣

Issue #2245, opened 2023-03-24
[Error Handling] Analytics Events Update


xoscar:
### Acceptance Criteria

**AC1**
As a user looking at the segment events
I should be able to see the conversion rate from
Running a test to successfully receive a trigger response

**AC2**
As a user looking at the segment events
I should be able to see the conversion rate from
Running a test to successfully receive a trace associated with it

**AC3**
As a user looking at the segment events
I should be able to see the conversion rate from
Running a test to successfully executing the associated test specs (pass or fail not important... just 'engine worked')

Trying to see the dropoff at these steps so we can know where to improve:

Test launched
    --> Triggered properly? (yes or no)
           -> Trace gathered properly?   (yes or no)
                  -> (maybe) Assertions executed properly?  (yes or no) (Yes means test 'worked')

Suggestion (per @jorgeepc) is to emit one event when the testRun process finish. And add the run state in the payload:
TRIGGER_FAILED
TRACE_FAILED
ASSERTION_FAILED
FINISHED

Name: test_finished
When: After a test finishes processing and moves to a final state where it will not be processed further
Where: backend
Data: finalState: { TRIGGER_FAILED | TRACE_FAILED | ASSERTION_FAILED | FINISHED }

PR #2375, opened 2023-04-12
feature(backend): adding analytics events for test run checkpoints

xoscar:
This PR adds analytics events for the checkpoints that end a test run execution

## Changes

- Adding analytics events at the different levels where a test run ends

## Fixes

- #2245 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/executor/assertion_runner.go
+++ b/server/executor/assertion_runner.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/kubeshop/tracetest/server/analytics"
 	"github.com/kubeshop/tracetest/server/expression"
 	"github.com/kubeshop/tracetest/server/model"
 	"github.com/kubeshop/tracetest/server/model/events"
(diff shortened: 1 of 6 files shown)


server/executor/assertion_runner.go
```diff
@@ -122,7 +123,12 @@ func (e *defaultAssertionRunner) runAssertionsAndUpdateResult(ctx context.Contex
 			log.Printf("[AssertionRunner] Test %s Run %d: fail to emit TestSpecsRunError event: %s\n", request.Test.ID, request.Run.ID, anotherErr.Error())
 		}
 
-		return model.Run{}, e.updater.Update(ctx, run.AssertionFailed(err))
+		run = run.AssertionFailed(err)
```
xoscar, line 126: I was trying to find a single place where we finished a test run no matter the outcome, but I didn't find it, it feels that wherever the error is found we are just updating the required fields.

I know that the `run.go` model has a `Finish` method but it didn't feel right to execute an analytics event from a model. 

We need to consider having a single spot to finish a test run that is not the model where we could execute side effects in the future, and not having to scatter the logic across these files.

danielbdias (approved):

jorgeepc (approved):

PR #2382, opened 2023-04-13
chore: reuse parameters to make the openapi file smaller

mathnogueira:
This PR reduces the size of the openapi file by reusing the common parameters when defining endpoints

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


api/openapi.yaml
```diff
@@ -66,35 +66,11 @@ paths:
...
-        - in: query
-          name: sortDirection
-          description: "indicates the sort direction for the transactions"
-          schema:
-            type: string
-            enum: [asc, desc]
+        - $ref: "./parameters.yaml#/components/parameters/take"
+        - $ref: "./parameters.yaml#/components/parameters/skip"
+        - $ref: "./parameters.yaml#/components/parameters/query"
+        - $ref: "./parameters.yaml#/components/parameters/runnableResourceSortBy"
```
danielbdias, line 72: Question: is there a way to "override" ref attributes? I believe that we have many similar `sortBy` fields with only the `enum` with different values.

api/openapi.yaml
```diff
@@ -1062,35 +803,11 @@ paths:
...
-        - in: query
-          name: sortDirection
-          description: "indicates the sort direction for the transactions"
-          schema:
-            type: string
-            enum: [asc, desc]
+        - $ref: "./parameters.yaml#/components/parameters/take"
+        - $ref: "./parameters.yaml#/components/parameters/skip"
+        - $ref: "./parameters.yaml#/components/parameters/query"
+        - $ref: "./parameters.yaml#/components/parameters/sortBy"
```
danielbdias, line 809: Here it should be a sortableSortBy, right?

danielbdias (approved):

api/openapi.yaml
mathnogueira: good catch

Issue #2374, opened 2023-04-12
Need input validation on url field of http request trigger


kdhamric:
AC1: As a user adding or editing a test
I should not be allowed to enter a url that is not valid
i.e. it should have the protocol and a validly structured url
or we should give an input verification error.

AC2:
We should show an example of the url in the URL field in the grey text.
ie change: 
Enter request url 
to 
Enter request url, such as http://yourhost/path

AC3:
In the red 'help text' on validation failure, show 'Please enter a request url, such as http://yourhost/path'

xoscar:
Issue was fixed for urls like:

```
www.google.com
amazon.com.mx
github.com
demo-pokemon-api.demo.svc.cluster.local/pokemon/import
```

But for others when specifying the port like:

```
demo-api:8081/pokemon
```

it doesn't work as the golang library takes `demo-api` as the Scheme...
![Image](https://user-images.githubusercontent.com/11051031/233164316-6eccc79c-9478-4e9d-8ab0-58965af20a3c.png)

PR #2384, opened 2023-04-13
feature(backend): defaulting url trigger to http if no scheme is present

xoscar:
This PR adds the `http` scheme for URLs that doesn't have it already

## Changes

- Updates http trigger to add scheme if not present

## Fixes

- #2374 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/e435543d25e741089e5b7f70fa123228


server/executor/trigger/http.go
```diff
@@ -81,7 +82,17 @@ func (te *httpTriggerer) Trigger(ctx context.Context, test model.Test, opts *Tri
...
 	}
-	req, err := http.NewRequestWithContext(ctx, strings.ToUpper(string(tReq.Method)), tReq.URL, body)
+
+	parsedUrl, err := url.Parse(tReq.URL)
+	if err != nil {
+		return response, err
+	}
+
+	if parsedUrl.Scheme == "" {
+		parsedUrl.Scheme = "http"
```
jorgeepc, line 92: Should we have a constant for this?

jorgeepc (approved):

server/executor/trigger/http.go
```diff
@@ -81,7 +82,17 @@ func (te *httpTriggerer) Trigger(ctx context.Context, test model.Test, opts *Tri
...
 		body = bytes.NewBufferString(tReq.Body)
 	}
-	req, err := http.NewRequestWithContext(ctx, strings.ToUpper(string(tReq.Method)), tReq.URL, body)
+
+	parsedUrl, err := url.Parse(tReq.URL)
+	if err != nil {
+		return response, err
+	}
+
+	if parsedUrl.Scheme == "" {
```
danielbdias, line 91: We need to test an edge case: if the address comes with a `:443`, should we change this scheme to `https`?

server/executor/trigger/http.go
schoren: I don't think a constant is needed here. it's not a value that will be referenced any other place, it's just a default used in this only context.

schoren (approved):

server/executor/trigger/http.go
xoscar: I think that's just a really weird use case, here's what postman shows
Screenshot 2023-04-14 at 10 03 19

danielbdias: In that case, we can live with that situation.

danielbdias (approved):

PR #2385, opened 2023-04-13
Add Data store as resource on backend

danielbdias:
This PR adds the Data Store as a resource into the Tracetest server.

On this PR, both the old and new structures coexist on the API. The next PR will update the OpenAPI code with make generate and change the references to use this structure.

Loom: https://www.loom.com/share/45b49110030d4b0688b51c059f97388c

## Changes

- Add DataStore resource package
- Updated internal references to use this package
- Removed DataStore model

## Fixes

- https://github.com/kubeshop/tracetest/issues/2341
- (partially) https://github.com/kubeshop/tracetest/issues/2334

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/tracedb/datastoreresource/_tests.go
schoren, line 1: this looks like a weird name. is it like this for any particular reason?

server/tracedb/datastoreresource/repository.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func (r *Repository) Update(ctx context.Context, dataStore DataStore) (DataStore, error) {
+	// enforce ID and default
+	dataStore.ID = "current"
+	dataStore.Default = true
+
+	// reuse the created_at field for auditing purposes,
+	// unless the client explicitly send it
+	createdAt, err := r.getCreatedAt(ctx, dataStore)
```
schoren, line 73: do we need to support this case? I think unless we have an explicit need for this, clients should not be allowed to set this

server/model/data_stores.go
```diff
@@ -111,6 +111,7 @@ func (ds DataStore) HasID() bool {
 }
 
 func (ds DataStore) Validate() error {
+	//TODO: replace for resource validate
```
schoren, line 114: would it make sense to address these TODOs in this PR?

server/tracedb/datastoreresource/repository.go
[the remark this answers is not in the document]
schoren: we discussed this and decided we'll keep the separated project. this approach allows more flexibility

server/model/data_stores.go
danielbdias: Yes! I'm working on it.

server/tracedb/datastoreresource/repository_test.go
```diff
@@ -0,0 +1,499 @@
...
+		dataStoreRepository,
+		resourcemanager.WithOperations(datastore.Operations...),
+		resourcemanager.WithIDGen(id.GenerateID),
+	)
+	manager.RegisterRoutes(router)
+
+	return manager
+}
+
+func cleanup(t *testing.T, manager resourcemanager.Manager) {
```
danielbdias, line 44: Without a cleanup structure, this test file surpasses more than 100 idle connections. Later we need to think about how we can improve the RM structure to avoid these open connections.

server/tracedb/datastoreresource/_tests.go
danielbdias: An ill-succeed experiment that I've done. 🤦 🙃

server/tracedb/datastoreresource/repository.go
danielbdias: We have this case today, I believe that we can remove it on the future.

jorgeepc (approved):
LGTM!

PR #2394, opened 2023-04-14
fix(cli): fix TRACETEST_DEV in docker compose installer

### LINKED ISSUE #2348: Substitute TRACETEST_DEV env var from the shell in generated docker compose file
**Context:** We want to use the `TRACETEST_DEV` environment variable in our shell to populate the value inside the Compose file generated by the installer.

**AC1**
As a user that installed a Tracetest server using the docker compose option,
I want to populate the `TRACETEST_DEV` environment variable using the value from my shell,
So the Tracetest server could use the variable internally and stop sending events.

jorgeepc:
This PR applies a fix to the installer to allow us to use the `TRACETEST_DEV` environment variable in our shell to populate the value inside the Compose file generated by the installer.

## Changes

- CLI installer fix

## Fixes

- fixes #2348 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/installer/docker_compose.go
+++ b/cli/installer/docker_compose.go
@@ -71,7 +71,7 @@ func dockerComposeInstaller(config configuration, ui cliUI.UI) {
 	dockerComposeFName := filepath.Join(dir, dockerComposeFilename)
 
 	dockerCmd := fmt.Sprintf(
-		"docker compose -f %s  up -d",
+		"docker compose -f %s up -d",
 		dockerComposeFName,
 	)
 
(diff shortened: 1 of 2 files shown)


cli/installer/docker_compose.go
```diff
@@ -259,6 +260,8 @@ func fixTracetestContainer(config configuration, project *types.Project, version
 	tts.Image = "kubeshop/tracetest:" + version
 	tts.Build = nil
 	tts.Volumes[0].Source = tracetestConfigFilename
+	tracetestDevEnv := "${TRACETEST_DEV}"
+	tts.Environment["TRACETEST_DEV"] = &tracetestDevEnv
```
jorgeepc, line 264: We lose this value the moment we load the yaml config because it gets evaluated as an empty string.

schoren (approved):
LGTM!

xoscar (approved):

PR #2403, opened 2023-04-18
chore(cicd): remove CI pipeline external infra dependencies

schoren:
This PR focuses on making our pipeline easier to use, a bit faster, and remove dependenies with external resources, mainly cloud kubernetes.

Previously, we deployed the code to a kubernetes instance and created a tunnel via port forwarding. This apprach has a few problems:

1. Different e2e test parts (web, server, cli) sometimes interfere with each other. Some test rely on an expected app state, and concurrent tests might modify that state. This means we cannot run different e2e in parallel
2. It's hard to reproduce CI results locally, because envs and states might be difficult to reproduce
3. We flood our docker hub repo with `pr-*` images. They need to be deployed somewhere to be accessible to the deployment.

With this new approach, each e2e test starts its own docker-compose instance. They run in parallel without ever interacting with each other.
The pipeline is defined to use bash scripts that live in the repo. This makes it easy to run the exact same tests with the exact same params on the CI environment and locally.

By leveraging docker export/import of images and GitHub Actions artifacts, we don't need to ever push the image to an external repo. 

There's also some new documentation about how to run and test the app locally in [CONTRIBUTING.md](https://github.com/kubeshop/tracetest/pull/2403/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055)

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test


testing/cli-e2etest/environment/manager.go
```diff
@@ -134,9 +134,15 @@ func (m *internalManager) Start(t *testing.T) {
...
-	// like https://golang.testcontainers.org/quickstart/ "Wait for Log" method
-	time.Sleep(readiness)
+	// wait until tracetest port is ready
+	waitForPort("11633")
+
+	if m.pokeshopEnabled {
+		// wait for pokeshop services
+		waitForPort("8081")
+		waitForPort("8082")
+		time.Sleep(10 * time.Second)
```
danielbdias, line 144: Even with `waitForPort` for some reason Pokeshop drops the connection for a while, this is why I added `time.Sleep` again.

CONTRIBUTING.md
```diff
@@ -72,3 +72,74 @@ If you have any doubts about coding standards you can follow [Effective Go](http
...
+
+As described, the project consists of 3 main components: CLI, server, and web. Each components' directory provides a standard language way to build and test:
+
+- **Web**: uses npm. Commands such as `npm start build`, `npn ci`, etc are available.
+- **CLI**: it is a go module. It can be built with `go build .`, tests are run with `go test ./...` etc
+- **Server**: it is a go module. It can be built with `go build .`, tests are run with `go test ./...` etc
+
+**CLI** and **Server** are mainly built from the project root using [GoReleaser](https://goreleaser.com/) to simplify packaging and distribution.
+
+The server is responsible for serving both the API and the web UI. The easiest way to mix those 2 things for distribution is with docker. We rely on GoReleaser again for packaging all those things together.
```
jorgeepc, line 87: `serving both the CLI and the web UI` ?

CONTRIBUTING.md
```diff
@@ -72,3 +72,74 @@ If you have any doubts about coding standards you can follow [Effective Go](http
...
+It becomes evident that it's not very practical during development to keep all this things in mind. We provide 2 main ways to build, run and test Tracetest:
+
+### Makefile
+
+We provide a [Makefile](./Makefile) with a few targets that helps build all the parts in a development configuration without a lot of intervention. The more common used targets are:
+
+1. **make help**: shows a list of commands with a description.
+2. **make build-docker**: builds the current code (web, cli, server) and packages it all in a tagged docker image. The default image will be "kubeshop/tracetest:ev"
+3. **make build-go**: build the cli and server only
+4. **make build-go**: build the web UI only
```
jorgeepc, line 100:
```suggestion
4. **make build-web**: build the web UI only
```

CONTRIBUTING.md
schoren: it doesn't serve the CLI, it serves the API that the CLI consumes. but it's also the "http server" that delivers the web UI (like index.html, *.js, etc). That's what this means by "serve". Not that it is "at the service of", but that it delivers the web responses for.

Does that make sense? Do you think there's a more clear way to state that?

jorgeepc: I see. No, I think it's ok in that way. Thanks for the details.

xoscar:
I love what we did here, as soon as we get v0.11.10 out the door lets merge this!

PR #2409, opened 2023-04-19
Updating tests to isolate database resources and connections per module

danielbdias:
This PR introduces two updates:
1) Update our test engine to start/stop Postgres containers only when testing a package. This guarantees that a database connection on one test will not leak for another test.
2) Introduces the cleanup concept on Resource Manager, to allow us to clean resources when testing.

This PR is derived from the work on #2385 

Loom: https://www.loom.com/share/a257723b111c4fbba0ea848247f26899

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/resourcemanager/testutil/test_resource.go
```diff
@@ -110,7 +111,6 @@ func testOperation(t *testing.T, op operationTester, rt ResourceTypeTest) {
 	for _, ct := range contentTypeConverters {
 		t.Run(ct.name, func(t *testing.T) {
 			ct := ct
-			t.Parallel()
```
danielbdias, line 113: By running all the tests in parallel, our database connections skyrocket sometimes. By removing serializing this test we allowed our tests to have a chance to cleanup some connections.

schoren (approved):
LGTM!

Issue #2383, opened 2023-04-13
[Network Improvements] Ease networking knowledge needed to use Tracetest


kdhamric:
Currently, some users struggle to use Tracetest due to networking issues. We want to ease this pain.

AC1: 
As a new user to OTel, 
I want to be able to use a minimal configuration in my app instrumentation 
and have the trace data easily appear in Tracetest when not utilizing a direct trace data store.

AC2: 
As a developer, when coding locally on my desktop, I want Tracetest to just work where possible, 
and guide me in the areas where some education is needed.

Proposed Solution:

1) Move to using the default gRPC port Tracetest listens for incoming spans via gRPC to 4317 to match the default (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md). This will allow a minimal configuration. 

2) Allow http as a method to send incoming spans, and listen for this on 4318 (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp)

3) On the Docker setup, expose 4317 and 4318 and route them to either Tracetest or the OTel Collector.

mathnogueira:
We would have to allow users to override the port that Tracetest listens to because it might conflict with an Otel collector running locally. This would prevent users from starting Tracetest.

This config should probably live in the config file and we also should allow users to define env variables to set them (I suspect most people would use env variables for that)

kdhamric:
Will we leverage the 'config resource' for this?

xoscar:
DRAFT: Tasks required to complete this story:

1. Update the otlp server to listen to HTTP incoming requests
2. Update the configuration resource manager to accept values for OTLP ports
3. Add test run events for received data incoming from the collector endpoints
4. Edit examples and recipes to use the new config


@mathnogueira what do you think?

kdhamric:
We may want to consider #1292 as part of this work.

xoscar:
@kdhamric that ticket doesn't have much info, could you add something so we can check if it would match to be done as part of this?

xoscar:
Currently considering this:


As we are making the OTLP endpoints to be more prominent by the [network ease changes](https://github.com/kubeshop/tracetest/issues/2383).
Should we also support a specific DataStore type to point directly to Tracetest? currently, we have OpenTelemetry and the examples are using the otel collector. But we really don't need to collector so we can have another "send open telemetry data directly to Tracetest" data store

xoscar:
[...]

PR #2412, opened 2023-04-19
feature(backend): Adding support for HTTP OTLP server

xoscar:
This PR enables the tracetest server to spin off an OTLP http port that listens to incoming spans

## Changes

- Splits otlp server into three parts (exporter, grpc, http)
- Generates an http port based on the open telemetry specs https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-request

## Fixes

- #2383 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/e05c8a794a3c47dcb660121f23ca58ee


examples/quick-start-jaeger-nodejs/tracetest/collector.config.yaml
```diff
@@ -15,10 +15,16 @@ exporters:
     endpoint: jaeger:14250
     tls:
       insecure: true
+  otlphttp:
+    endpoint: http://host.docker.internal:4317
 
 service:
   pipelines:
-    traces/1:
+    # traces/1:
```
xoscar, line 23: This will be removed for the final PR

examples/quick-start-jaeger-nodejs/tracetest/collector.config.yaml
```diff
@@ -15,10 +15,16 @@ exporters:
     endpoint: jaeger:14250
     tls:
       insecure: true
+  otlphttp:
+    endpoint: http://host.docker.internal:4317
```
mathnogueira, line 19: For http, the correct port is `4318`:

```suggestion
    endpoint: http://host.docker.internal:4318
```

server/app/app.go
```diff
@@ -287,7 +281,22 @@ func registerSPAHandler(router *mux.Router, cfg httpServerConfig, analyticsEnabl
 				Env,
 			),
 		)
+}
 
+func registerOtlpServer(app *App, testDB model.Repository) {
+	exporter := otlp.NewExporter(testDB)
+	grpcOtlpServer := otlp.NewGrpcServer(":4318", exporter)
+	httpOtlpServer := otlp.NewHttpServer(":4317", exporter)
```
mathnogueira, line 289: It's the other way around: 
```suggestion
	grpcOtlpServer := otlp.NewGrpcServer(":4317", exporter)
	httpOtlpServer := otlp.NewHttpServer(":4318", exporter)
```

server/app/app.go
```diff
@@ -287,7 +281,22 @@ func registerSPAHandler(router *mux.Router, cfg httpServerConfig, analyticsEnabl
...
+	go grpcOtlpServer.Start()
+	go httpOtlpServer.Start()
+
+	fmt.Println("OTLP server started on :4318 (grpc) and :4317 (http)")
```
mathnogueira, line 293:
```suggestion
	fmt.Println("OTLP server started on :4317 (grpc) and :4318 (http)")
```

server/otlp/http_server.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (s *HttpServer) Stop() {}
+
+func (s HttpServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var body []byte
+	if b, err := io.ReadAll(r.Body); err == nil {
+		body = b
+	} else {
+		fmt.Println(err)
```
mathnogueira, line 48: We need to return failure responses in those cases otherwise the collector will consider the operation as successful. This applies to all errors

server/otlp/http_server.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		fmt.Println("Error when parsing request", err.Error())
+	}
+
+	_, err = s.exporter.Export(r.Context(), &request)
+	if err != nil {
+		fmt.Println("Error when exporting request", err.Error())
+	}
+
+	w.WriteHeader(http.StatusOK)
```
mathnogueira, line 62: We have to return a success message here for the collector to mark all spans as exported. Reference: https://opentelemetry.io/docs/reference/specification/protocol/otlp/#otlphttp-response

server/otlp/exporter.go
```diff
@@ -3,49 +3,33 @@ package otlp
...
 
-type Server struct {
-	pb.UnimplementedTraceServiceServer
-
-	addr string
-	db   model.Repository
-
-	gServer *grpc.Server
+type IExporter interface {
+	Export(ctx context.Context, request *pb.ExportTraceServiceRequest) (*pb.ExportTraceServiceResponse, error)
```
mathnogueira, line 16: I know the OTLP proto calls it an `export` operation, but in the Tracetest perspective, it should be Import or Ingest (I prefer this one). Can you change it to make it more clear of what this interface should be doing?

mathnogueira (changes requested):
Some changes are needed, but overall, very good.

server/otlp/http_server.go
```diff
@@ -0,0 +1,222 @@
...
+
+	err := proto.Unmarshal(body, &request)
+	if err != nil {
+		return nil, err
+	}
+
+	return &request, nil
+}
+
+func (s HttpServer) parseJson(body []byte) (*pb.ExportTraceServiceRequest, error) {
```
xoscar, line 90: Ideally, we should be able to support both protobuf and JSON body protocols.

The problem is that the defined struct for the incoming request https://github.com/open-telemetry/opentelemetry-proto-go/blob/main/otlp/collector/trace/v1/trace_service.pb.go#L48 requires the resource spans in a JSON to be inside a field called `resource_spans` but the node exporters like  https://www.npmjs.com/package/@opentelemetry/exporter-trace-otlp-http send the data in a different format. This is the format we get:

```
{
  "resourceSpans": [
    {
      "resource": {
        "attributes": [
          {
            "key": "service.name",
            "value": { "stringValue": "unknown_service:node" }
          },
          {
            "key": "telemetry.sdk.language",
            "value": { "stringValue": "nodejs" }
          },
          {
            "key": "telemetry.sdk.name",
            "value": { "stringValue": "opentelemetry" }
          },
```

We are using `json.Unmarshal` and as those fields don't match we are not getting the correct data.
A similar scenario is happening with tempo and can be found here 

https://github.com/kubeshop/tracetest/blob/6a203b23bce8924dc3fc6a1929647fb0ec396aee/server/traces/otel_http_converter.go

server/otlp/http_server.go
```diff
@@ -0,0 +1,222 @@
...
+	case protoBufContentType:
+		return s.parseProtoBuf(body)
+	case jsonContentType:
+		return s.parseJson(body)
+	}
+
+	return nil, fmt.Errorf("content-type %s not supported", contentType)
+}
+
+func (s HttpServer) validateRequest(r *http.Request) (string, string, error) {
```
schoren, line 131: most of this could be replaced by using gorilla mux and define a route. It might look overkill but it's A LOT less code to maintain, supports the gzip middleware out of the box, and we are already using it, so we keep our "http services" consistent across packages

schoren (changes requested):
This is great work oscar! my main concern is that the HTTP server is not following the same patterns as the main http service, which leads to a lot of code that can be replaced by libraries, and having 2 ways of doing the same thing

server/otlp/http_server.go
xoscar: done

server/otlp/http_server.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return s.hServer.Serve(listener)
+}
+
+func (s *HttpServer) Stop() {}
+
+func (s HttpServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	contentType, contentEncoding, err := s.validateRequest(r)
+	shouldDecompress := contentEncoding == gzipEncoding
```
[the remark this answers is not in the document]
xoscar: done

server/otlp/http_server.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	if r.Method != http.MethodPost {
+		return contentType, contentEncoding, fmt.Errorf("method %s not supported", r.Method)
+	}
+	if contentType != "" && contentType != protoBufContentType && contentType != jsonContentType {
+		return contentType, contentEncoding, fmt.Errorf("content-type %s not supported", contentType)
+	}
+
+	return contentType, contentEncoding, nil
+}
```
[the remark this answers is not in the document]
xoscar: removed :D

server/otlp/http_server.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+}
+
+func (s *httpServer) Start() error {
+	r := mux.NewRouter()
+	r.HandleFunc("/v1/traces", s.Export).Methods("POST")
+
+	s.hServer = &http.Server{
+		Addr:    s.addr,
+		Handler: handlers.CompressHandler(decompressBodyHandler(handlers.ContentTypeHandler(r, protoBufContentType, jsonContentType))),
```
schoren, line 49: it look like you're using 2 separated gzip handlers: the `handlers.CompressHandler` and `decompressBodyHandler`. Am I understanding this correctly?

schoren (changes requested):
This is looking really awesome. My only remaining question is about what I think is a duplicated gzip compression handler in the http server. Am I reading that wrong?

server/otlp/http_server.go
xoscar: Yes, so the predefined mux compress handler only "compresses" the response, but does not "decompresses" the input, so I had to add a new handler to do that.

mathnogueira (approved):

PR #2422, opened 2023-04-21
Add data store as resource on CLI

danielbdias:
This PR adds CLI commands to DataStore using the new resource format.

**Note:** just be reviewed only after merging #2385

## Changes

- Added DataStore commands
- Deprecate old datastore commands

## Fixes

- https://github.com/kubeshop/tracetest/issues/2342

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/cmd/datastore_legacy_cmd.go
danielbdias, line 1: I've moved all legacy commands here to make it easy to track them and to be easier to get rid of them in future versions.

cli/actions/datastore.go
```diff
@@ -0,0 +1,147 @@
...
+
+	var dataStore openapi.DataStore
+	mapstructure.Decode(fileContent.Definition().Spec, &dataStore)
+
+	return d.update(ctx, fileContent, currentConfigID)
+}
+
+func (d *dataStoreActions) List(ctx context.Context, args ListArgs) error {
+	return ErrNotSupportedResourceAction
+}
```
jorgeepc, line 61: I remember we decided on having a `list [resource]` and a `get [resource] --id [id]` for every resource.

cli/actions/datastore.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Errorf("could not create file: %w", err)
+	}
+
+	_, err = file.WriteRaw()
+	return err
+}
+
+func (d *dataStoreActions) Delete(ctx context.Context, id string) error {
+	return nil
+}
```
jorgeepc, line 90: We need to support this. Delete the DataStore and return to `No Tracing` mode

jorgeepc (commented):
Hey @danielbdias looks good. Added a couple of comments

cli/actions/datastore.go
xoscar: I believe the idea of not supporting a list for data stores is because there is only one current at a time. Similar to polling profile or config

PR #2436, opened 2023-04-25
fix: improve test and transaction output when it fails

mathnogueira:
This PR improves the generated output when a test or a transaction step fails. Instead of showing `Failed to execute transaction: <reason>` it shows the test name and the link for the user to see more details on the UI.

image

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
LGM!

jorgeepc (approved):

PR #2452, opened 2023-04-27
Add migration to fix data_store ids

danielbdias:
This PR adds a server migration to guarantee that preexisting data_stores will continue working on Tracetest.

## Changes

- Add migration

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/migrations/25_update_data_stores_id.up.sql
+++ b/server/migrations/25_update_data_stores_id.up.sql
@@ -2,7 +2,7 @@ BEGIN;
 
 UPDATE data_stores
 SET id = 'current'
-WHERE id = (select id from data_stores order by created_at limit 1);
+WHERE id = (select id from data_stores where is_default='true' order by created_at limit 1);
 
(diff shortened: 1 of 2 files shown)


server/migrations/25_update_data_stores_id.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+UPDATE data_stores
+SET id = 'current'
+WHERE id = (select id from data_stores order by created_at limit 1);
```
jorgeepc, line 5: Shouldn't we get the one with `is_default = true` ?

danielbdias: 🤦 sure! changing it!

danielbdias: query updated to consider `is_default`

mathnogueira (approved):

Issue #2165, opened 2023-03-15
Remove `The "TRACETEST_DEV" variable is not set.` warning from `tracetest server install`


danielbdias:
When installing the `docker compose` version, we receive the following error:

```sh
WARN[0009] The "TRACETEST_DEV" variable is not set. Defaulting to a blank string. 
```

<img width="914" alt="image" src="https://user-images.githubusercontent.com/12397879/225337564-6464ef1b-9737-4798-9139-83c631e95f95.png">

Today it is harmless, but it could confuse, giving a wrong first impression that the installation had a problem.

PR #2462, opened 2023-04-28
Fix installer tracetest dev docker warning and add options to run installer directly without user input.

danielbdias:
This PR removes a warning that users receive when trying to install Tracetest with docker (as shown on the image below) and add two optional params to skip user input on `tracetest server install`.
image

## Changes

- Added empty "TRACETEST_DEV" env when reading `docker-compose` for the first time to avoid warning
- Adding arguments to avoid user input when running CLI

## Fixes

- #2165

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/installer/docker_compose.go
+++ b/cli/installer/docker_compose.go
@@ -325,6 +325,9 @@ func getCompleteProject(ui cliUI.UI, config configuration) *types.Project {
 	project, err := loader.Load(types.ConfigDetails{
 		WorkingDir:  workingDir,
 		ConfigFiles: configFiles,
+		Environment: map[string]string{
+			"TRACETEST_DEV": "",
+		},
 	})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cli/installer/docker_compose.go
```diff
@@ -325,6 +325,9 @@ func getCompleteProject(ui cliUI.UI, config configuration) *types.Project {
 	project, err := loader.Load(types.ConfigDetails{
 		WorkingDir:  workingDir,
 		ConfigFiles: configFiles,
+		Environment: map[string]string{
+			"TRACETEST_DEV": "",
+		},
```
danielbdias, line 330: This is the main change of this PR to avoid warnings. We use the same set of tools as `docker-compose`, which emits a warning on `loader.Load` even if we are not running the containers.

schoren (approved):
LGTM!

jorgeepc (approved):

PR #2473, opened 2023-05-02
fix(pipelines): fixing environment format

xoscar:
This PR updates the environment format used for the pipelines, it also fixes an issue of having the environment id to be read-only as the generated types were not adding it to structs when marshaling it into a string

## Changes

- environment format for pipelines
- read-only id fix

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## NOTE 
I found a potential problem, when using an environment as part of a test/transaction run, for the first environment creation the id can be omitted, but for the following runs the id needs to be present in order to update the existing one.

This will cause teams having to manually add the ids to their env files in case it already exists

### DIFF
--- a/api/environments.yaml
+++ b/api/environments.yaml
@@ -18,7 +18,6 @@ components:
       properties:
         id:
           type: string
-          readOnly: true
         name:
           type: string
         description:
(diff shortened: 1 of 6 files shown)


jorgeepc (approved):
Looks good! Why the generated openapi files are changing a lot? 😢

xoscar:
> Looks good! Why the generated openapi files are changing a lot? 😢

Could it be because of generating them from Linux vs windows vs mac? 🤷🏾

PR #2474, opened 2023-05-02
fix order of fields when listing resources

mathnogueira:
This PR fixes the order of the fields when listing resources. All fields would be sorted alphabetically, when they should have a logical order (based on the backend structs).

Before:
```yaml
count: 1
items:
    - spec:
          createdat: "2023-04-27T20:15:24.383412Z"
          description: ""
          id: production
          name: Production
          values:
              - key: HOST
                value: http://localhost
              - key: PORT
                value: "11633"
      type: Environment
```

After:
```yaml
count: 1
items:
    - type: Environment
      spec:
          id: production
          name: Production
          description: ""
          createdat: "2023-04-27T20:15:24.383412Z"
          values:
              - key: HOST
                value: http://localhost
              - key: PORT
                value: "11633"
```

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


xoscar (approved):
I like this! 🔥

jorgeepc (approved):

PR #2486, opened 2023-05-04
feat: change openapi spec for transactions to consider resource manager

mathnogueira:
This PR changes the CRUD operations on transactions to use the resource manager strategy

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -84,7 +84,7 @@ paths:
                   items:
                     type: array
                     items:
-                      $ref: "./transactions.yaml#/components/schemas/Transaction"
+                      $ref: "./transactions.yaml#/components/schemas/TransactionResource"
             text/yaml:
               schema:
                 type: object
@@ -94,7 +94,7 @@ paths:
                   items:
                     type: array
                     items:
-                      $ref: "./transactions.yaml#/components/schemas/Transaction"
+                      $ref: "./transactions.yaml#/components/schemas/TransactionResource"
         400:
           description: "invalid query for transactions, some data was sent in incorrect format."
         500:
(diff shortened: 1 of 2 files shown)


api/openapi.yaml
```diff
@@ -74,22 +74,34 @@ paths:
...
-                items:
-                  $ref: "./transactions.yaml#/components/schemas/Transaction"
+                type: object
+                properties:
+                  count:
+                    type: integer
+                  items:
+                    type: array
+                    items:
+                      $ref: "./transactions.yaml#/components/schemas/Transaction"
```
jorgeepc, line 87: Should we update this Transaction def to include the `type` and `spec` format?

schoren (changes requested):
@mathnogueira there is one change that we need to implement, regarding the use of augmented entities. We have now 2 separated fields for Steps (full test) and stepIDs (just the ids).

For backwards compatibility with the yaml file, I used the `steps` name for the stepIDs, and   `fullSteps` for the field containing the entire test definition of each step. Feel free to use a better name for this.

api/openapi.yaml
schoren: +1

mathnogueira: oh yeah, forgot we don't have those fields in this object. Fixing that!

mathnogueira:
> fullSteps

Something like this?

```yaml
Transaction:
    type: object
    properties:
        steps:
          type: array
          description: list of steps of the transaction containing just each test id
          items:
            type: string
        fullSteps:
          type: array
          description: list of steps of the transaction containing the whole test object
          items:
            $ref: "./tests.yaml#/components/schemas/Test"
```

schoren:
@mathnogueira yeap, exactly that!

schoren (approved):

PR #2487, opened 2023-05-04
feat(server): improve resource manager test suit debug output

schoren:
This PR improves the resource manager reusable test suit. It dumps the http response when a test fails, so it's eaiser to understand why the test is not passing.

Example:

Screenshot 2023-05-04 at 15 30 09

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/resourcemanager/testutil/common_functions.go
+++ b/server/resourcemanager/testutil/common_functions.go
@@ -4,12 +4,26 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
+	"net/http/httputil"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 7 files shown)


server/resourcemanager/testutil/common_functions.go
```diff
@@ -4,13 +4,27 @@ import (
...
+
+	if success {
+		return
+	}
+
+	b, _ := httputil.DumpResponse(resp, true)
+	t.Log("\n", string(b))
+	t.FailNow()
+}
+
```
danielbdias, line 27: ❤️

danielbdias (approved):

Issue #2306, opened 2023-04-03
[CLI Improvements] Add dashboard command


schoren:
defined [here](https://github.com/kubeshop/tracetest/issues/2182#issuecomment-1490819722).

Add a new `dashboard` command that opens the dashboard in the default browser. The dashboard address is the same as the server url used for the API requests.

We are not going to do any validation at this time, other than having a valid server url configured, either from the config file or from the `--server-url` flag.

PR #2490, opened 2023-05-04
feature(cli): Adding dashboard command

xoscar:
This PR adds the dashboard command that opens the configured tracetest server URL using the default browser

## Changes

- Adds new dashboard command to open default server url
- Validates if the config is not empty

## Fixes

- #2306 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


jorgeepc (approved):

cli/utils/common.go
```diff
@@ -34,3 +38,21 @@ func StringReferencesFile(path string) bool {
 	// otherwise the user also could send a directory by mistake
 	return info != nil && !info.IsDir()
 }
+
+func OpenBrowser(url string) error {
```
danielbdias, line 42: Philosophical question: should we detect if the client has the UI enabled on the SO before opening the browser to emit a warning and stop the command? 
I was thinking of something to warn the user if they are trying to execute this command in a CI container.

danielbdias (approved):

cli/utils/common.go
```diff
@@ -1,9 +1,13 @@
 package utils
 
 import (
+	"fmt"
 	"io"
+	URL "net/url"
```
schoren, line 6: this is not a very standard thing to do in go. we usually just use the package name directly, specially with stdlib packages. In the case that you have a conflicting var name like in the `OpenBrowser` func, just rename the variable. In this kind of short funcs, you could just use `u string` for the var name, or `rawURL` to indicate that it's an unparsed URL, just as quick examples

cli/utils/common.go
schoren: I think we agreed on just launching without any other checks, because there is a large number of things that can go wrong and it's difficult to check them all, at least for the initial version

schoren (changes requested):
Looks good except for a few non-go-looking bits

cli/cmd/dashboard_cmd.go
```diff
@@ -0,0 +1,43 @@
...
+	GroupID: cmdGroupMisc.ID,
+	Use:     "dashboard",
+	Short:   "Opens the Tracetest Dashboard URL",
+	Long:    "Opens the Tracetest Dashboard URL",
+	PreRun:  setupCommand(),
+	Run: func(_ *cobra.Command, _ []string) {
+		analytics.Track("Dashboard", "cmd", map[string]string{})
+		url := fmt.Sprintf("%s://%s", cliConfig.Scheme, cliConfig.Endpoint)
+
+		if cliConfig.IsEmpty() {
```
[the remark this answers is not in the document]
xoscar: Done!

xoscar:
Hey @schoren thanks for the feedback!, I just updated the PR with the fixes and improvements you mentioned

schoren (approved):
Looking great!

PR #2493, opened 2023-05-05
fix(frontend): migrate transactions to resource architecture

jorgeepc:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


api/triggers.yaml
```diff
@@ -7,15 +7,12 @@ components:
...
-        triggerSettings:
-          type: object
-          properties:
-            http:
-              $ref: "./http.yaml#/components/schemas/HTTPRequest"
-            grpc:
-              $ref: "./grpc.yaml#/components/schemas/GRPCRequest"
-            traceid:
-              $ref: "./traceid.yaml#/components/schemas/TRACEIDRequest"
+        http:
```
xoscar, line 10: There's a mismatch between the resource version and the regular version, I removed the `triggerSettings` to have them both work the same

server/http/mappings/tests.go
```diff
@@ -109,11 +109,9 @@ func (m OpenAPI) Outputs(in maps.Ordered[string, model.Output]) []openapi.TestOu
 func (m OpenAPI) Trigger(in model.Trigger) openapi.Trigger {
 	return openapi.Trigger{
 		TriggerType: string(in.Type),
-		TriggerSettings: openapi.TriggerTriggerSettings{
-			Http:    m.HTTPRequest(in.HTTP),
-			Grpc:    m.GRPCRequest(in.GRPC),
-			Traceid: m.TRACEIDRequest(in.TraceID),
-		},
```
xoscar, line 116: Removing trigger settings

server/tests/transactions.go
```diff
@@ -66,7 +66,7 @@ func (t Transaction) NewRun() TransactionRun {
 	return TransactionRun{
 		TransactionID:      t.ID,
 		TransactionVersion: t.GetVersion(),
-		CreatedAt:          time.Now(),
+		CreatedAt:          time.Now().UTC(),
```
xoscar, line 69: This was a bug, we need to always use UTC from the server

web/src/models/Transaction.model.ts
```diff
@@ -2,33 +2,43 @@ import {Model, TTransactionsSchemas} from 'types/Common.types';
...
   }
 >;
 
 function Transaction({
+  spec: rawTransaction = {},
+}: TRawTransactionResource): Transaction {
+  return Transaction.FromRawTransaction(rawTransaction);
+}
+
+Transaction.FromRawTransaction = ({
```
xoscar, line 22: Currently, we have to support two versions of the transactions API object:
- resource (from the resource endpoints)
- regular raw version (from the result list home page)

web/src/redux/apis/endpoints/Transaction.endpoint.ts
```diff
@@ -1,18 +1,21 @@
 import {HTTP_METHOD} from 'constants/Common.constants';
 import {TracetestApiTags} from 'constants/Test.constants';
-import Transaction, {TRawTransaction} from 'models/Transaction.model';
+import Transaction, {TRawTransaction, TRawTransactionResource} from 'models/Transaction.model';
 import TransactionService from 'services/Transaction.service';
 import {TTestApiEndpointBuilder} from 'types/Test.types';
 import {TDraftTransaction} from 'types/Transaction.types';
 
+const defaultHeaders = {'content-type': 'application/json', 'X-Tracetest-Augmented': 'true'};
```
xoscar, line 8: Including the augmented header

server/tests/transactions.go
schoren: why? I don't think we set any TZ on any other place

web/src/redux/apis/endpoints/Transaction.endpoint.ts
schoren: not sure how this works, but just to be safe: for requesting the yaml definition, you need to NOT pass the header.

xoscar: Yep we have that covered 😄

Issue #2304, opened 2023-04-03
[CLI Improvements] Support `all` resource


schoren:
We want to support the `all` resource for verbs. It acts as a pseudo resources that actually wraps all available resources.

For example, if you have configured a DataStore, a Demo, and a test, you could export them all to a yaml file like this:

```
tracetest list all -o yaml > dump.yaml
```

you could import that dump with the following command:

```
tracetest apply all -f dump.yaml
```

Verbs supported:
- list
- apply

`get` and `delete` always require an ID, and that ID only makes sense when refering to a specific resource, so it doesn't really apply for those.

schoren:
There is an inconsistency here. Not all resources support the `list` verb. For example, you cannot list `Config`s since there is only one. One solution would be to make `all` ignore those constraints and somehow force the resource to be part of the list. This is a bit inconsistent. We could get rid of the inconcistency by "faking" the `list` verb support for such resources, but it still feels hacky.

Another solution would be to merge the `get` and `list` verbs, as we disucssed at some point. With this approach, we would replace this with `get all`, that would work for single and multiple resources (Config and Tests, for example). To get a single row you'd do `tracetest get test --id someid` and skipping the `--id` flag would just get the list. We discussed this approach, refering to how `kubectl` does the `get` verb.


@kdhamric @olha23 thoughts?

olha23:
@schoren not sure that i am the best person to suggest here but i think that both solutions have their pros and cons, so it ultimately depends on specific use case and preferences. If the goal is to simplify the CLI and merge the list and get verbs, then introducing a new "get all" verb could work. This would make it consistent for all resources and avoid confusion between the list and get verbs.

kdhamric:
The issue currently with the config resource is that it is hard to understand how to 'get' it.

If I do:
tracetest get config             
ERROR	id of the resource to get must be specified

So... I need to know the id. So I would naturally do:
tracetest list config
ERROR	failed to list for type: config	{"error": "the specified resource type doesn't support the action"}

I am then 'stuck' not knowing the id for a config (it is current, but how can I know this as a new user)

Solution:
I think we should support list. By default, it shows a nice, column based output with key info... particularly the id

schoren:
@kdhamric we defined that "unique" resources like Config do not require the `--id`, as defined [here](https://github.com/kubeshop/tracetest/issues/2182#issuecomment-1490694741).

Does that make any difference in your thoughts?

kdhamric:
[...]

PR #2501, opened 2023-05-07
feature(cli): adding support for the all flag when listing resources

xoscar:
This PR enables the `--all` flag for the `list` verb for every resource.

Use `tracetest list <resource> --all` to get all resources at once, it can be combined with the different output types: `json`, `yaml` and `pretty`

## Changes

- Enables all flag from the CLI
- Adds list endpoint for config, datastores and polling
- Updates OpenAPI with the new list resource specs

## Fixes

- #2304 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


jorgeepc (approved):
💣

Issue #2305, opened 2023-04-03
[CLI Improvements] add validations for command's flags and arguments


schoren:
Commands have a well defined set of flags and/or arguments, and what they should look like. 

We currently validate one parameter at a time, so if a user has 2 incorrect parameters, they need to run the command once, see the 1st error, fix it. Run the command again, see the 2nd error, fix it. Run the command again.

We can validate the entire command parameters at once, and show all errors at once, so the user doesn't need to do multiple retries for the same command.

Issue #2298, opened 2023-04-03
[CLI Improvements] validate that server and CLI version match


schoren:
When running any cli command that communicates with a server, we want to make sure that they are both speaking the same version, to avoid unexpected behaviors.

schoren:
Probably a question for @kdhamric and @olha23. We talked about not allowing to run commands if version don’t match, but we didn’t go into more details

xoscar:
Team, by going through this issue I found a couple of things we might need to consider:
1. What happens if the server is behind the CLI? Should we prompt an error message?
2. If we want to enforce that the server and CLI versions match, we'll have to trigger a request to get the server version before executing any command/verb. Is this what we want?

schoren:
1. I think we defined a first iteration where ANY kind of version missmatch triggers a message on the CLI side prompting users to upgrade.
2. If I remember correctly we already talked about having the server either exposing a `/version` (or similar) endpoint. I think we can also consider addding an `X-Tracetest-Version` header in all responses, so that there's no need for a special request to validate the response. On  any request made by the CLI, it can compare the versions.

kdhamric:
@schoren response sound good to me

PR #2523, opened 2023-05-11
feature(CLI): CLI overhaul improvements

xoscar:
## Reasoning
While working on the CLI improvement tickets there were multiple issues that had to be fixed in order to allow us to scale the codebase, these are:
- The config setup was creating instances for components that not all  of the commands would use `setupCommand()`
- We were using hoisted variables to keep track of parameters and setup (cli config, logger, etc)
- No clear way of running validation checks
- Problems when understanding the data flow when executing a command, what block initializes which variables, and what gets updated, validated
- Lack of ways to share functionality between commands

Two key issues that showcase the issues we had are:
- https://github.com/kubeshop/tracetest/issues/2305
- https://github.com/kubeshop/tracetest/issues/2298

Where we don't have clear ways or layers to run validations

## Command types

**Flows.** These are interactive commands, guiding users through multiple steps while asking for input
**Resources.** Everything related to the server resources management
**Misc.** Simple commands like displaying the version or the dashboard
**Global.** Space for the root command to be configured
**Deprecated.** Everything that is no longer active

### Encapsulation

Every command type includes the following:
- **Setup.** Initializes the required dependencies
- **Register.** Adds the commands to the tree
- **Parameters.** Types and functionality to handle command arguments
- **Formatters.** Custom formatting logic for the command types
- **Actions.** Comand logic (triggering an API, saving files, etc)
- **Commands.** Composing the command instances with the different pieces (setup, parameters, formats, actions, etc)

### Composing, composing, composing
The global layer contains the base instances and logic for all of the command types, every type can use them as a base to create customized instances depending on what is needed. Example:

The global setup initializes base instances for the config, logger, and analytics
And the resources setup composes a new setup which includes the resource registry and resource actions

**Commands can also create a custom setup if required**

### Middlewares
These are composable functions for the three main command execution sections, pre-run, run, and post-run. Allowing us to share custom logic across the whole command tree.

**Similar to the rest of the setup, commands can work without the need of middlewares**

### New folder structure
```
/cli
  /global
  /resources
  /deprecated
  /misc
  /flows
```

## Changes
- Stopping using "global variables" for parameters and setup
- Splitting the different commands by types (misc, resources, flows, deprecated)
- Having visible layers for setup, parameters, execution, and cleanup
- Enabling composable runners, setup, and validators using middleware
- Allowing explicit command registration on creation
- Flexibility

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Reasoning
https://www.loom.com/share/bfe74cbd8c33443c94275867d42f4bf5

## SHOW ME THE MONEY
https://www.loom.com/share/d1d86055da714e2ab11d9d74484b2a62


kdhamric (approved):
I watched both the Loom videos. Sounded good if the rest of the team is onboard.

cli/flows/server.go
```diff
@@ -0,0 +1,42 @@
...
+
+	root.Cmd.AddCommand(server.Cmd)
+	return server
+}
+
+var server Server
+
+func GetServer() Server {
+	if server.Cmd == nil {
+		panic("server not initialized")
```
danielbdias, line 38: We can be more specific here to avoid confusion with the Tracetest API Server start:
```suggestion
		panic("server command was not initialized")
```

cli/global/middlewares/teardown.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/kubeshop/tracetest/cli/analytics"
+	global_setup "github.com/kubeshop/tracetest/cli/global/setup"
+	"github.com/spf13/cobra"
+)
+
+func WithTeardownMiddleware(setup *global_setup.Setup) Middleware {
+	return func(runFn RunFn) RunFn {
+		return func(cmd *cobra.Command, args []string) (string, error) {
```
danielbdias, line 11: This command feels sometimes shady for me because it obliges us to know that we need to do a teardown. What happens if do we forget to call it and use `setup.Logger` or `analytics`? 

To think later: is there a way to keep these commands register to suffer the cleanup only if used? With that we could keep the golang behavior: 
```go
    resource := someAction()
    defer resourceCleanup()
```

cli/global/setup/setup.go
```diff
@@ -0,0 +1,126 @@
...
+		err := option(s, args)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	return "", nil
+}
+
+func WithInitAnalytics() SetupOption {
```
danielbdias, line 46: Thinking about the `Teardown` discussion: this option feels more to a decorator that uses a resource for the entire command run than just an option. Maybe we could decorate commands that uses analytics and logger to have the startup/teardown logic encapsulated in one place. It makes sense?

danielbdias: I was thinking on something like this:

```go
func DecorateCommandWithAnalytics(cmd *cobra.Command) cmd {
    // decorate PreRun
    // decorate PostRun

    return decoratedCommand
}
```

cli/resources/setup/setup.go
```diff
@@ -0,0 +1,104 @@
...
+		resourceRegistry.Register(environmentActions)
+
+		setup.ResourceRegistry = resourceRegistry
+		return nil
+	}
+}
+
+func WithResourceActions() SetupOption {
+	return func(setup *Setup, args []string) error {
+		resourceType := args[0]
```
danielbdias, line 95: We should validate here if there is at least one argument:

```suggestion
		if len(args) <= 0 {
            return fmt.Errorf("you should pass a resourceType to run the setup")
        }
        resourceType := args[0]
```

cli/main.go
```diff
@@ -1,9 +1,23 @@
...
+	"fmt"
+	"os"
+
+	"github.com/kubeshop/tracetest/cli/global"
+	global_register "github.com/kubeshop/tracetest/cli/global/register"
 )
 
+func Execute() {
+	root := global.NewRoot()
+	global_register.Register(root)
```
danielbdias, line 13: On this command we register all other commands, right? It makes sense to centralize all command registrations in one single file, just to see all commands that the CLI has?

danielbdias (approved):
The new structure is pretty interesting, and I believe we can proceed with it. 

I've added some comments on things we could consider in the future, but they can be addressed in other PRs.

Good job @xoscar !

cli/global/setup/setup.go
xoscar: Makes sense, I'm going to work on an update on this between hackathon off time haha

PR #2534, opened 2023-05-12
fix environment migration

mathnogueira:
This PR finishes migrating the environment from the old api endpoints to the new resource format.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/model/environments.go
danielbdias, line 1: Instead of having an `Environment2`, can we kill this model? By doing that we are able to see if we migrated all the env stuff too. Does it make sense?

danielbdias (approved):

server/model/environments.go
mathnogueira: Forgot to delete it haha

cli/formatters/environment.go
```diff
@@ -56,11 +59,19 @@ func (f EnvironmentsFormatter) ToStruct(file *file.File) (interface{}, error) {
 	var environmentResource openapi.EnvironmentResource
 	nullableEnvironment := openapi.NewNullableEnvironmentResource(&environmentResource)
 
+	if strings.HasSuffix(file.Path(), ".yaml") || strings.HasSuffix(file.Path(), ".yml") {
```
schoren, line 62: Do we need this validation? Maybe we could try to parse the file, assuming that file definitions *must* be yaml, and throw an error if it's not parseable. Does that make sense?

server/http/controller.go
```diff
@@ -29,19 +30,20 @@ import (
...
-	runner          runner
-	newTraceDBFn    func(ds datastoreresource.DataStore) (tracedb.TraceDB, error)
-	mappers         mappings.Mappings
-	triggerRegistry *trigger.Registry
-	version         string
+	tracer                trace.Tracer
+	testDB                model.Repository
+	runner                runner
+	newTraceDBFn          func(ds datastoreresource.DataStore) (tracedb.TraceDB, error)
+	environmentRepository environment.Repository
```
schoren, line 37: you are only using the `Get` function, so maybe instead of depending on the repo directly, you could depend on a local interface:
```
type envGetter interface {
  Get(context.Context, id.ID) (environment.Environment, error) 
}

type controller struct {
  //...
  environmentRepository envGetter
}
```

that way if we need to pass a mock or any other substitution, we only need to implement that func and not the entire repo

server/environment/model.go
```diff
@@ -22,6 +23,16 @@ type (
 )
 
 func (e Environment) Validate() error {
+	if e.Name == "" {
```
schoren, line 26: ❤️ validation being used

server/http/controller_test.go
```diff
@@ -120,6 +121,7 @@ func setupController(t *testing.T) controllerFixture {
 			nil,
 			nil,
 			mappings.New(traces.NewConversionConfig(), comparator.DefaultRegistry(), mdb),
+			environment.NewRepository(nil),
```
schoren, line 124: referecing my comment about depending on an interface, this could also be just `nil` instead of an invalid repo, which reads nicer I think (personal opinion)

server/model/repository.go
```diff
@@ -32,15 +32,6 @@ type RunRepository interface {
 	GetLatestRunByTestVersion(context.Context, id.ID, int) (Run, error)
 }
 
-type EnvironmentRepository interface {
```
schoren, line 35: ❤️

schoren (approved):
Left a few suggestions, but this looks great! love how much code you killed

cli/formatters/environment.go
mathnogueira: is it guaranteed that the file will always be yaml?

mathnogueira: A solution (I don't like it much) is to try to parse it as YAML, and then as JSON as a fallback:

```go
func (f EnvironmentsFormatter) ToStruct(file *file.File) (interface{}, error) {
	var environmentResource openapi.EnvironmentResource
	nullableEnvironment := openapi.NewNullableEnvironmentResource(&environmentResource)

	if err := yaml.Unmarshal([]byte(file.Contents()), &environmentResource); err == nil {
		return environmentResource, nil
	}

	err := nullableEnvironment.UnmarshalJSON([]byte(file.Contents()))
	if err != nil {
		return nil, err
	}
	return environmentResource, nil
}
```

schoren: the original implementation fallbacks to json, and that is also not guaranteed. I think in the end its a product decision: do we want to support files in formats other than YAML?

danielbdias: We could assume that it will be YAML and return an error message in case of a parsing error, to avoid having too much parsing logic or guessing on the CLI. Does it make sense?

mathnogueira: I'm considering the file ContentType. It's hardcoded as YAML. But I still left the json fallback there just in case we change that in the future

PR #2556, opened 2023-05-19
feat: transaction cli

mathnogueira:
This PR adds the code for the CLI to migrate the transaction commands to the resource manager structure

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/a.yaml
```diff
@@ -0,0 +1,7 @@
+type: Transaction
```
mathnogueira, line 1: For testing. Will be deleted when PR is ready

server/tests/transactions.go
```diff
@@ -609,7 +609,6 @@ func (r *TransactionsRepository) readRow(ctx context.Context, row scanner, augme
 
 func removeNonAugmentedFields(t *Transaction) {
 	t.CreatedAt = nil
-	t.Version = nil
```
mathnogueira, line 612: We use that in the list endpoint, so I removed this from the augmented version

cli/actions/transaction.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if newFile.HasID() {
+		rawTransaction, err := actions.formatter.ToStruct(&newFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse file into transaction: %w", err)
+		}
+
+		transaction := rawTransaction.(openapi.TransactionResource)
+		return actions.resourceClient.Update(ctx, newFile, *transaction.Spec.Id)
```
schoren, line 46: won't this fail if the file has an ID but it's not yet created? I think this is what happens with environments

server/tests/transactions.go
schoren: this will also impact on the yaml file generation. why do you need the version on the transaction list? and if so, it would be easier to just request the augmented list

schoren (changes requested):

cli/actions/transaction.go
danielbdias: Indeed we receive a 404 when we don't have an ID.

cli/utils/api.go
```diff
@@ -86,13 +86,14 @@ func (resourceClient ResourceClient) NewRequest(url, method, body, contentType s
 	request.Header.Set("x-client-id", analytics.ClientID())
 	request.Header.Set("Content-Type", contentType)
 	request.Header.Set("Accept", contentType)
+	request.Header.Set("X-Tracetest-Augmented", fmt.Sprint(augmented))
```
danielbdias, line 89: Should we add this header only when `augmented` is `true`?

cli/cmd/get_cmd.go
```diff
@@ -35,6 +35,10 @@ var getCmd = &cobra.Command{
 			return "", err
 		}
 
+		if output == string(formatters.JSON) {
+			ctx = context.WithValue(ctx, "X-Tracetest-Augmented", true)
+		}
+
```
danielbdias, line 41: Is there a reason to propagate this value in a context (like propagating it deeper on the call stack)? We could use the options pattern to pass this value.

mathnogueira: This was the easiest way to propagate the value. The problem with the option would be that we would change the interface of all the client methods.

mathnogueira: but we can do it if you think that's better

schoren (approved):
LOGTM!

PR #2561, opened 2023-05-22
fix: add versioning for trigger encoding

schoren:
This PR fixes a retrocompat issue, where tests created before this version would fail to be correctly read from DB. This was caused by a change in the JSON encoding of the `model.Trigger` entity.

This PR fixes this by adding basic versioning support for the entity

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/model/trigger_json.go
+++ b/server/model/trigger_json.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/fluidtruck/deepcopy"
 )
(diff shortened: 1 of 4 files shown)


jorgeepc (approved):

server/model/trigger_json.go
```diff
@@ -0,0 +1,72 @@
...
+		return deepcopy.DeepCopy(v1, t)
+	}
+
+	// v2
+	v2 := triggerJSONV2{}
+	err = json.Unmarshal(data, &v2)
+	if err != nil {
+		return err
+	}
+	if v2.valid() {
```
danielbdias, line 67: Is it expected to fail silently (returning `nil`) if both `.valid()` methods return false?

danielbdias (approved):

server/model/trigger_json.go
schoren: Good catch. I added an error here

PR #2570, opened 2023-05-23
fix(cli): fix upsert for environments

schoren:
This PR fixes environment upsert in cli

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/actions/environments.go
```diff
@@ -51,12 +51,17 @@ func (environment environmentsActions) Apply(ctx context.Context, fileContent fi
 
 	mapstructure.Decode(fileContent.Definition().Spec, &envResource.Spec)
 
-	if envResource.Spec.Id == nil || *envResource.Spec.Id == "" {
-		result, err := environment.resourceClient.Create(ctx, fileContent)
-		return result, err
+	if envResource.Spec.GetId() != "" {
+		_, err := environment.Get(ctx, envResource.Spec.GetId())
+		if err != nil {
```
danielbdias, line 56: We need to add validation here because we need guarantees that this error happens because of a missing entity. 

One way is to emit a specialized error like "ResourceNotFound" to do that. I did this here: https://github.com/kubeshop/tracetest/commit/9e9ec20c2de6a80f7a308bc05b450b5ef21054d1#diff-660f1c4de144c4d56d5c007996cd5725aa2bb25a43caf6cb6f7600424790062fR154

I've added this validation on my CLI e2e branch to print a "Resource not found message" in case of a user asks for an entity that doesn't exist.

danielbdias (commented):
The PR sounds fine to solve this problem. 

However, we must revisit a minor issue our `Get` method has on the CLI of not handling 404 errors well. I've added a comment on that. Does it makes sense?

cli/actions/environments.go
schoren: I like that

danielbdias (approved):

PR #2571, opened 2023-05-23
Adding upsert capability for Environment resource

danielbdias:
This PR adds upsert capabilities to the `Environment` resource, allowing us to create an environment with an ID

## Changes

- Updated Environment Resource manager Update method to have behave like an upsert

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/environment/manager.go
+++ b/server/environment/manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


schoren:
@danielbdias 2 points why I don't like this PR:

1. it makes `Update` actually an `upsert`, so the intention is not clear and introduces non-obvious side effects.
2. this upsert thing will be useful for other resources, if not all, so it's probablly a good idea to make this a new endpoint of the resourcemanager.

I opened this [pr to fix the CLI for environments](https://github.com/kubeshop/tracetest/pull/2570) and I think that's a better place to have this "hotfix"

PR #2600, opened 2023-05-26
feature(cli): adding flag and parameter validations for commands

xoscar:
This PR adds parameter based logic to validate different flags and args for each of the commands, returns all errors at once depending on the validation checks

## Changes

- Adds new middleware to support parameter validation
- Adds new parameters interface and validators for commands

## Fixes

- #2305 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/9efaab50749045bf9ce4557b93230806

### DIFF
--- a/cli/parameters/installer.go
+++ b/cli/parameters/installer.go
@@ -3,6 +3,20 @@ package parameters
 import (
 	"github.com/kubeshop/tracetest/cli/installer"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
+)
+
+var (
+	AllowedRunEnvironments = []installer.RunEnvironmentType{
+		installer.DockerRunEnvironmentType,
+		installer.KubernetesRunEnvironmentType,
+		installer.NoneRunEnvironmentType,
+	}
+	AllowedInstallationMode = []installer.InstallationModeType{
+		installer.WithDemoInstallationModeType,
+		installer.WithoutDemoInstallationModeType,
+		installer.NotChosenInstallationModeType,
+	}
 )
 
 type InstallerParams struct {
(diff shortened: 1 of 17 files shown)


cli/parameters/installer.go
```diff
@@ -0,0 +1,42 @@
...
+			Message:   "run-environment must be one of 'none', 'docker' or 'kubernetes'",
+		})
+	}
+
+	if cmd.Flags().Lookup("mode").Changed && p.InstallationMode != installer.NotChosenInstallationModeType && p.InstallationMode != installer.WithDemoInstallationModeType && p.InstallationMode != installer.WithoutDemoInstallationModeType {
+		errors = append(errors, ParamError{
+			Parameter: "mode",
+			Message:   "mode must be one of 'not-chosen', 'with-demo' or 'just-tracetest'",
+		})
+	}
```
danielbdias, line 32: Can we move this validation to a "Contains" style? Something like this:

```go
// after the imports
var (
  AllowedRunEnvironments = []installation.RunEnvironmentType{
    installer.DockerRunEnvironmentType,
    installer.KubernetesRunEnvironmentType,
    installer.NoneRunEnvironmentType,
  }
  AllowedInstallationMode = []installation.InstallationModeType{
    installer.WithDemoInstallationModeType,
    installer.WithoutDemoInstallationModeType,
    installer.NotChosenInstallationModeType,
  }
)

// here on the validations
if cmd.Flags().Lookup("run-environment").Changed && !slices.Contains(AllowedRunEnvironments, p.RunEnvironment) {
 // same code
}

if cmd.Flags().Lookup("mode").Changed && !slices.Contains(AllowedInstallationMode, p. InstallationMode) {
 // same code
}
```

cli/parameters/resources.go
```diff
@@ -0,0 +1,120 @@
...
+	ResourceName string
+}
+
+var _ Params = &ResourceParams{}
+var ValidResources = []string{"config", "datastore", "demo", "environment", "pollingprofile", "transaction"}
+
+func (p *ResourceParams) Validate(cmd *cobra.Command, args []string) []ParamError {
+	errors := make([]ParamError, 0)
+
+	p.ResourceName = args[0]
```
danielbdias, line 95: We should validate if the resource name came in args (i.e., `if len(args) > 0`). Otherwise, we can have a panic here.

danielbdias (commented):
The overall PR is fine! Before merging it can you address just the `len(args)` comment, please?

cli/parameters/installer.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"github.com/kubeshop/tracetest/cli/installer"
+	"github.com/spf13/cobra"
+)
+
+type InstallerParams struct {
+	Force             bool
+	RunEnvironment    installer.RunEnvironmentType
+	InstallationMode  installer.InstallationModeType
```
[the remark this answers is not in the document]
xoscar: Yeah, I think we should focus on validations for the moment, we can update types later on

cli/parameters/installer.go
xoscar: I didn't know about the slices lib haha thanks for pointing that out!

cli/parameters/resources.go
xoscar: Donerino 😄

jorgeepc (approved):
I like it. Muy bueno! 🔥

cli/parameters/installer.go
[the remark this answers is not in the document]
xoscar: I got your point now, the problem I see is that the validations are done one at a time, and ideally we'll want to show all possible errors at once, does it make sense?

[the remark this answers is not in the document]
danielbdias: It makes sense! Later we can drop this support for the `Value` interface and focus on the validation that we do here.

PR #2618, opened 2023-05-30
docs: add table driven tests script example

schoren:
This PR adds an example of how one can use tracetest to run table driven test. It provides a script, an example csv data file and yaml test definition, and some basic documentation.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


examples/table-driven-test/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+![test run result](./test-run.png)
+
+## How it works
+
+Table driven tests works by iterating over a list of values, and running the same test with different parameters each time.
+Tracetest supports running tests with parameters via environment variables. We are only missing an automated way to pass a list of values.
+
+The `table-test.bash` script does exactly that. It relies on bash and a few POSIX commands to parse a given CSV file.
```
adnanrahic, line 22:
```suggestion
The `table-test.bash` script does exactly that. It relies on Bash and a few POSIX commands to parse a given CSV file.
```

examples/table-driven-test/README.md
```diff
@@ -0,0 +1,32 @@
...
+For each row of the file, it creates a new environment in tracetest, named after the test ID and current csv line number.
```
adnanrahic, line 23:
```suggestion
For each row of the file, it creates a new environment in Tracetest, named after the test ID and current CSV line number.
```

PR #2622, opened 2023-05-31
feat: improved poller algorithm

mathnogueira:
This PR introduces another algorithm parameter for our trace poller. When we don't detect changes in the trace anymore, we try matching the selectors with spans. If all selectors get at least one span, we consider the polling successful, otherwise, try again in the next polling cycle. If we have no changes in the trace and we cannot match the selectors with spans from the trace 3 times in a row, we end the polling and mark it as complete (to prevent issues when writing TDD)

There are more things that could be done and  I'll do in the future if needed:
* Make this configurable via a polling profile option (either another strategy, or just an option inside the `periodic` object)

## Example

| # Run | Action |
|---|---|
| # 01  | Got initial trace |
| # 02  | Got more spans, continue |
| # 03  | Got more spans, continue |
| # 04  | No more spans, but not all selectors are working |
| # 05  | No more spans, but not all selectors are working |
| # 06  | No more spans, but not all selectors are working. But we tried 3 times already, so finish polling |
```

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/executor/default_poller_executor.go
+++ b/server/executor/default_poller_executor.go
@@ -78,11 +78,9 @@ func NewPollerExecutor(
 		eventEmitter: eventEmitter,
 	}
 
-	selectedBasedExecutor := NewSelectorBasedPoller(defaultExecutor)
-
 	return &InstrumentedPollerExecutor{
 		tracer:         tracer,
-		pollerExecutor: selectedBasedExecutor,
+		pollerExecutor: defaultExecutor,
 	}
 }
 
--- a/server/executor/default_poller_executor.go
+++ b/server/executor/default_poller_executor.go
@@ -194,8 +194,12 @@ func (pe DefaultPollerExecutor) donePollingTraces(job *PollingRequest, traceDB t
 	if !traceDB.ShouldRetry() {
 		return true, "TraceDB is not retryable"
 	}
-	pp := *pe.ppGetter.GetDefault(job.Context()).Periodic
-	maxTracePollRetry := pp.MaxTracePollRetry()
+	pp := pe.ppGetter.GetDefault(job.Context())
+	if pp.Periodic == nil {
+		return false, "Polling profile not configured"
+	}
+
+	maxTracePollRetry := pp.Periodic.MaxTracePollRetry()
 	// we're done if we have the same amount of spans after polling or `maxTracePollRetry` times
 	if job.count == maxTracePollRetry {
 		return true, fmt.Sprintf("Hit MaxRetry of %d", maxTracePollRetry)
--- a/server/executor/default_poller_executor.go
+++ b/server/executor/default_poller_executor.go
@@ -157,11 +157,6 @@ func (pe DefaultPollerExecutor) ExecuteRequest(request *PollingRequest) (bool, s
 		return false, "", run, nil
 	}
 
-	err = pe.eventEmitter.Emit(ctx, events.TracePollingSuccess(request.test.ID, request.run.ID, reason))
-	if err != nil {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: server/executor/trace_poller.go [hunk 1]
```diff
@@ -72,25 +73,42 @@ type tracePoller struct {
 }
 
 type PollingRequest struct {
-	ctx        context.Context
-	test       model.Test
-	run        model.Run
-	count      int
-	hadRequeue bool
```
mathnogueira, line 79: The idea behind changing this object is to make it extendable via headers (just like network protocols work by attaching headers to the packet). Thus, I removed the `context` which can be serialized via context propagation and the `hadRequeue`.

This way, we can have the counter for the selector based polling executor without changing this object and impacting other strategies..

server/executor/selector_based_poller_executor.go
```diff
@@ -0,0 +1,65 @@
...
+func (pe selectorBasedPollerExecutor) ExecuteRequest(request *PollingRequest) (bool, string, model.Run, error) {
+	ready, reason, run, err := pe.pollerExecutor.ExecuteRequest(request)
+	if !ready {
+		request.SetHeader(selectorBasedPollerExecutorRetryHeader, "0")
+		return ready, reason, run, err
+	}
+
+	currentNumberTries := pe.getNumberTries(request)
+	if currentNumberTries >= selectorBasedPollerExecutorMaxTries {
+		return true, "not all selectors matched, but trace haven't changed in a while", run, err
```
schoren, line 32: would the end user, at some point, have a clear undestanding of what `in a while` means? Like, is the amount of retries related to this error at some point? If not, maybe add the amount of retries/time in this message?

server/executor/trace_poller.go
```diff
@@ -72,25 +73,42 @@ type tracePoller struct {
...
+	pr.headers[name] = value
+}
+
+func (pr *PollingRequest) Header(name string) string {
+	return pr.headers[name]
 }
 
 func (pr PollingRequest) IsFirstRequest() bool {
-	return !pr.hadRequeue
+	return pr.Header("requeued") != "true"
```
schoren, line 96: another point where a `return !pr.GetBool("requeued")` would look cool

server/executor/default_poller_executor.go
```diff
@@ -189,7 +194,7 @@ func (pe DefaultPollerExecutor) donePollingTraces(job *PollingRequest, traceDB t
 	if !traceDB.ShouldRetry() {
 		return true, "TraceDB is not retryable"
 	}
-	pp := *pe.ppGetter.GetDefault(job.ctx).Periodic
+	pp := *pe.ppGetter.GetDefault(job.Context()).Periodic
```
schoren, line 197: since you're here, can you add a `nil` check for `Peridiodic`? It should never be nil, but it's better to be safe than panic

danielbdias (approved):

xoscar:
Question, do we have/want event logs for the new polling strategy? so we can see the failed selectors from the client side

server/executor/selector_based_poller_executor.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/kubeshop/tracetest/server/model"
+)
+
+const (
+	selectorBasedPollerExecutorRetryHeader = "SelectorBasedPollerExecutor::retryCount"
```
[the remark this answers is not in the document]
mathnogueira: > Question, do we have/want event logs for the new polling strategy? so we can see the failed selectors from the client side

added some logs to it! Good catch

xoscar (approved):

Issue #2565, opened 2023-05-23
datastore not provisioning


BartNetJS:
This is my yaml snipped to provision tracetest:

```
provisioning.yaml: |-
    |
      ---
      # Datastore is where your application stores its traces. You can define several different datastores with
      # different names, however, only one is used by Tracetest.
      # You can see all available datastore configurations at https://kubeshop.github.io/tracetest/supported-backends/
      type: DataStore
      spec:
        name: Jaeger
        # Indicates that this datastore is a jaeger instance
        type: jaeger
        isdefault: true
        # Configures how tracetest connects to jaeger.
        jaeger:
          endpoint: jaeger-query.jp-integration.svc.cluster.local:16685
          tls:
            insecure: true
      ---
      type: Config
      spec:
        analyticsEnabled: true
      ---
      type: PollingProfile
      spec:
        name: Custom Profile
        strategy: periodic
        default: true
        periodic:
          timeout: 30s
          retryDelay: 500ms
```

After install the datastore is not provisioned:

![image](https://github.com/kubeshop/tracetest/assets/4865355/6397db7b-662a-4cd1-a48e-860efb0da2c4)

adnanrahic:
Hi @BartNetJS !

Thanks for opening this issue. Could you provide a bit more info about your setup?

- What version of Tracetest are you using?
- Is the provisioning file being loaded? ie. do the filenames match?
- Is it the first time this particular Tracetest server is started? (The provision only works if that particular Tracetest server has not been started yet)

If you could also provide steps to reproduce, that would be awesome!

BartNetJS:
Hi @adnanrahic 

The version is 0.11.8

The provision.yaml file exist in the config folder:
![image](https://github.com/kubeshop/tracetest/assets/4865355/44cfa51d-5403-4e65-bdba-7c8e578e8452)

I tried a helm install and helm upgrade.

The initial yaml file was created with 
`helm template tracetest kubeshop/tracetest > tracetest-kubernetes-manifests.yaml`
as mentioned in the documentation.

I've changed the url for the jaeger instance and did an upgrade.
I noted this og:

```
2023/05/23 08:22:26 Config file used:  /app/config/config.yaml
tracetest-server v0.11.8 (main)
Starting
New install? false
```

And then deleted the release and deployed again.
I saw that **New install? false** was still there.
And the datastore was not configured.

BartNetJS:
I also can't change the datastore in th UI, see gif
[...]

PR #2631, opened 2023-06-01
fix: make resource manager take server prefix into consideration

mathnogueira:
This PR makes the resource manager take the server path prefix into consideration when building the routes. Related to one problem related in #2565 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


danielbdias (approved):

schoren:
@mathnogueira can you add some kind of test for for this? I wouldn't like to have a regression on this

schoren (approved):

xoscar (approved):

PR #2681, opened 2023-06-07
feat: test models + list tests endpoint

mathnogueira:
This PR adds the list tests endpoint. I'll add the resource manager test in the next PR (POST /api/tests) because it will be way easier to do that way. It's hard to implement it right now because there's no easy way of adding the test into the database (I don't want to mess up with the `models.Test` object.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


.github/workflows/pull-request.yaml
```diff
@@ -88,7 +88,7 @@ jobs:
           cache: true
           cache-dependency-path: go.work
       - name: Run unit tests
-        run: cd server; make test
+        run: cd server; make test -B
```
mathnogueira, line 91: For some reason, this was not running the tests and was printing `test is up-to-date`. Adding the `-B` forces make to run this regardless of the cache.

server/app/app.go
```diff
@@ -254,17 +256,13 @@ func (app *App) Start(opts ...appOption) error {
...
-	registerDemosResource(demoRepo, apiRouter, db, provisioner, tracer)
-
-	registerDataStoreResource(dataStoreRepo, apiRouter, db, provisioner, tracer)
-	registerlinterResource(linterRepo, apiRouter, db, provisioner, tracer)
+	registerConfigResource(configRepo, apiRouter, provisioner, tracer)
+	registerPollingProfilesResource(pollingProfileRepo, apiRouter, provisioner, tracer)
+	registerEnvironmentResource(environmentRepo, apiRouter, provisioner, tracer)
+	registerDemosResource(demoRepo, apiRouter, provisioner, tracer)
+	registerDataStoreResource(dataStoreRepo, apiRouter, provisioner, tracer)
+	registerlinterResource(linterRepo, apiRouter, provisioner, tracer)
```
schoren, line 264: the `L` should be uppercase here
```suggestion
	registerLinterResource(linterRepo, apiRouter, provisioner, tracer)
```

server/app/app.go
```diff
@@ -320,7 +318,7 @@ func registerOtlpServer(app *App, testDB model.Repository, eventEmitter executor
 	})
 }
 
-func registerlinterResource(linterRepo *linterResource.Repository, router *mux.Router, db *sql.DB, provisioner *provisioning.Provisioner, tracer trace.Tracer) {
+func registerlinterResource(linterRepo *linterResource.Repository, router *mux.Router, provisioner *provisioning.Provisioner, tracer trace.Tracer) {
```
schoren, line 321: the `L` should be uppercase here
```suggestion
func registerLinterResource(linterRepo *linterResource.Repository, router *mux.Router, provisioner *provisioning.Provisioner, tracer trace.Tracer) {
```

server/http/custom_routes.go
```diff
@@ -43,7 +43,6 @@ func (c *customController) Routes() openapi.Routes {
 
 	routes[c.getRouteIndex("GetTestRuns")].HandlerFunc = c.GetTestRuns
 
-	routes[c.getRouteIndex("GetTests")].HandlerFunc = paginatedEndpoint[openapi.Test](c.service.GetTests, c.errorHandler)
```
schoren, line 46: ❤️

server/test/test_json.go
```diff
@@ -0,0 +1,42 @@
...
+	Name       string      `json:"name,omitempty"`
+	Assertions []Assertion `json:"assertions"`
+}
+
+type namedAssertion struct {
+	Name       string
+	Assertions []Assertion
+}
+
+type keyValueSpec struct {
```
schoren, line 16: I think it would be better to version the entire `TestSpec` json instead of just the things that change, like here: https://github.com/kubeshop/tracetest/blob/main/server/model/trigger_json.go

This makes it easier to keep updating the json, adding new versions, with a standardized approach

server/test/test_json_test.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/kubeshop/tracetest/server/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSpecRetrocompability(t *testing.T) {
```
schoren, line 12: Same for the tests: https://github.com/kubeshop/tracetest/blob/main/server/model/trigger_json_test.go

It's easier to understand the chronological order of the versions

server/test/test_repository.go
```diff
@@ -0,0 +1,236 @@
...
+
+	testMaxVersionQuery = `
+	INNER JOIN (
+		SELECT id as idx, max(version) as latest_version FROM tests GROUP BY idx
+		) as latest_tests ON latest_tests.idx = t.id AND t.version = latest_tests.latest_version
+	`
+)
+
+func (r *repository) List(ctx context.Context, take, skip int, query, sortBy, sortDirection string) ([]Test, error) {
+	hasSearchQuery := query != ""
```
schoren, line 84: we have some reusable sql functions in the `pkg/sqlutil` package. for example:

- search: https://github.com/kubeshop/tracetest/blob/main/server/tests/transactions.go#L433
- sort: https://github.com/kubeshop/tracetest/blob/main/server/tests/transactions.go#L328

server/test/test_repository.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	return count, nil
+}
+
+type scanner interface {
+	Scan(dest ...interface{}) error
+}
+
+func (r *repository) readTestRows(ctx context.Context, rows *sql.Rows) ([]Test, error) {
```
schoren, line 163: These new repos aren't shared, so we can normalize in the functions not including the entity name
```suggestion
func (r *repository) readRows(ctx context.Context, rows *sql.Rows) ([]Test, error) {
```

server/test/test_repository.go
```diff
@@ -0,0 +1,236 @@
...
+			return []Test{}, err
+		}
+
+		tests = append(tests, test)
+	}
+
+	return tests, nil
+}
+
+func (r *repository) readTestRow(ctx context.Context, row scanner) (Test, error) {
```
schoren, line 178: Same here

```suggestion
func (r *repository) readRow(ctx context.Context, row scanner) (Test, error) {
```

server/test/test_repository.go
```diff
@@ -0,0 +1,236 @@
...
+			test.Summary.LastRun.Passes = *pass
+		}
+		if fail != nil {
+			test.Summary.LastRun.Fails = *fail
+		}
+
+		return test, nil
+	default:
+		return Test{}, err
+	}
```
schoren, line 235: since we don't check for the error type anymore, we can remove the switch and make the code more "standard" looking:

```suggestion
	if err != nil {
		return Test{}, fmt.Errorf("cannot read row: %w", err)
	}

	err = json.Unmarshal(jsonServiceUnderTest, &test.Trigger)
	if err != nil {
		return Test{}, fmt.Errorf("cannot parse trigger: %w", err)
	}

	err = json.Unmarshal(jsonSpecs, &test.Specs)
	if err != nil {
		return Test{}, fmt.Errorf("cannot parse specs: %w", err)
	}

	err = json.Unmarshal(jsonOutputs, &test.Outputs)
	if err != nil {
		return Test{}, fmt.Errorf("cannot parse outputs: %w", err)
	}

	if lastRunTime != nil {
		test.Summary.LastRun.Time = *lastRunTime
	}
	if pass != nil {
		test.Summary.LastRun.Passes = *pass
	}
	if fail != nil {
		test.Summary.LastRun.Fails = *fail
	}

	return test, nil
```

api/triggers.yaml
```diff
@@ -17,9 +21,13 @@ components:
     TriggerResult:
       type: object
       properties:
+        # TODO: remove after test migration
         triggerType:
           type: string
           enum: ["http", "grpc", "traceid"]
+        type:
+          type: string
+          enum: ["http", "grpc", "traceid"]
```
xoscar, line 30: We can update the FE to use the new version right?

server/app/app.go
```diff
@@ -254,17 +256,13 @@ func (app *App) Start(opts ...appOption) error {
...
-
-	registerDataStoreResource(dataStoreRepo, apiRouter, db, provisioner, tracer)
-	registerlinterResource(linterRepo, apiRouter, db, provisioner, tracer)
+	registerConfigResource(configRepo, apiRouter, provisioner, tracer)
+	registerPollingProfilesResource(pollingProfileRepo, apiRouter, provisioner, tracer)
+	registerEnvironmentResource(environmentRepo, apiRouter, provisioner, tracer)
+	registerDemosResource(demoRepo, apiRouter, provisioner, tracer)
+	registerDataStoreResource(dataStoreRepo, apiRouter, provisioner, tracer)
+	registerlinterResource(linterRepo, apiRouter, provisioner, tracer)
+	registerTestResource(testRepo, apiRouter, provisioner, tracer)
```
xoscar, line 265: 🔥

server/test/tests.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (t Test) GetID() id.ID {
+	return t.ID
+}
+
+func (t Test) Validate() error {
+	return nil
+}
+
+func (sar SpanAssertionResult) SafeSpanIDString() string {
```
schoren, line 167: Can we rename this? the `safe` is implied by the return type. it's a string, which always has a value, and no error
```suggestion
func (sar SpanAssertionResult) SpanIDString() string {
```

server/test/trigger/grpc.go
```diff
@@ -0,0 +1,65 @@
+package trigger
```
xoscar, line 1: I like that we moved the triggers here!

server/test/tests.go
```diff
@@ -0,0 +1,256 @@
...
+
+	if e.Expression == nil {
+		if e.LiteralValue.Type == "attribute" {
+			return fmt.Sprintf("attr:%s", e.LiteralValue.Value)
+		}
+
+		return e.LiteralValue.Value
+	}
+
+	return fmt.Sprintf("%s %s %s", e.LiteralValue.Value, e.Operation, e.Expression.String())
```
schoren, line 192: we could reduce the cyclomatic complexity here

```suggestion
	if e.Expression != nil {
		return fmt.Sprintf("%s %s %s", e.LiteralValue.Value, e.Operation, e.Expression.String())
	}
	
	if e.LiteralValue.Type == "attribute" {
		return fmt.Sprintf("attr:%s", e.LiteralValue.Value)
	}

	return e.LiteralValue.Value
```

xoscar (approved):
Hey @mathnogueira this looks great! One question have you done anything about the resource endpoint FE uses for the dashboard? or should stay as it is for the moment?

Awesome job! 🔥

server/test/trigger/traceid.go
```diff
@@ -0,0 +1,11 @@
+package trigger
+
+const TriggerTypeTRACEID TriggerType = "traceid"
```
schoren, line 3:
```suggestion
const TriggerTypeTraceID TriggerType = "traceid"
```

server/test/trigger/trigger.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Compatibility with older versions
+type triggerFormat struct {
+	Type    TriggerType     `json:"type"`
+	HTTP    *HTTPRequest    `json:"http,omitempty"`
+	GRPC    *GRPCRequest    `json:"grpc,omitempty"`
+	TraceID *TraceIDRequest `json:"traceid,omitempty"`
+}
+
+type oldTriggerFormat struct {
```
schoren, line 31: again, I think having "versioned" formats makes it easier to understand the chronology: https://github.com/kubeshop/tracetest/blob/main/server/model/trigger_json.go

server/test/trigger/trigger.go
```diff
@@ -0,0 +1,61 @@
...
+	Type    TriggerType     `json:"triggerType"`
+	HTTP    *HTTPRequest    `json:"http,omitempty"`
+	GRPC    *GRPCRequest    `json:"grpc,omitempty"`
+	TraceID *TraceIDRequest `json:"traceid,omitempty"`
+}
+
+func (t *Trigger) UnmarshalJSON(data []byte) error {
+	var trigger triggerFormat
+	err := json.Unmarshal(data, &trigger)
+	if err != nil || trigger.Type == "" {
```
schoren, line 41: I think the current main version of this function is more future-friendly with the versioning: https://github.com/kubeshop/tracetest/blob/main/server/model/trigger_json.go#LL49C6-L49C6

schoren (changes requested):
This is looking great. There are some improvements that we can make, since we're already working on this are, so I added some comments.

Specially around versioning of json representations. I think it's very important that we standardize on a way to handle this, since it's happened a few times so far, and it's very likely to keep happening

mathnogueira:
@schoren I changed all the things you mentioned

schoren (approved):
this is great! lef a few optional nits, but it's ready to merge!

Issue #2590, opened 2023-05-25
Add Microsoft Application Insights as a supported trace datastore


kdhamric:
We want to add Microsoft Application Insights as a supported trace datastore so Azure customers using tracing can utilize Tracetest and trace-based testing techniques against their distributed apps. We have confirmed with the team at Microsoft that the trace data can be queried via the Azure Monitor Log Analytics API (https://learn.microsoft.com/en-us/azure/azure-monitor/logs/api/overview). 

AC1: Microsoft Application Insights is added to the Settings screen under Configure Data Store

AC2: Once configured, Tracetest is able to use the API to successfully pull back a trace from Azure Monitor and build tests based on it.

AC3: Microsoft Application Insights is added as a supported trace data store in the docs under [https://docs.tracetest.io/configuration/overview](https://docs.tracetest.io/configuration/overview). 

AC4: There is a recipe showing how to use Microsoft Application Insights with an Azure based project and test the project via Tracetest.

AC5: This integration is added to the Tracetest.io integration page.

PR #2691, opened 2023-06-08
feature: Azure Application Insights Integration

xoscar:
This PR adds the Azure Application Insights data store integration

## Changes

- Allows data stores to have a direct vs a collector config
- Enables Azure App insights as data store
- Enables config and start up for new data store

## Fixes

- #2590 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/6ff8f264bb7748329050ca0132c71aa7


api/dataStores.yaml
```diff
@@ -84,6 +86,17 @@ components:
...
         token:
           type: string
+    AzureAppInsights:
+      type: object
+      properties:
+        useAzureActiveDirectoryAuth:
+          type: boolean
+        accessToken:
+          type: string
+        connectionType:
```
xoscar, line 96: I think this connection type is something we can move to be part of the base data store object, I don't want to do it now but it will be very useful in the future

server/tracedb/datastoreresource/resource_types.go
```diff
@@ -202,21 +223,26 @@ func (ds DataStore) GetID() id.ID {
 }
 
 func (ds DataStore) IsOTLPBasedProvider() bool {
+	if ds.Type == DatastoreTypeAzureAppInsights {
```
xoscar, line 226: In the future, I'd prefer having each integrator decide if they are otlp or not based on the values and such. this patch is temporary

mathnogueira (approved):
Looks good!

kdhamric (commented):
Looks great!!

Issue #2682, opened 2023-06-07
[Analyzer Results] Do all steps of a test rather than failing on an Analyzer Score not being reached


kdhamric:
Currently, the new analyzer, if enabled, checks to see if the score is below the threshold value and stops the test processing, leaving the test in a state of ANALYZING_ERROR. Rather than doing this, processing the test should continue, progressing through to the FINISHED state.

AC1: When running a test, the processing should not stop
        based on the score for the analyzer being below a certain threshold

AC2: If the Analyzer is enabled, 
         the Trace screen should indicate if the overall score, ie the Trace Analyzer Result, 
         was below the threshold for a successful run. It should clearly indicate what the overall 
         Trace Analyzer Score was, and what the minimal acceptable score was at the time this test run was run,
         and whether this was a passing test score or not. Suggested wording (although a graphic may work better):
         Overall Trace Analyzer Score: 61  Minimum acceptable score: 70. This trace analyzer result is: Failed
         
We will need other followup issues to make the CLI handle failure to meet the score, but this first change is needed so test specs run properly regardless of the passing or failing of the test score.

xoscar:
@olha23 do you mind taking a look at this, we would need to have a place in the `trace` tab to show that the analyzer failed because of the minimum score rule

jorgeepc:
Looks good. One question @kdhamric, we are going to fail the test if the score for the analyzer is below the threshold, right? I mean, we are going to execute the specs but we want to mark the test as failed. Is that correct?

kdhamric:
In the UI, we want to start indicating score information and whether the test met the minimum score. @olha23 has mocks in #2650 that show in the results view the overall score, and the image is in red or green to indicate whether the trace analyzer passed the minimum score.

We do, however, need to mark the entire test run as passed or failed, which is indicated by the green or red circle at the beginning of a test run, based on some logic. Were there any assertions that failed? Is the trace analyzer on, and if so, was the overall score below the minimum? We may have a setting to say 'score the traces, show pass or fail for it, but do not fail the overall test (ie just show scores and whether we passed or not).

Rather than having this logic in the front end and in the CLI, we may want to centralize it and deliver it as part of the test run info. Creating an issue to describe the problem.

olha23:
@kdhamric i created the mockup for this, please check here. I prefer V2 
[...]

PR #2697, opened 2023-06-09
feature: Improving Analyzer Failure State

xoscar:
This PR improves the way we handle error states for the analyzer. Instead of stopping the test execution, now clients will be reading the linter results to validate if there was a failure or not, and show the proper results.

## Changes

- Removes stopping of test runs based on analyzer errors
- Updates analyzer errors UI to better showcase the minimum score error
- Updates transaction execution card and test run card to show the proper status

## Fixes

- #2682 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/6624b56c915b4c41a07f52179bbbdbf7

### DIFF
--- a/server/linter/resource/model.go
+++ b/server/linter/resource/model.go
@@ -3,9 +3,7 @@ package linter_resource
 import (
 	"fmt"
 
-	"github.com/kubeshop/tracetest/server/model"
 	"github.com/kubeshop/tracetest/server/pkg/id"
-	"golang.org/x/exp/slices"
 )
 
 type (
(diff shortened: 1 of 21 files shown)


server/linter/resource/model.go
```diff
@@ -45,36 +43,3 @@ func (l Linter) HasID() bool {
 func (l Linter) GetID() id.ID {
 	return l.ID
 }
-
-func (l Linter) ValidateResult(result model.LinterResult) error {
```
xoscar, line 49: Not needed, clients will check score vs minimum score to validate it

server/openapi/model_linter_result.go
```diff
@@ -10,6 +10,8 @@
 package openapi
 
 type LinterResult struct {
+	MinimumScore int32 `json:"minimumScore,omitempty"`
```
xoscar, line 13: We are now going to persis the minimum score that was used during the execution, not using the global

web/src/constants/Theme.constants.ts
```diff
@@ -25,6 +25,7 @@ export const theme: DefaultTheme = {
     md: '14px',
     lg: '16px',
     xl: '18px',
+    xxxl: '24px',
```
jorgeepc, line 28: `xxxl`? 👀

jorgeepc (approved):
LGTM!

web/src/constants/Theme.constants.ts
xoscar: I don't know man, 20 should be xxl si I just went for it 😂

Issue #2155, opened 2023-03-13
Port number when setting up Grafana with http not shown correctly


kdhamric:
AC1: As a user setting up Grafana Tempo via HTTP,
the warning message about the default port does not seem correct
as it appears to expect the grpc port

AC2: As a user setting up Grafana Tempo via HTTP,
I would want the test connection to verify that I am talking to Graafana Tempo 
and not a random http server (google.com in this example)
by querying a known resource on the server (list of traces, etc)

When setting up Grafana Tempo with a http address, if no port is entered our error messages are misleading as it does not show a port number (perhaps is should default to 443 for https, 80 for http):

![image](https://user-images.githubusercontent.com/1385894/224843658-285ce4ee-fd64-4c58-9269-0d6a91c130ff.png)

PR #2704, opened 2023-06-09
fix(backend): adding empty port as supported for tempo

xoscar:
This PR removes the warning for http ports when using tempo as a data store

## Changes

- Adds empty port as valid for tempo
- Updates libs to parse urls

## Fixes

- #2155 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/tracedb/datasource/http.go
+++ b/server/tracedb/datasource/http.go
@@ -9,9 +9,9 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
 	"strings"
 
+	"github.com/goware/urlx"
 	"github.com/kubeshop/tracetest/server/model"
 	"github.com/kubeshop/tracetest/server/tracedb/connection"
 	"github.com/kubeshop/tracetest/server/tracedb/datastoreresource"
@@ -25,7 +25,7 @@ type HttpClient struct {
 }
 
 func NewHttpClient(name string, config *datastoreresource.HttpClientConfig, callback HttpCallback) DataSource {
-	endpoint, _ := url.Parse(config.Url)
+	endpoint, _ := urlx.Parse(config.Url)
 	client := &http.Client{
 		Transport: &http.Transport{
 			TLSClientConfig: getTlsConfig(config.TLS),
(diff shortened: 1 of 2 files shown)


server/tracedb/datasource/http.go
```diff
@@ -25,7 +25,7 @@ type HttpClient struct {
 }
 
 func NewHttpClient(name string, config *datastoreresource.HttpClientConfig, callback HttpCallback) DataSource {
-	endpoint, _ := url.Parse(config.Url)
+	endpoint, _ := urlx.Parse(config.Url)
```
xoscar, line 28: This fixes an annoying but with the URL lib not setting up the expected values when parsing the url

server/tracedb/tempodb.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func tempoDefaultPorts() []string {
-	return []string{"9095"}
+	return []string{"9095", ""}
```
xoscar, line 26: For http urls like www.grafana.com/traces the port is "not" setup

Ideally, I'd like us to change this logic to allow each integration read the config file and based on the values decide what is the expected list of port instead of having a static list

server/tracedb/datasource/http.go
danielbdias: What is the main difference between using `url.Parse` and `urlx.Parse`? Is there a way to keep `url.Parse`?

xoscar: Here's the main difference https://github.com/goware/urlx#difference-between-urlx-and-neturl

We are using it here as well https://github.com/kubeshop/tracetest/blob/0bcfa3a3624e29ec8c71d534f3341dd12c315f2d/server/executor/trigger/http.go

danielbdias: What is my concern: https://github.com/goware/urlx is a small library without changes for 2 years and we are using only a portion of its capability. 
My thought is that we can transform the `url.URL` return like this lib does.

danielbdias (approved):

Issue #2713, opened 2023-06-12
[Analyzer Results] Fix validations for Tracetest custom attributes


xoscar:
Currently, the tracetest analyzer is validating the custom telemetry data created by the server itself, we should fix this by aligning the produced attributes to the semantic conventions.

The attributes that are causing issues are:
1. name
2. startTime
3. endTime
4. parent_id

#### Acceptance Criteria

**AC1**
As a user looking at the analyzer results
I should see no problems with the custom attributes and spans generated by Tracetest

PR #2717, opened 2023-06-12
feature(backend): updating metadata fields to match semantic conventions

xoscar:
This PR updates the tracetest metadata attributes to match the semantic conventions

## Changes

- Creates MetadataFields type for tracetest related stuff
- Implements usage of constant types for spans
- Updates FE to use new fields

## Fixes

- #2713 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/f78af5a0e33f4a1e9e812acb5115c856

### DIFF
--- a/server/expression/data_store.go
+++ b/server/expression/data_store.go
@@ -29,12 +29,12 @@ func (ds AttributeDataStore) getFromAlias(name string) (string, error) {
 	alias, found := attributeAlias[name]
 
 	if !found {
-		return "", fmt.Errorf(`attribute "%s" not found`, alias)
+		return "", fmt.Errorf(`attribute "%s" not found`, name)
 	}
 
 	value := ds.Span.Attributes.Get(alias)
 	if value == "" {
-		return "", fmt.Errorf(`attribute "%s" not found`, alias)
+		return "", fmt.Errorf(`attribute "%s" not found`, name)
 	}
 
 	return value, nil


server/model/spans.go
```diff
@@ -9,6 +9,18 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
+type TracetestMetadataField string
```
schoren, line 12: do we really need this type? it seems that all the references to values for this convert it to string, and no other operation is done. Given that this is just a constant to avoid repetition, and it's very unlikely that we'd want to attach methods to it, does it make sense to keep it as a simple "string"?

server/model/spans.go
```diff
@@ -9,6 +9,18 @@ import (
...
+
+const (
+	TracetestMetadataFieldStartTime TracetestMetadataField = "tracetest.span.start_time"
+	TracetestMetadataFieldEndTime   TracetestMetadataField = "tracetest.span.end_time"
+	TracetestMetadataFieldDuration  TracetestMetadataField = "tracetest.span.duration"
+	TracetestMetadataFieldType      TracetestMetadataField = "tracetest.span.type"
+	TracetestMetadataFieldName      TracetestMetadataField = "tracetest.span.name"
+	TracetestMetadataFieldParentID  TracetestMetadataField = "tracetest.span.parent_uid"
```
schoren, line 20: is this field renamed to `parent_Uid` intentionally or is it a typo?
```suggestion
	TracetestMetadataFieldParentID  TracetestMetadataField = "tracetest.span.parent_id"
```

server/model/traces.go
```diff
@@ -25,7 +25,7 @@ func NewTrace(traceID string, spans []Span) Trace {
 
 	rootSpans := make([]*Span, 0)
 	for _, span := range spanMap {
-		parentID := span.Attributes["parent_id"]
+		parentID := span.Attributes[string(TracetestMetadataFieldParentID)]
```
schoren, line 28: this values are not the same. the original key is `parent_id`, but the value defined in the constant is `tracetest.span.parent_uid`. Is this intentional? Doesn't this changes the key?

schoren (changes requested):
this looks great! the constants make the code nicer. I think there might be a typo on one of the keys, and also this is changing one of the keys, not sure if that's intentional or not

server/http/mappings/traces.go
```diff
@@ -57,6 +57,7 @@ func (m OpenAPI) Span(in model.Span) openapi.Span {
 		EndTime:    in.EndTime.UnixMilli(),
 		Attributes: attributes,
 		Children:   m.Spans(in.Children),
+		Name:       in.Name,
```
xoscar, line 60: I just found this bug we have had for a while now haha

server/expression/data_store.go
```diff
@@ -21,10 +25,24 @@ func (ds AttributeDataStore) Source() string {
 	return "attr"
 }
 
+func (ds AttributeDataStore) getFromAlias(name string) (string, error) {
```
xoscar, line 28: Adding alias check, in this case we only support `name` for the moment

server/model/spans.go
xoscar: Done 😄

server/model/spans.go
xoscar: This was definitely a fat finger

server/model/traces.go
xoscar: Fixed!

server/expression/data_store.go
```diff
@@ -21,10 +25,24 @@ func (ds AttributeDataStore) Source() string {
...
+	alias, found := AttributeAlias[name]
+	if found {
+		value := ds.Span.Attributes.Get(alias)
+		if value == "" {
+			return "", fmt.Errorf(`attribute "%s" not found`, alias)
+		}
+
+		return value, nil
+	}
```
schoren, line 37: I think we can reduce nesting here: 
```suggestion
	if !found  {
	  return "", fmt.Errorf(`attribute "%s" not found`, alias)
	}
	value := ds.Span.Attributes.Get(alias)
	if value == "" {
		return "", fmt.Errorf(`attribute "%s" not found`, alias)
	}
	
	return value, nil
```

server/expression/data_store.go
```diff
@@ -13,6 +13,10 @@ type DataStore interface {
 	Get(name string) (string, error)
 }
 
+var AttributeAlias = map[string]string{
```
schoren, line 16: can this be unexported?

schoren (approved):
left 2 suggestions that IMO might make the code cleaner, other than that it looks great!

jorgeepc (approved):
Tested it, looks good!

PR #2740, opened 2023-06-15
fix(backend): matching trigger span duration and trigger time durations

xoscar:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/model/traces.go
+++ b/server/model/traces.go
@@ -105,8 +105,8 @@ func spanType(attrs Attributes) string {
 }
 
 func spanDuration(span Span) string {
-	timeDifference := span.EndTime.Sub(span.StartTime)
-	return fmt.Sprintf("%d", int64(timeDifference))
+	timeDifference := timeDiff(span.StartTime, span.EndTime)
+	return fmt.Sprintf("%d", durationInNanoseconds(timeDifference))
 }
 
 func (t *Trace) Sort() Trace {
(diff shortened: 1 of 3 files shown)


server/model/traces.go
```diff
@@ -105,8 +105,8 @@ func spanType(attrs Attributes) string {
 }
 
 func spanDuration(span Span) string {
-	timeDifference := span.EndTime.Sub(span.StartTime)
-	return fmt.Sprintf("%d", int64(timeDifference))
+	timeDifference := timeDiff(span.StartTime, span.EndTime)
+	return fmt.Sprintf("%d", durationInNanoseconds(timeDifference))
```
schoren, line 109: is this `durationInNanoseconds` needed? you need to follow an extra indirection that is only a single line.

schoren (approved):
LGTM! I would remove that very tiny function but it's not critical

jorgeepc (approved):

server/model/traces.go
schoren: Actually, you don't even need to cast to int, `%d` can handle int64 correctly: https://go.dev/play/p/8Vth4PCiQyO

```suggestion
	return fmt.Sprintf("%d", timeDifference.Nanoseconds())
```

xoscar: Shoot sorry, I didn't see this. I'll add them in my next PR

PR #2743, opened 2023-06-15
feat: improve analyzer errors

### LINKED ISSUE #2677: [Analyzer] UX - Error message improvement
Please check the video https://www.loom.com/share/b3efe5ef26044db59c78cd4f0f72191b

See example:
### Instead of this text:

```
**Required Attributes by Span Type:**
Attribute "http.target" is missing from span of type "http"
Attribute "net.host.name" is missing from span of type "http"

**Attribute Naming:**
Attribute "name" doesnt follow naming convention
Attribute "tracetest.span.startTime" doesnt follow naming convention
Attribute "tracetest.span.endTime" doesnt follow naming convention
Attribute "parent_id" doesnt follow naming convention

**Enforce HTTPS Protocol:**
Insecure http schema found for attribute: http.scheme. Value: http
```

### i propose use this template:

```
**Missing Attributes for Span Type:**
The span of type "http" is missing the required attributes:
- "http.target"
- "net.host.name"

**Attribute Naming:**
The following attributes do not adhere to the naming convention:
- "name"
- "tracetest.span.startTime"
- "tracetest.span.endTime"
- "parent_id"

**Enforce HTTPS Protocol:**
The attribute "http.scheme" has an insecure value of "http". 
Please ensure that the value is set to "https" for security purposes.

Please review and update the necessary attributes and naming conventions to ensure proper functionality.
```

Overall, the improvements aim to provide clearer explanations and actionable instructions to the user, making the error message more user-friendly and facilitating better understanding and resolution of the encountered issues.

jorgeepc:
This PR improves the implementation of `Analyzer Errors` by evolving the current `errors` string array to a proper struct that can handle future enhancements like suggestions, error level, expected values, etc.

```go
type Error struct {
	Error       string   `json:"error"`
	Value       string   `json:"value"`
	Expected    string   `json:"expected"`
	Level       string   `json:"level"`
	Description string   `json:"description"`
	Suggestions []string `json:"suggestions"`
}
```

## Changes

- Update the errors string array to a struct
- Improve error messages from rules
- Group error messages in the webApp

## Fixes

- fixes #2677 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom

https://www.loom.com/share/134da01378e54eae922298297f4fca9c?sid=a6484e4a-aa09-45f4-a0a9-b38df3e3b186


api/linters.yaml
```diff
@@ -94,11 +94,28 @@ components:
...
+        error:
+          type: string
+        value:
+          type: string
+        expected:
+          type: string
+        level:
+          type: string
+        description:
+          type: string
```
xoscar, line 117: Should we include a field for the documentation URL? or is it your idea that it could be part of the `suggestions` field?

web/src/constants/Analyzer.constants.ts
```diff
@@ -0,0 +1,8 @@
+export const ERROR_HEADER: Record<string, string> = {
+  attribute_naming_error: 'The following attributes do not adhere to the naming convention:',
```
xoscar, line 2: Should we generate this from the BE? this could be useful to have as part the CLI as well, so it is not duplicated

web/src/services/Span.service.ts
```diff
@@ -92,7 +92,14 @@ export type TLintBySpanContent = {
   pluginName: string;
   passed: boolean;
   spanId: string;
-  errors: string[];
+  errors: {
```
xoscar, line 95: Do we have explicitly add the type here? or can we derive it from one of the auto-generated types?

xoscar (changes requested):
Hey @jorgeepc great job! I added a couple of questions, let me know what you think

xoscar (approved):

api/linters.yaml
jorgeepc: I think we can decide that later. Now that we have the `Error` struct we can increment the fields without breaking the compatibility.

PR #2746, opened 2023-06-16
feat: add upsert method and enable it for environments

mathnogueira:
This PR adds a new method for upsert the resource into the database. And enables it for transactions

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/resourcemanager/resource_manager.go
+++ b/server/resourcemanager/resource_manager.go
@@ -137,6 +137,10 @@ func (m *manager[T]) RegisterRoutes(r *mux.Router) *mux.Router {
 		m.instrumentRoute(subrouter.HandleFunc("", m.create).Methods(http.MethodPost).Name(fmt.Sprintf("%s.Create", m.resourceTypePlural)))
 	}
 
+	if slices.Contains(enabledOps, OperationUpsert) {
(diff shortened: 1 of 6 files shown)


server/resourcemanager/resource_manager.go
```diff
@@ -235,6 +239,55 @@ func (m *manager[T]) create(w http.ResponseWriter, r *http.Request) {
...
+			created, err := m.rh.Create(r.Context(), targetResource.Spec)
+			if err != nil {
+				writeError(w, encoder, http.StatusInternalServerError, fmt.Errorf("cannot create entity: %w", err))
+				return
+			}
+
+			writeResponse(http.StatusCreated, created)
+			return
+		}
+	}
```
danielbdias, line 280: Suggestion: we can join the two to avoid nesting:
```suggestion
	if err != nil && err == sql.ErrNoRows {
		created, err := m.rh.Create(r.Context(), targetResource.Spec)
		if err != nil {
			writeError(w, encoder, http.StatusInternalServerError, fmt.Errorf("cannot create entity: %w", err))
			return
		}

		writeResponse(http.StatusCreated, created)
		return
	}
```

mathnogueira: Your suggestion made me realize I was ignoring the `err` when it was not a `sql.ErrNoRows`

danielbdias (approved):

Issue #2751, opened 2023-06-16
[Deployment Tab] Introduce CLI dynamic command


xoscar:
AC6: We will want to expose several techniques that can be used to to run a test. These will include:
- CLI
- REST
- and more in the future. Lets start with just these two

AC7: On the CLI screen, include a paragraph explaining how to use the CLI to run a test. It will be something like:
First, download the CLI via the [instructions](https://docs.tracetest.io/getting-started/installation#install-the-tracetest-cli). 
Second, download the test above and save it.
Third, execute the line below:
tracetest -s http://localhost:11633/ test run -d mytest.yaml -w

AC8: We should have a toggle switch that says 'Wait for test to complete' that you can turn on or off. If one, include the -w on the CLI line.

AC9: We should have a 'Specify hostname of Tracetest server' that can be enabled or disabled. If enabled, there should be a field asking for the host name, which should be autofilled based on the host name the browser's UI is on. Turning this on or off should alter the CLI line, removing or adding the '-s http://localhost:11633/' part of the command line.

AC10: You should be able to copy the CLI line that is being built

AC11: Ability to specify 'Use file or test id' toggle, which will alter the CLI command to either use the id or specify a file

AC12: Ability to specify how you want the output of the test: Pretty, JSON, yaml

AC13: Ability to specify whether you want Junit output by the test. If they turn this on, a textfield should default to 'result.junit'

AC14: Have a switch for 'Use current environment' that defaults to on. The CLI command should specify the environment if this is turn on.

AC15: Run CLI via Docker image? (choices yes or no, default is no). Saying yes changes the time to a docker command such as 
```sh
docker run --rm -it -v$(pwd):$(pwd) -w $(pwd) --entrypoint tracetest kubeshop/tracetest:latest -s http://host.docker.internal:11633/ test run  --definition <file-path> --wait-for-result
```

Designs:
https://www.figma.com/file/nXy4eBvpiQ3P4Jer0ogJOw/v0.11-release?type=design&t=ntanOzY9D42jZ5Op-0

PR #2764, opened 2023-06-19
feature(frontend): Automate Tab - CLI Command

xoscar:
This PR adds the CLI command controls and input for users to copy paste

## Changes

- Adds a new section to the automate tab
- Incudes controls for the command
- Updates code block component to use the same across the app

## Fixes

- #2751 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/f62bc225d38846b4af20d26d86255cd2


adnanrahic:
This is awesome!

One suggestion might be cool:

- Can we also enable naming the test definition file?
- This could be a field in the left tile right above the code block.

My reasoning:

- This would let the used give the test definition a specific name before downloading the file.
- This would make the right tile use the specific name for the test definition from the left tile.

If this is too complex, I'm not super invested in having it. It's already pretty awesome!

xoscar:
> This is awesome!
> 
> One suggestion might be cool:
> 
> * Can we also enable naming the test definition file?
> * This could be a field in the left tile right above the code block.
> 
> My reasoning:
> 
> * This would let the used give the test definition a specific name before downloading the file.
> * This would make the right tile use the specific name for the test definition from the left tile.
> 
> If this is too complex, I'm not super invested in having it. It's already pretty awesome!

@adnanrahic great suggestion, can we create an issue for this and involve @olha23 so she can think about a design for this?

web/src/components/RunDetailAutomateMethods/methods/CLICommand/SwitchControl.tsx
```diff
@@ -0,0 +1,19 @@
...
+  text: string;
+  value: boolean;
+  onChange(value: boolean): void;
+}
+
+const SwitchControl = ({value, onChange, text}: IProps) => {
+  return (
+    <S.SwitchContainer>
+      <Switch onChange={onChange} checked={value} />
+      {text}
```
jorgeepc, line 14: Minor: here you can add an `id` to the switch component and a `<label htmlFor="id">{text}</label>`. With this you can use the switch by clicking on the text.

jorgeepc (approved):
I like this 🔥 A couple of comments/questions.

- I think we should have a `Copy` cmd button. It's not that clear that you have to click the code block to copy it.
- Do we support `yaml` as an output for test run?

PR #2817, opened 2023-06-23
feat(server): add yaml stream encoding for resource manager

schoren:
This PR adds support for multi-document YAML requests and responses in the `resourcemanager` package. Previously, it supported single-document YAML.

The implementations uses a new encoder that can be used with the `text/yaml-stream` content-type/accept headers.

With this change, we can handle **requests** and **responses** that contain multiple YAML documents separated by `---` lines.
This means that a client can send a multi document request and it will work, but only if all the documents are of the same type (i.e. all transactions, or all tests, but not a mix of transactions and tests).

See it in action:
https://www.loom.com/share/dee1b1c35a544ee3b31bd00f1058ab1c

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/resourcemanager/encoding.go
+++ b/server/resourcemanager/encoding.go
@@ -50,7 +50,7 @@ func (e yamlStreamEncoder) Accepts(contentType string) bool {
 }
 
 func (e yamlStreamEncoder) Marshal(in interface{}) (out []byte, err error) {
-	targetField, err := getYamlstreamField(in, "items", reflect.Slice)
+	targetField, err := getYamlStreamField(in, "items", reflect.Slice)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


server/resourcemanager/encoding.go
```diff
@@ -35,6 +39,111 @@ var yamlEncoder = basicEncoder{
...
+	if countField.IsValid() {
+		countField.SetInt(int64(targetField.Len()))
+	}
+
+	return nil
+}
+
+// getYamlstreamField returns the field in `target` with the name as the value of its yamlstream tag.
+// it returns an error if the field is not found or if the field is not of the specified kind.
+func getYamlstreamField(target interface{}, name string, kind reflect.Kind) (reflect.Value, error) {
```
danielbdias, line 121:
```suggestion
func getYamlStreamField(target interface{}, name string, kind reflect.Kind) (reflect.Value, error) {
```

server/resourcemanager/encoding_test.go
```diff
@@ -149,3 +149,89 @@ name: example
 	})
 
 }
+
+func TestYamlstreamEncoding(t *testing.T) {
```
danielbdias, line 153:
```suggestion
func TestYamlStreamEncoding(t *testing.T) {
```

danielbdias (approved):
Great PR! I added two small suggestions, but they are not important for now.

Issue #2768, opened 2023-06-20
[Automate Tab] Support Test Definition Name Update


xoscar:
Based on @adnanrahic's comment:

```
One suggestion might be cool:

Can we also enable naming the test definition file?
This could be a field in the left tile right above the code block.

My reasoning:

This would let the used give the test definition a specific name before downloading the file.
This would make the right tile use the specific name for the test definition from the left tile.
If this is too complex, I'm not super invested in having it. It's already pretty awesome!
```

Update the CLI command technique to support test definition name changes

#### Acceptance Criteria

**AC1**
As a user looking at the CLI command automate technique
I should be able to update the definition file name that will be used for the command

xoscar:
@kdhamric @olha23 could you take a look at this?

kdhamric:
I think I know what you are suggesting, but not exactly. If you and @adnanrahic think it would be an improvement, do it.

olha23:
@xoscar I’m not sure that i understand what i need to do here. Something like this? https://www.figma.com/file/nXy4eBvpiQ3P4Jer0ogJOw/v0.11-release?type=design&node-id=250%3A13639&t=ciXYmunOH3EMEgOt-1

adnanrahic:
Added my 2 cents in the Figma design.

olha23:
updated

PR #2830, opened 2023-06-27
chore(frotend): Test Definition Name Input

xoscar:
This PR adds a way for users to update the test definition file name and it updates the CLI command accordingly

## Changes

- Adds InputOverlay component
- Adds logic to carry over the file name across the page
- Updates CLI command and download file name from users input

## Fixes

- #2768 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/cce3af0532af4b09801aa7a23ae2cde1

### DIFF
--- a/web/src/components/RunDetailAutomate/RunDetailAutomate.tsx
+++ b/web/src/components/RunDetailAutomate/RunDetailAutomate.tsx
@@ -1,7 +1,9 @@
+import {snakeCase} from 'lodash';
 import Test from 'models/Test.model';
+import {useState} from 'react';
 import TestRun from 'models/TestRun.model';
 import * as S from './RunDetailAutomate.styled';
-import TestDefinition from '../RunDetailAutomateDefinition';
+import RunDetailAutomateDefinition from '../RunDetailAutomateDefinition';
 import RunDetailAutomateMethods from '../RunDetailAutomateMethods/RunDetailAutomateMethods';
 
 interface IProps {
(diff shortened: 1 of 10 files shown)


web/src/components/RunDetailAutomate/RunDetailAutomate.tsx
```diff
@@ -1,23 +1,29 @@
...
-    <S.SectionLeft>
-      <TestDefinition test={test} />
-    </S.SectionLeft>
-    <S.SectionRight>
-      <RunDetailAutomateMethods test={test} run={run} />
-    </S.SectionRight>
-  </S.Container>
-);
+const RunDetailAutomate = ({test, run}: IProps) => {
+  const [fileName, setFileName] = useState<string>(`${snakeCase(test.name)}.yaml`);
```
xoscar, line 15: I went back and forth on having this to be a local state or something from a provider, but to keep it simple I just add it as a state. If we need to carry over multiple things for the techniques we can have an automate provider instead.

adnanrahic:
This is beautiful!!!

web/src/components/RunDetailAutomate/RunDetailAutomate.tsx
jorgeepc: I like it. Keep it simple.

jorgeepc (approved):

PR #2832, opened 2023-06-28
feat(cli): dynamic list of available resources

schoren:
This PR updates the setup of commands using the new resourcemanager client to use a dynamically generated  list of available (registered) resources for the help message. This removes the need to manually maintain a separated list.


jorgeepc (approved):

cli/cmd/config.go
```diff
@@ -70,138 +200,8 @@ func setupCommand(options ...setupOption) func(cmd *cobra.Command, args []string
...
-							} else {
-								item.SetP(date.Format(time.DateTime), "spec.summary.lastRun.time")
-							}
-						}
-					},
-				},
-			),
-		)
+		hc := resourcemanager.NewHTTPClient(cliConfig.URL(), extraHeaders)
+		*httpClient = *hc
```
danielbdias, line 204: Question: is there a reason to do this pointer trick instead of doing something like this:
```suggestion
		httpClient = &resourcemanager.NewHTTPClient(cliConfig.URL(), extraHeaders)
```

cli/cmd/get_cmd.go
```diff
@@ -2,47 +2,48 @@ package cmd
...
 
 func init() {
+	getCmd = &cobra.Command{
+		GroupID: cmdGroupResources.ID,
+		Use:     "get " + resourceList(),
+		Short:   "Get resource",
+		Long:    "Get a resource from your Tracetest server",
+		PreRun:  setupCommand(),
+		Run: WithResourceMiddleware(func(_ *cobra.Command, args []string) (string, error) {
+			resourceType := args[0]
```
danielbdias, line 24: Can we get a panic here if the user runs the command just as `tracetest get` ?
Should we validate the `args` size and print the help on this case?

cli/cmd/list_cmd.go
```diff
@@ -2,55 +2,56 @@ package cmd
...
 func init() {
+	listCmd = &cobra.Command{
+		GroupID: cmdGroupResources.ID,
+		Use:     "list " + resourceList(),
+		Short:   "List resources",
+		Long:    "List resources from your Tracetest server",
+		PreRun:  setupCommand(),
+		Run: WithResourceMiddleware(func(_ *cobra.Command, args []string) (string, error) {
+			resourceType := args[0]
+			ctx := context.Background()
```
danielbdias, line 25: Same question as I did on `get`. Can we have a panic here if the user executes only `tracetest list`?

danielbdias (approved):

cli/cmd/get_cmd.go
schoren: [WithResourceMiddleware](https://github.com/kubeshop/tracetest/blob/f112ab7e887afb558435116cc6017337d9a59e23/cli/cmd/middleware.go#L65) validates that the resource is set before getting to this function. Here's an example of running `list` without passing a resource: 

Screenshot 2023-06-28 at 15 34 46

cli/cmd/config.go
schoren: Yes. In order to have this dynamic list available, we need to register all resources before the commands are registered. To do this, the resources are registered as a package variable [here](https://github.com/kubeshop/tracetest/pull/2832/files#diff-7e99687b5d846708c7b5a928628f9646f5c6f4bd52c231a75a46b78b09ca57adR53), and the commands are registered within `init` functions. This guarantees that the registry will be ready before the commands.

The problem with this is that the resources depened on the `resourcemanager.HTTPClient`, which in turns depends on the config being ready. The config gets ready in the `setupCommand` function, which is called as a `preRun` of the commands. So in order to configure a command, you need to have already configured the command. This is a ciruclar reference initialization, and that's not allowed.

The solution I found was to setup the registry with a pointer to an empty `HTTPClient`, and when the `setupCommand` runs on the `preRun` stage of the command, it replace the pointer with the correctly configured client.

Does this make sense?

cli/cmd/list_cmd.go
schoren: same answer as on `get`. this ges validated on the middleware before getting here

cli/cmd/config.go
danielbdias: Yes, thanks!

danielbdias: Just one detail: can we add a small comment on the code explaining that? Maybe just a link to you explanation on the PR

cli/cmd/get_cmd.go
danielbdias: It makes sense. My fear about accessing indexes in languages that throw out-of-index errors and friends is that sooner or later, we can receive that error (usually on dev mode when we coded something wrong by mistake xD).

One thing that we can think about in the future is that since the `WithResourceMiddleware` is tailored to the resources API, we could unpack the first argument and change the interface to `func(_ *cobra.Command, resourceType string, args []string) (string, error)` and allow `args` to have only the args for that command in fact.

schoren: > sooner or later, we can receive that error
This is true. Assuming that the middleware will always work perfectly might not be the safest approach. This is actually being parsed to the `resourceParams.ResourceName` property, so I replaced the calls to `args[0]` to `resourceParams.ResourceName`.

What do you think @danielbdias? It doesn't guarantee success, but at least it doesn't panic

cli/cmd/config.go
schoren: Good idea. Added!

cli/cmd/get_cmd.go
danielbdias: It sounds good to me!

PR #2829, opened 2023-06-26
feat(cli): refactor list formatter for better resource manager support

schoren:
This PR is a technical refactor that allows leveraging the server's resourcemanager formatting capabilities. 

Currently, this PR only updates the LIST cli verb. Meaning that `tracetest list [any resource]` will use this new code. However, the rest of the verbs are unmodified. They'll be updated in following PRs.

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

PR #2841, opened 2023-06-30
feat(cli): update apply command with new resourcemanager client

schoren:
This PR follows up on https://github.com/kubeshop/tracetest/pull/2829, adding support for the Apply verb in the same manner.

It also adds a feature in the server resource manager so that it returns 405 method not allwed on resources that did not enable the requested endpoint. This removes the need for the CLI to know what resource implements each method


testing/cli-e2etest/environment/jaeger/resources/data-store.yaml
```diff
@@ -4,7 +4,6 @@ spec:
   name: jaeger
   type: jaeger
   default: true
-  createdAt: 2023-05-09T18:22:34.840021Z
```
schoren, line 7: This, and all similar updates within the cli e2e yaml files are because the new apply implementation doesn't override the file. it just injects the ID if needed. This means 2 things;

1. no more commits to these files because of running tests locally and changing create dates,
2. the original file format is now preserved, including comments and other user generated formattings

cli/cmd/apply_cmd.go
```diff
@@ -2,46 +2,35 @@ package cmd
...
+	"github.com/kubeshop/tracetest/cli/pkg/resourcemanager"
 	"github.com/spf13/cobra"
 )
 
 var applyParams = &parameters.ApplyParams{}
 
 var applyCmd = &cobra.Command{
 	GroupID: cmdGroupResources.ID,
-	Use:     fmt.Sprintf("apply %s", strings.Join(parameters.ValidResources, "|")),
+	Use:     "apply " + resourceList(),
```
schoren, line 15: this list is now dynamic, it doesn't need to be manually maintained

cli/pkg/fileutil/file.go
```diff
@@ -0,0 +1,99 @@
+package fileutil
```
schoren, line 1: This new `fileutil` package is based from the old `file` package, but it's more limited in scope. It only focuses on handling the ID injection on applied files

mathnogueira (approved):

cli/pkg/resourcemanager/client.go
```diff
@@ -28,14 +28,23 @@ type HTTPClient struct {
...
+			},
+		},
 		baseURL:      baseURL,
 		extraHeaders: extraHeaders,
 	}
 }
 
 func (c HTTPClient) url(resourceName string, extra ...string) *url.URL {
-	url, _ := url.Parse(fmt.Sprintf("%s/api/%s/%s", c.baseURL, resourceName, strings.Join(extra, "/")))
+	urlStr := c.baseURL + path.Join("/api", resourceName, strings.Join(extra, "/"))
```
schoren, line 46: this changes remove a trailing `/` that we were adding to the urls without id, like `GET /transactions/` instead of `GET /transactions`. This was causing a bug. See https://github.com/kubeshop/tracetest/pull/2841/files#diff-09cf25bcafe115bcabac1fb6c58cce0a7a7b1b74536c4927f530947e659cda92R32

cli/pkg/resourcemanager/delete.go
```diff
@@ -4,15 +4,12 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 )
 
 const VerbDelete Verb = "delete"
 
 func (c client) Delete(ctx context.Context, id string, format Format) (string, error) {
-	if c.deleteSuccessMsg == "" {
```
schoren, line 12: each command support is decided by the server. we try to send the request, and if the method is not enabled, we get a `405 method not allowed`. This also works for any methods now. This effectively moves and encapsulates the knowledge of what verb supports each resource on the server

cli/pkg/resourcemanager/delete.go
```diff
@@ -40,5 +37,13 @@ func (c client) Delete(ctx context.Context, id string, format Format) (string, e
 		return "", fmt.Errorf("could not Delete resource: %w", err)
 	}
 
-	return c.deleteSuccessMsg, nil
+	msg := ""
+	if c.deleteSuccessMsg != "" {
```
schoren, line 41: we still need to override the default message in some cases

server/config/config_repository.go
```diff
@@ -114,6 +107,11 @@ func (*Repository) SortingFields() []string {
 	return []string{"name"}
 }
 
+func (r *Repository) Create(ctx context.Context, updated Config) (Config, error) {
```
schoren, line 110: requried for resourcemanager `upsert` to be enabled

server/datastore/datastore_repository.go
```diff
@@ -71,6 +63,11 @@ func (r *Repository) getCreatedAt(ctx context.Context, dataStore DataStore) (str
 	return oldDataStore.CreatedAt, nil
 }
 
+func (r *Repository) Create(ctx context.Context, updated DataStore) (DataStore, error) {
```
schoren, line 66: requried for resourcemanager `upsert` to be enabled

server/executor/pollingprofile/polling_profile_repository.go
```diff
@@ -23,6 +23,11 @@ func (r *Repository) SetID(profile PollingProfile, id id.ID) PollingProfile {
 	return profile
 }
 
+func (r *Repository) Create(ctx context.Context, updated PollingProfile) (PollingProfile, error) {
```
schoren, line 26: requried for resourcemanager `upsert` to be enabled

server/linter/analyzer/analyzer_repository.go
```diff
@@ -63,6 +63,11 @@ func (r *Repository) SetID(linters Linter, id id.ID) Linter {
 	return linters
 }
 
+func (r *Repository) Create(ctx context.Context, linter Linter) (Linter, error) {
```
schoren, line 66: requried for resourcemanager `upsert` to be enabled

server/resourcemanager/operations.go
```diff
@@ -14,7 +14,6 @@ const (
 	OperationNoop   Operation = ""
 	OperationList   Operation = "list"
 	OperationCreate Operation = "create"
-	OperationUpsert Operation = "upsert"
```
schoren, line 17: if update and create and enabled, upsert is enabled automatically

server/resourcemanager/resource_manager.go
```diff
@@ -129,29 +143,41 @@ func (m *manager[T]) RegisterRoutes(r *mux.Router) *mux.Router {
 
 	enabledOps := m.EnabledOperations()
 
+	listHandler := m.methodNotAllowed
```
schoren, line 146: we used to not register unsupported routes. This caused that requests that should be handled by resourcemanager went directly to the SPA handler, making the response be the main `index.html` contents from the web directory.

Now, we always register all routes, but return a nice `405 method not allowed` for disabled methods

server/resourcemanager/resource_manager.go
```diff
@@ -249,46 +282,27 @@ func (m *manager[T]) upsert(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// if there's no ID given, create the resource
```
schoren, line 285: `upsert` was rewriting code already available in the `create` and `update` methods, so a little refactoring reduced code duplication

Issue #2773, opened 2023-06-21
[Analyzer] Configuration - Backend


xoscar:
As part of the configuration changes, we need to allow users to update different sections of the analyzer, which include:
- Plugin enabled/disabled status
- Rule error levels
- Rule weights for score calculation

As well as to use the configurations when executing the analyzer step for each of the test runs

#### Acceptance Criteria
**AC1**
As a user configuring the analyzer
I should be able to configure which plugins I want to enable for my test runs

**AC2**
As a user configuring the analyzer
I should be able to set the critical level for each of the rules within a plugin
and the options should be `error`, `warning`, `skipped`

**AC3**
As a user configuring the analyzer
I should be able to set the weight for each of the rules within a plugin
Having the sum of the weights to be 100%

**AC4**
As a user executing a test run
And I see the analyzer results
I should only see the results from the plugins I have enabled

**AC5**
As a user executing a test run
And I see the analyzer results
The global score should be calculated based on the enabled plugins and the rules with `error` or `warning` critical levels

**AC6**
As a user executing a test run
And I see the analyzer results
I should see the critical level result for each of the rules based on the config

```[tasklist]
### Tasks
- [ ] Update OpenAPI specs to reflect the new contract
- [ ] Update Analyzer resource to support the new configuration
- [ ] Update Analyzer result model to persist the new list of fields
- [ ] Implement Runner changes to use the new configuration values
```

PR #2842, opened 2023-06-30
feature(backend): Enabling Analyzer Configuration

xoscar:
This PR updates the architecture for plugins and rules, keeping configuring plugins more straightforward and flexible.

## Changes

- Decouples rules from plugins and allows for a configuration-based setup
- Moves the `linter` out of the `models` to the `linter` directory
- Updates plugins and rules interfaces to support an analyzer configuration
- Moves metadata away from linter execution, leaving it to be config only
- Enables configuration for `error level` and `weight` for rules

## Fixes

- #2773 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

Analyzer Arch Changes
https://www.loom.com/share/03e95872d383496aa81591be9d772123

Features & bug fixes
https://www.loom.com/share/0517b986f19941f3b41c5187d2344500


server/executor/linter_runner.go
```diff
@@ -100,12 +100,11 @@ func (e *defaultlinterRunner) startWorker() {
 func (e *defaultlinterRunner) onRequest(request LinterRequest) {
 	ctx := request.Context()
 	lintResource := e.analyzerGetter.GetDefault(ctx)
-	linter := linter.NewLinter(lintResource, linter.AvailablePlugins...)
 
-	shouldSkip, reason := linter.ShouldSkip()
+	shouldSkip := lintResource.ShouldSkip()
```
xoscar, line 104: The analyzer resource should make the decisions of running or not the linter as it contains the configuration

server/linter/analyzer/analyzer_entities.go
```diff
@@ -28,9 +30,25 @@ type (
...
+		Id      string       `json:"id"`
+		Enabled bool         `json:"enabled"`
+		Rules   []LinterRule `json:"rules"`
+
+		// internal fields
+		Name        string `json:"-"`
+		Description string `json:"-"`
+	}
+
+	LinterRule struct {
```
xoscar, line 42: This a new block, allowing users to configure rules

server/linter/analyzer/analyzer_entities.go
```diff
@@ -40,18 +58,130 @@ func (l Linter) Validate() error {
 	}
 
 	for _, p := range l.Plugins {
-		if p.Name == "" {
-			return fmt.Errorf("plugin name cannot be empty")
+		plugin, ok := findPlugin(p.Id, DefaultPlugins)
```
xoscar, line 61: Validates that the plugins and rules coming from the new resource instance match what we have in the system, this should be more flexible in the future when we implement custom analyzer profiles

server/linter/analyzer/analyzer_entities.go
```diff
@@ -40,18 +58,130 @@ func (l Linter) Validate() error {
...
+	}
+
+	return plugins
+}
+
+func (l Linter) ShouldSkip() bool {
+	return !l.Enabled
+}
+
+func (l Linter) WithMetadata() (Linter, error) {
```
xoscar, line 142: Combines the analyzer configuration (Database, CLI, FE) with the metadata stored within the repo, as it is not available for users to update (today)

server/linter/analyzer/analyzer_metadata.go
```diff
@@ -0,0 +1,139 @@
+package analyzer
+
+var (
+	// plugins
+	StandardsId = "standards"
```
xoscar, line 5: Names, ids, descriptions, etc

server/linter/linter.go
```diff
@@ -2,107 +2,81 @@ package linter
...
 	"github.com/kubeshop/tracetest/server/model"
 )
 
 var (
-	AvailablePlugins = []model.Plugin{
-		linter_plugin_standards.NewStandardsPlugin(),
-		linter_plugin_security.NewSecurityPlugin(),
-		linter_plugin_common.NewCommonPlugin(),
-	}
+	commonPlugin = plugins.NewPlugin(
```
xoscar, line 13: Static registration of the linter configuration, we do not have much flexibility today, but in the future, the linter runner could generate a dynamic plugin registry with custom rules and plugins defined by user profile, test or test run levels

server/linter/plugins/plugins_entities.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (p BasePlugin) Id() string {
+	return p.id
+}
+
+func (p BasePlugin) RuleRegistry() rules.RuleRegistry {
+	return p.ruleRegistry
+}
+
+func (p BasePlugin) Execute(ctx context.Context, trace model.Trace, config analyzer.LinterPlugin) (analyzer.PluginResult, error) {
```
xoscar, line 28: All current plugins have the same execution logic, so instead of having to create a new `plugin` file for each, we use the default, and if needed, we can create specific plugin files implementing the `Plugin` interface to override execute method with custom logic

server/linter/rules/rules_entities.go
```diff
@@ -0,0 +1,27 @@
...
+
+func NewRule(id string) BaseRule {
+	return BaseRule{id}
+}
+
+func (r BaseRule) Id() string {
+	return r.id
+}
+
+func (r BaseRule) Evaluate(ctx context.Context, trace model.Trace, config analyzer.LinterRule) (analyzer.RuleResult, error) {
```
xoscar, line 25: Rules on the other hand, have very custom logic, which makes it more difficult to encapsulate in a single default method, for these, we should create an implementation of the `Rule` interface per rule

server/resourcemanager/resource_manager.go
```diff
@@ -222,6 +222,15 @@ func (m *manager[T]) create(w http.ResponseWriter, r *http.Request) {
 		targetResource.Spec = m.rh.SetID(targetResource.Spec, m.config.idgen())
 	}
 
+	if err := targetResource.Spec.Validate(); err != nil {
```
xoscar, line 225: I added this validate step to the resource manager, as it wasn't currently in use, not sure if I should better move this to another PR

danielbdias (approved):
The code overall is ok! Good job!

One thing to consider in the future is how we can test these rules to guarantee that they are working as intended and are only sensitive to changes in the future without us knowing it.

xoscar:
> The code overall is ok! Good job!
> 
> One thing to consider in the future is how we can test these rules to guarantee that they are working as intended and are only sensitive to changes in the future without us knowing it.

Hey @danielbdias thanks for the review, I have added tests for the analyzer and linter critical points, and will add more for rules in the following PRs!

server/linter/rules/not_empty_attribute_rule.go
```diff
@@ -0,0 +1,56 @@
...
+				passed = false
+			}
+
+			res = append(res, analyzer.Result{
+				SpanID: span.ID.String(),
+				Passed: len(emptyAttributes) == 0,
+				Errors: errors,
+			})
+		}
+	}
```
schoren, line 53: Looks like there's a dobule loop that could be avoided here
```suggestion
			errors := make([]analyzer.Error, 0, len(emptyAttributes))
			for name, value := range span.Attributes {
				if value == "" {
					errors = append(errors, analyzer.Error{
						Value:       emptyAttribute,
						Description: fmt.Sprintf(`Attribute "%s" is empty`, emptyAttribute),
					})
				}
			}

			if len(errors) > 0 {
				passed = false
			}

			res = append(res, analyzer.Result{
				SpanID: span.ID.String(),
				Passed: passed,
				Errors: errors,
			})
		}
	}
```

server/linter/rules/required_attribute_rule.go
```diff
@@ -0,0 +1,37 @@
...
+	if config.ErrorLevel != analyzer.ErrorLevelDisabled {
+		for _, span := range trace.Flat {
+			res = append(res, r.validateSpan(span))
+		}
+
+		for _, result := range res {
+			if !result.Passed {
+				allPassed = false
+			}
+		}
```
schoren, line 33: same here

```suggestion
		for _, span := range trace.Flat {
			analyzerResult := r.validateSpan(span)
			res = append(res, analyzerResult)
			if !analyzerResult.Passed {
				allPassed = false
			}

		}
```

server/linter/rules/rules_entities.go
schoren: I would say there's no need for this `BaseRule`. it can be replaced by a `func (r MyRule) ID() string {return someconstantID}` func, satisfying the interface. This also makes the ID immutable, and that's good in this context

server/linter/rules/rules_registry.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/kubeshop/tracetest/server/linter/analyzer"
+	"github.com/kubeshop/tracetest/server/model"
+)
+
+type Rule interface {
+	Id() string
+	Evaluate(context.Context, model.Trace, analyzer.LinterRule) (analyzer.RuleResult, error)
+}
+
+type RuleRegistry interface {
```
schoren, line 16: Who depends on this interface? can the dependees directly reference a struct? What benefits do we get from this interface?

If a client depends on this interface, it will be requiring this package. Given that we only support one implementation of the `RuleRegistry`, and it's in memory without any dependencies, doesn't look like we would need to mock it, for example, or replace it by another implementation.

server/resourcemanager/resource_manager.go
```diff
@@ -255,6 +255,15 @@ func (m *manager[T]) doCreate(w http.ResponseWriter, r *http.Request, encoder En
 		specs = m.rh.SetID(specs, m.config.idgen())
 	}
 
+	if err := specs.Validate(); err != nil {
```
schoren, line 258: niceeeee

server/resourcemanager/resource_manager.go
```diff
@@ -288,6 +297,15 @@ func (m *manager[T]) upsert(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if err := targetResource.Spec.Validate(); err != nil {
```
schoren, line 300: this is already handled on `doCreate`, and if you move [this](https://github.com/kubeshop/tracetest/pull/2842/files#diff-73567862de3c0b65102c45d3df1f98a49f2649fe588140f55a56fa693068bf51R351) to happen within the `doUpdate` method, we can have both cases covered and remove the need for extra work here

schoren (changes requested):
It looks good in general. I think we should try to embrace go's naming conventions, so it'd be great to rename all the `Id`s to `ID`.

I outlined a couple of duplicated loops and bugs, not necessarily related specifically to this work, but since we're here 🤷 

Finally, more oriented to the architecutre:
1. I would remove the `BaseRule`, it's being used kind of as a base class, but it doesn't make a lot of sense here IMO.
2. We should try to reduce the amount of interfaces that are not strictly required. "Accept interfaces, return structs". You can leverage go's duck typing and declare interfaces where you want to avoid direct references to a package.
3.

server/linter/rules/rules_registry.go
xoscar: Makes sense, I think the best case is just to remove the interface

xoscar:
Hey @schoren thanks for the review it was really insightful! I think I addressed all of your comments and updated the code accordingly, let me know if there is something else I could do

schoren (approved):
Thanks @xoscar! this looks very consistent now. I would strongly suggest to remove the remaining duplicated loops ([here](https://github.com/kubeshop/tracetest/pull/2842/files/e7e3d02e5effa551f21fd12021d9edae10ca7ede#r1252418244) and [here](https://github.com/kubeshop/tracetest/pull/2842/files/e7e3d02e5effa551f21fd12021d9edae10ca7ede#r1252418797)) and remove the `BaseRule`, since it's now unused (I think?)

Other than that, LGTM

Issue #2774, opened 2023-06-21
[Analyzer] Configuration - Frontend


xoscar:
As part of the configuration changes, we need to allow users to update different sections of the analyzer, which include:
- Plugin enabled/disabled status
- Rule error levels
- Rule weights for score calculation

As well as to use the configurations when executing the analyzer step for each of the test runs

#### Acceptance Criteria
**AC1**
As a user configuring the analyzer
I should be able to configure which plugins I want to enable for my test runs

**AC2**
As a user configuring the analyzer
I should be able to set the critical level for each of the rules within a plugin
and the options should be `error`, `warning`, `skipped`

**AC3**
As a user configuring the analyzer
I should be able to set the weight for each of the rules within a plugin
Having the sum of the weights to be 100%

**AC4**
As a user executing a test run
And I see the analyzer results
I should only see the results from the plugins I have enabled

**AC5**
As a user executing a test run
And I see the analyzer results
The global score should be calculated based on the enabled plugins and the rules with `error` or `warning` critical levels

**AC6**
As a user executing a test run
And I see the analyzer results
I should see the critical level result for each of the rules based on the config

```[tasklist]
### Tasks
- [ ] Implement new contract based on BE changes
- [ ] Update Analyzer section to display rule critical status
- [ ] Update Analyzer settings form to match new configuration
```

xoscar:
@olha23 do you mind taking a look at this?

olha23:
@xoscar do we need to show general Minimum score in settings?

olha23:
@xoscar can you check this please https://www.figma.com/file/nXy4eBvpiQ3P4Jer0ogJOw/v0.11-release?type=design&node-id=530%3A14948&mode=design&t=zwXpU1uVjzHXlJSE-1


and I’m not sure what i need to do with analyzer results

olha23:
@xoscar i updated mockups based on our conversation https://www.figma.com/file/nXy4eBvpiQ3P4Jer0ogJOw/v0.11-release?type=design&node-id=530%3A14948&mode=design&t=lIXWdTgS9EnSW0Bg-1

PR #2850, opened 2023-07-04
feature(frontend): Analyzer Configuration

xoscar:
This PR adds the analyzer configuration changes for the frontend, including the settings page and analyzer results

## Changes

- Includes the new configuration fields as part of the settings page
- Updates the analyzer results section with the configured error levels and weights

## Fixes

- #2774 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/f7b4e4e7fe1f46ada25ac3cda543c40e


web/src/components/AnalyzerResult/GlobalResult.tsx
```diff
@@ -32,14 +32,12 @@ const GlobalResult = ({score, minimumScore}: IProps) => {
           <Percentage score={score} />
         </S.GlobalScoreContainer>
       </S.GlobalScoreWrapper>
-      {!!minimumScore && (
```
xoscar, line 35: We should be always displaying this as per Olly's feedback

jorgeepc:
If you disable a plugin, this is not present in the AnalyzerResults section. However, for disabled rules they are present in the AnalyzerResults section. Should we keep the same behavior for disabled plugins and rules?

jorgeepc:
A couple of findings:
- I'm seeing an error when trying to save the analyzer resource:
`"cannot parse body: json: cannot unmarshal string into Go struct field LinterRule.spec.plugins.rules.weight of type int"`
- This component has a weird UI:
Screenshot 2023-07-04 at 14 55 11
Screenshot 2023-07-04 at 14 55 23

schoren (approved):

PR #2851, opened 2023-07-04
feat(frontend): add select as current span in timeline

jorgeepc:
This PR adds the `select as current span` feature to the Timeline visualization. We already had this feature as part of the Graph visualization, but we were missing it in our Timeline. We are refactoring a couple of components to be able to reuse the logic in both places.

## Changes

- refactor select as current span components and hook
- implement select as current in timeline visualization

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom

https://www.loom.com/share/808d0d3bfdb147f78f626014a74a93f6?sid=c177ff95-5784-452b-92e7-28c340892bce


xoscar (approved):
Pretty clean 🧼 !

PR #2853, opened 2023-07-04
Cli test run

schoren:
This PR fixes the run command. it uses the new resource manager client

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


cli/actions/export_test_action.go
```diff
@@ -69,13 +69,14 @@ func (a exportTestAction) getDefinitionFromServer(ctx context.Context, args Expo
...
-	}
+	// request := a.client.ApiApi.GetTestVersionDefinitionFile(ctx, args.TestId, expectedVersion)
+	// definitionString, _, err := a.client.ApiApi.GetTestVersionDefinitionFileExecute(request)
+	// if err != nil {
+	// 	return "", fmt.Errorf("could not get test definition: %w", err)
+	// }
 
-	return definitionString, nil
+	// return definitionString, nil
+	return "", fmt.Errorf("deprecated")
```
danielbdias, line 79: Are we sending any message to the user saying to use `get` instead of `export` now?

cli/pkg/fileutil/path.go
```diff
@@ -3,8 +3,33 @@ package fileutil
...
+func RelativeTo(path, relativeTo string) string {
+	if filepath.IsAbs(path) {
+		return path
+	}
+
+	return filepath.Join(relativeTo, path)
+}
+
+func LooksLikeFilePath(path string) bool {
+	return strings.HasPrefix(path, "./") ||
```
danielbdias, line 27: Being a devil's advocate here, this method will fail for Windows if a user uses a full path, no?

We could have a transaction step as `C:/my-folder/my-test.yaml`

server/http/controller.go
```diff
@@ -646,6 +647,7 @@ func (c *controller) GetTransactionVersion(ctx context.Context, tID string, vers
 func (c *controller) RunTransaction(ctx context.Context, transactionID string, runInformation openapi.RunInformation) (openapi.ImplResponse, error) {
 	transaction, err := c.transactionRepository.GetAugmented(ctx, id.ID(transactionID))
 	if err != nil {
+		fmt.Println("*-************* 1")
```
danielbdias, line 650: 🙃 😂 
```js
console.log("estou aqui")
console.log("estoy aca")
console.log("I'm here!")
```

danielbdias (approved):

cli/actions/export_test_action.go
schoren: no, this will be fixed in a follow up PR

server/http/controller.go
schoren: xD

cli/pkg/fileutil/path.go
schoren: You are right, I didn't think of that case. This is the current behaviour I think, so I'll keep it as is now. If the case happens, we'll figure out how to fix it :P

PR #2884, opened 2023-07-07
fix: don't update transaction file other than adding id to it

mathnogueira:
This PR prevents the CLI from updating the transaction file from being updated with the pre-processed file content.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
LGM!

Issue #2776, opened 2023-06-21
[Analyzer] Global Configuration - Test Run Failure Gates - Backend


xoscar:
As part of the changes to support a specific failure point for test runs, we need to update the backend to support these changes.

#### Acceptance Criteria

**AC1**
As a user configuring the server
I should be able to set the global gates for test runs
Which should be a list of any of the following
- test specs
- analyzer score
- analyzer rule

**AC2**
As a user executing a test
I should be able to attach the gates to the request
Which overrides the global configuration

**AC3**
As a user that executed a test
I should see the configuration of the gates used for the run
And the result of such calculation passed/failed

**AC4**
As a user that executed a test
The test run execution should not stop if any of the gates are marked as failed

**AC5**
As a user that executed a test
The test run should include the information about the client source of the test trigger

```[tasklist]
### Tasks
- [ ] Update the openAPI specs to support failure point and result
- [ ] Update Configuration resource to support failure point
- [ ] Allow failure point config during the run test request
- [ ] Calculate failure result based on the configuration/request
- [ ] Persist failure configuration and result at the test run level
- [ ] Persist the source of the test run (CLI/FE/REST) at the test run level
```

PR #2888, opened 2023-07-07
feature(backend): Required Gates

xoscar:
This PR adds the capability for the server to handle required gates for test runs. 

Required gates are a list of checkpoints where the test runner can be configured to mark the execution as failed or passed, without stopping the state machine steps.

## Changes

- Adds test runner resource
- Includes required gates logic for the test runner
- Adds migrations for the database
- Includes mappings and persistence of data for test runs

## Fixes

- #2776 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

https://www.loom.com/share/2d16fb9a483e4becb75832a2e591d10c


api/testRunner.yaml
```diff
@@ -0,0 +1,51 @@
+openapi: 3.0.0
```
xoscar, line 1: I went with the proposed option @danielbdias suggested of creating a new test runner resource, I think this will allow us to have a more flexible way of dealing with customizable options when executing a test.

server/executor/runner.go
```diff
@@ -152,6 +160,14 @@ func (r persistentRunner) Run(ctx context.Context, testObj test.Test, metadata t
 
 	r.listenForStopRequests(ctx, cancelCtx, run)
 
+	// configuring required gates
+	if requiredGates != nil {
```
xoscar, line 164: Required gates can be overwritten by the clients when executing a test run and if it's not part of the request we fallback to the global configuration

server/test/run.go
```diff
@@ -159,9 +165,44 @@ func (r Run) SuccessfulLinterExecution(linter analyzer.LinterResult) Run {
...
+	}
+
+	if r.Linter.Score < linter.MinimumScore {
+		r = r.RequiredGateFailed(testrunner.RequiredGateAnalyzerScore)
+	}
+
 	return r
 }
 
+func (r Run) ConfigureRequiredGates(gates []testrunner.RequiredGate) Run {
```
xoscar, line 179: This logic is to support side effects from the linter and analyzer results to update the gates.
It also includes logic to generate the results from a test run without it for retro-compat

server/executor/runner.go
```diff
@@ -31,14 +32,18 @@ type RunResult struct {
 }
 
 type Runner interface {
-	Run(context.Context, test.Test, test.RunMetadata, environment.Environment) test.Run
+	Run(context.Context, test.Test, test.RunMetadata, environment.Environment, *testrunner.RequiredGates) test.Run
```
xoscar, line 35: I was thinking of moving both the env and test runner gates to an options struct, but I think this will be something that will change soon as part of the event-driven logic we might be applying for the next big September update 👁️

server/app/app.go
```diff
@@ -198,6 +199,7 @@ func (app *App) Start(opts ...appOption) error {
 	linterRepo := analyzer.NewRepository(db)
 	testRepo := test.NewRepository(db)
 	runRepo := test.NewRunRepository(db)
+	trRepo := testrunner.NewRepository(db)
```
jorgeepc, line 202: minor: just for consistency with other resource's repo I would call it `testRunnerRepo`

server/migrations/27_add_test_runner.down.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+DROP TABLE "test_runners";
```
jorgeepc, line 3: I think the table could be just `test_runner`, what do you think?

jorgeepc (approved):
Looks great 💥

server/executor/testrunner/testrunner_repository.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (r *Repository) Update(ctx context.Context, updated TestRunner) (TestRunner, error) {
+	// enforce ID and default
+	updated.ID = "current"
+
+	tx, err := r.db.BeginTx(ctx, nil)
+	defer tx.Rollback()
+	if err != nil {
+		return TestRunner{}, err
+	}
```
danielbdias, line 50: We have no guarantees that `tx` is not `nil` in an error condition (which can result in a `nil` pointer panic). To avoid that we can swap the error handling, what do you think?

```suggestion
	if err != nil {
		return TestRunner{}, err
	}
	defer tx.Rollback()
```

danielbdias (approved):

server/executor/testrunner/testrunner_entities.go
```diff
@@ -0,0 +1,87 @@
...
+	RequiredGates: []RequiredGate{
+		RequiredGateTestSpecs,
+		RequiredGateAnalyzerRules,
+		RequiredGateAnalyzerScore,
+	},
+}
+
+type (
+	RequiredGate  string
+	RequiredGates []RequiredGate
```
schoren, line 38: is this type needed? I don't see any method attached to it

server/executor/testrunner/testrunner_entities.go
```diff
@@ -0,0 +1,87 @@
...
+const (
+	ResourceName       = "TestRunner"
+	ResourceNamePlural = "TestRunners"
+)
+
+var Operations = []resourcemanager.Operation{
+	resourcemanager.OperationGet,
+	resourcemanager.OperationList,
+	resourcemanager.OperationUpdate,
+}
```
schoren, line 24: I think this is not being used. we should try to remove the references to resource manager from packages

server/executor/testrunner/testrunner_repository.go
```diff
@@ -0,0 +1,149 @@
...
+		SELECT
+			"name",
+			"required_gates"
+		FROM test_runners`
+)
+
+func (r *Repository) GetDefault(ctx context.Context) TestRunner {
+	tr, _ := r.Get(ctx, id.ID("current"))
+	return tr
+}
```
schoren, line 94:
```suggestion
func (r *Repository) GetDefault(ctx context.Context) TestRunner {
	tr, err := r.Get(ctx, id.ID("current"))
	if err != nil {
		return &DefaultTestRunner
	}
	
	return tr
}
```

server/executor/runner.go
```diff
@@ -152,6 +160,14 @@ func (r persistentRunner) Run(ctx context.Context, testObj test.Test, metadata t
 
 	r.listenForStopRequests(ctx, cancelCtx, run)
 
+	// configuring required gates
+	if requiredGates != nil {
+		run = run.ConfigureRequiredGates(*requiredGates)
+	} else {
+		testRunner := r.trGetter.GetDefault(ctx)
+		run = run.ConfigureRequiredGates(testRunner.RequiredGates)
+	}
```
schoren, line 169:
```suggestion
	if requiredGates == nil {
	  rg := r.trGetter.GetDefault(ctx) 
    requiredGates = &rg
	}
	run = run.ConfigureRequiredGates(testRunner.RequiredGates)
```

schoren (changes requested):
My main request change is to remove the reference to resourcemanager for the unused list of operations. The rest are readability suggestions

server/migrations/27_add_test_runner.down.sql
xoscar: All of the tables have plural names so I think I'm going to stick with that

xoscar:
@schoren thanks for your review! I think I addressed all of your comments, I'm going to merge this to the feature branch so we can move forward

Issue #2834, opened 2023-06-28
Add CLI command to test DataStore connection


danielbdias:
Using the same idea that we have on our Frontend, we could have a command to validate if a datastore resource file is valid and we have a valid connection to help the users to check the configuration is valid.

We could have a `validate` verb to check that, doing something like this:
```sh
> tracetest validate datastore -f my-data-store.yaml

✔ Port checking:  You are using a commonly used port
✔ Connectivity:   Tracetest connected to "jaeger-query.tracetest:16685"
✔ Authentication: Tracetest managed to authenticate with Jaeger
✔ Fetch traces:   Traces were obtained successfully
```

PR #2891, opened 2023-07-07
feat(cli): adding datastore validate command on CLI

danielbdias:
This PR adds a `validate datastore` command on the CLI to allow users to check if Tracetest can connect to a Datastore.

Loom: https://www.loom.com/share/fb2654c9bc6647f2a41f0622c08f12bd?sid=9d761c5e-2590-4bda-9908-107572e43eee

## Changes

- Added `validate datastore` command

## Fixes

- closes https://github.com/kubeshop/tracetest/issues/2834

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


xoscar:
Great job @danielbdias everything looks good, I think we just have to define what a `validate` cmd would do for other "resources" or if this is only for data stores, also, the name is something I think we could think about, as I believe we are not "validating" but instead testing the connection between the server and the data store right?

The command in question is:

```bash
tracetest validate datastore -f <file_path>
```

I think this is something @kdhamric and @olha23 should be involved in.

kdhamric:
We could use tracetest run... wait ;>. The UI uses the words 'Test Connection', and the successful message on testing it is 'All tests successful - configuration is valid'.

Structure:
tracetest {VERB} datastore -f ./mydatastoreconfig.yaml

Test is 'out' as we already use it. Other possibilities:
validate (the word Daniel selected)
check

If we were going to start allowing a verification of the yaml for resources (ie tracetest validate test ./mytest.yaml) I could see leaving the validate keyword open.

danielbdias:
Closing it for now to discuss with the team how we can add it to the CLI with a clearer setup for the command. After defining this feature better, we can reopen this PR if needed.

We will continue to track it on https://github.com/kubeshop/tracetest/issues/2834.

PR #2905, opened 2023-07-10
feat: events app insights

mathnogueira:
This PR adds support for events in Microsoft app insights data store

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/tracedb/azureappinsights.go
+++ b/server/tracedb/azureappinsights.go
@@ -115,70 +115,181 @@ func (db *azureAppInsightsDB) GetTraceByID(ctx context.Context, traceID string) 
 	return parseAzureAppInsightsTrace(traceID, table)
 }
 
-type columnIndex map[string]int
+type spanTable struct {
+	rows []spanRow
+}
+
+func (st *spanTable) Spans() []spanRow {
+	output := make([]spanRow, 0)
+	for _, row := range st.rows {
+		if row.Type() != "trace" {
+			output = append(output, row)
+		}
+	}
+
+	return output
+}
+
+func (st *spanTable) Events() []spanRow {
+	output := make([]spanRow, 0)
+	for _, row := range st.rows {
+		if row.Type() == "trace" {
+			output = append(output, row)
+		}
+	}
+
+	return output
+}
+
+type spanRow struct {
+	values map[string]any
+}
+
+func (sr *spanRow) Get(name string) any {
+	return sr.values[name]
+}
+
+func (sr *spanRow) Type() string {
+	return sr.values["itemType"].(string)
+}
+
+func (sr *spanRow) ParentID() string {
+	return sr.values["operation_ParentId"].(string)
+}
+
+func (sr *spanRow) SpanID() string {
+	return sr.values["id"].(string)
+}
+
+func newSpanTable(table *azquery.Table) spanTable {
+	spanRows := make([]spanRow, 0, len(table.Rows))
+	for _, row := range table.Rows {
+		spanRows = append(spanRows, newSpanRow(row, table.Columns))
+	}
+
+	return spanTable{spanRows}
+}
+
+func newSpanRow(row azquery.Row, columns []*azquery.Column) spanRow {
+	values := make(map[string]any)
+	for i, column := range columns {
+		name := *column.Name
+		if value := row[i]; value != nil {
+			values[name] = value
+		}
+	}
+
+	return spanRow{values}
+}
 
 func parseAzureAppInsightsTrace(traceID string, table *azquery.Table) (model.Trace, error) {
-	columnIndex := mapColumnNames(table.Columns)
-	spans := make([]model.Span, len(table.Rows))
-
-	for i, row := range table.Rows {
-		span, err := parseRowToSpan(row, columnIndex)
-		if err != nil {
-			return model.Trace{}, err
-		}
-
-		spans[i] = span
+	spans, err := parseSpans(table)
+	if err != nil {
+		return model.Trace{}, err
 	}
 
 	return model.NewTrace(traceID, spans), nil
 }
 
-var columnNamesMap = map[string]string{
-	"traceId":    "operation_Id",
-	"spanId":     "id",
-	"parentId":   "operation_ParentId",
-	"name":       "name",
-	"attributes": "customDimensions",
-	"startTime":  "timestamp",
-	"duration":   "duration",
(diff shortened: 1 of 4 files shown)


server/tracedb/azureappinsights.go
```diff
@@ -115,70 +115,181 @@ func (db *azureAppInsightsDB) GetTraceByID(ctx context.Context, traceID string)
...
+		spanMap[span.ID.String()] = &span
 	}
 
-	return model.NewTrace(traceID, spans), nil
+	for _, eventRow := range eventRows {
+		parentSpan := spanMap[eventRow.ParentID()]
+		event, err := parseEvent(eventRow)
+		if err != nil {
+			// don't fail conversion, just print a message
+			fmt.Printf(`failed to convert span event "%s" from span "%s" from AppInsights\n`, event.Name, eventRow.ParentID())
```
schoren, line 219: why do we want to ignore these errors?

server/tracedb/azureappinsights.go
```diff
@@ -191,13 +302,8 @@ func parseRowToSpan(row azquery.Row, columnIndex columnIndex) (model.Span, error
...
 }
 
-func parseSpanId(span *model.Span, row azquery.Row, index int) error {
-	if index == -1 {
-		return fmt.Errorf("spanId column not found")
-	}
-
-	rawSpanId := row[index].(string)
-	spanId, err := trace.SpanIDFromHex(rawSpanId)
+func parseSpanId(span *model.Span, value any) error {
```
schoren, line 305:
```suggestion
func parseSpanID(span *model.Span, value any) error {
```

server/tracedb/azureappinsights.go
```diff
@@ -224,12 +326,8 @@ func parseAttributes(span *model.Span, row azquery.Row, index int) error {
 	return nil
 }
 
-func parseParentId(span *model.Span, row azquery.Row, index int) error {
-	if index == -1 {
-		return fmt.Errorf("parentId column not found")
-	}
-
-	rawParentId, ok := row[index].(string)
+func parseParentId(span *model.Span, value any) error {
```
schoren, line 329:
```suggestion
func parseParentID(span *model.Span, value any) error {
```

server/tracedb/azureappinsights.go
```diff
@@ -224,12 +326,8 @@ func parseAttributes(span *model.Span, row azquery.Row, index int) error {
...
+	rawParentId, ok := value.(string)
```
schoren, line 330:
```suggestion
	rawParentID, ok := value.(string)
```

server/tracedb/azureappinsights.go
```diff
@@ -115,70 +115,181 @@ func (db *azureAppInsightsDB) GetTraceByID(ctx context.Context, traceID string)
...
+func newSpanRow(row azquery.Row, columns []*azquery.Column) spanRow {
+	values := make(map[string]any)
+	for i, column := range columns {
+		name := *column.Name
+		if value := row[i]; value != nil {
+			values[name] = value
+		}
+	}
+
+	return spanRow{values}
```
mathnogueira, line 182: The reason I created those structs here is that AppInsights returns a table with the following fields:

```go
struct {
  Columns []string
  Rows    []struct {
    Values []string
  }
}
```

To get a value from a row, you have to know which index of the array it is. For that, you need to lookup the `Columns` array: 

```go
spanID := row[32]`
// or
`spanID := row[columns["name"]]`
```

So, by converting these fields into a `map[string]any` was an alternative that would make it easier to use (and debug).

schoren (approved):
left a few comments. I don't fully understand the appinsight model, so another pair of eyes would be good

server/tracedb/azureappinsights.go
```diff
@@ -115,70 +115,181 @@ func (db *azureAppInsightsDB) GetTraceByID(ctx context.Context, traceID string)
...
+
+	return model.NewTrace(traceID, spans), nil
+}
 
-	for i, row := range table.Rows {
-		span, err := parseRowToSpan(row, columnIndex)
+func parseSpans(table *azquery.Table) ([]model.Span, error) {
+	spanTable := newSpanTable(table)
+	spanRows := spanTable.Spans()
+	eventRows := spanTable.Events()
```
mathnogueira, line 197: AppInsights returns everything as `rows`. So we use the field `itemType` to know if that is a span (`itemType != "trace"`) or an event (`itemType == "trace"`).

server/tracedb/azureappinsights.go
```diff
@@ -115,70 +115,181 @@ func (db *azureAppInsightsDB) GetTraceByID(ctx context.Context, traceID string)
...
 	}
 
-	return model.NewTrace(traceID, spans), nil
+	for _, eventRow := range eventRows {
+		parentSpan := spanMap[eventRow.ParentID()]
+		event, err := parseEvent(eventRow)
+		if err != nil {
+			// don't fail conversion, just print a message
+			fmt.Printf(`failed to convert span event "%s" from span "%s" from AppInsights\n`, event.Name, eventRow.ParentID())
+		}
```
mathnogueira, line 220: I didn't want to fail the conversion here. The errors that return here are caused by parsing errors (timestamp and jJSON unmarshal). The reasoning is that I rather have fewer things in a span than not have any spans at all.

server/tracedb/azureappinsights.go
mathnogueira: I didn't want to fail the conversion here. The errors that return here are caused by parsing errors (timestamp and jJSON unmarshal). The reasoning is that I rather have fewer things in a span than not have any spans at all.

server/tracedb/azureappinsights.go
mathnogueira: No. I exclude all `nil` values from the map before calling those functions

server/tracedb/azureappinsights.go
schoren: so if there's a parse error, the entire event doesn't show? 
maybe we could add an event with some kind of "parsing error" message?

xoscar (approved):
@mathnogueira this is looking great, I like how you removed the need for the maps I added in the v1

server/tracedb/azureappinsights.go
mathnogueira: I was thinking about that, and ignore it. Let's fail the datastore if we get an error

PR #2939, opened 2023-07-13
feat(frontend): add transaction automate page

jorgeepc:
This PR adds the `transaction` automate page to the transactionRun detail section.

## Changes

- new transaction automate page
- refactor components to share UI+behavior for tests and transactions

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom

https://www.loom.com/share/4c6dc58fd7bb48daa2ec0d07aa941e26?sid=5a707c44-a90f-4763-a3fb-0fec4799fd89

## Screenshot

Screenshot 2023-07-13 at 17 35 06


xoscar (approved):
Hey @jorgeepc the code looks good, but I'm not sure about having a `trigger` concept at the transaction level, as technically we are not sending a request like we do for tests 🤔

jorgeepc:
> Hey @jorgeepc the code looks good, but I'm not sure about having a `trigger` concept at the transaction level, as technically we are not sending a request like we do for tests 🤔

@xoscar thanks for the review. We can change that name, maybe something like `"Detail"`? I was thinking more of the idea that you can still trigger the transaction but I think you are right, we don't have a `"Trigger"` phase for transactions.

PR #2943, opened 2023-07-14
fix(frontend): fix trace panel size

### LINKED ISSUE #2936: Fix size of the screens
Left and right parts must be the same by default, it’s mean there split the screen in half. Now their width is different, which is very noticeable when switching tabs, see

Image

:

jorgeepc:
This PR fixes the size of the panels for the Trace tab in order to have the same sizes for all the tabs (Trigger, Trace, Test and Automate)

## Changes

- fix panel size

## Fixes

- fixes #2936 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/components/RunDetailTrace/AnalyzerPanel.tsx
+++ b/web/src/components/RunDetailTrace/AnalyzerPanel.tsx
@@ -10,7 +10,7 @@ interface IProps {
 
 const panel = {
   name: 'ANALYZER',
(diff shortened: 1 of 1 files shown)


danielbdias (approved):

xoscar (approved):
🪄

PR #2964, opened 2023-07-18
feat: add structure for agent

mathnogueira:
This PR adds the basic code to start the agent (no behavior yet)

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


agent/config/config.go
```diff
@@ -0,0 +1,29 @@
...
+type Config struct {
+	ServerURL string `json:"serverURL"`
+	DevMode   bool   `json:"devMode"`
+	APIKey    string `json:"apiKey"`
+}
+
+func LoadConfig() (Config, error) {
+	viper.SetEnvPrefix("tracetest")
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
```
mathnogueira, line 19: The idea is to have the agent to be loaded via flags `--apiKey` or via env variables `TRACETEST_APIKEY`

mathnogueira: As I'm not sure how we gonna handle configuration with files yet, I'll add support for that later

agent/initialization/start.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"fmt"
+
+	"github.com/kubeshop/tracetest/agent/config"
+)
+
+// Start the agent with given configuration
+func Start(config config.Config) {
+	fmt.Println("Starting agent")
+}
```
mathnogueira, line 12: In the server, we would call this function to start the agent along with the server

agent/cmd/root.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	cliExitInterceptor func(code int)
+)
+
+var rootCmd = cobra.Command{
+	Use: "agent",
+}
```
mathnogueira, line 16: In the `tracetest cli` we are going to import this and add as a child of the `tracetest` command. This way you can call it as `tracetest agent start`

agent/cmd/root.go
```diff
@@ -0,0 +1,42 @@
...
+
+var (
+	cliExitInterceptor func(code int)
+)
+
+var rootCmd = cobra.Command{
+	Use: "agent",
+	Short: "Manages the tracetest agent"
+	Long: "Manages the tracetest agent"
+}
```
mathnogueira, line 18: In the tracetest CLI we are going to import this and add it as a child of the `tracetest` command. This way you can call it via `tracetest agent start`

agent/initialization/start.go
mathnogueira: It will be. When I say `start the agent` is literally starting the process. As we have to have the agent running along with the tracetest server, we have to include everything in the same process as you cannot run two different things in a docker container

xoscar: Oh makes sense!

xoscar (approved):

PR #2970, opened 2023-07-20
feat: create client for receiving trigger requests from the server

mathnogueira:
This PR adds the agent client responsible for receiving trigger requests from a server

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


agent/client/client.go
```diff
@@ -0,0 +1,96 @@
...
+	"google.golang.org/grpc"
+)
+
+type Config struct {
+	APIKey    string
+	AgentName string
+}
+
+type SessionConfig struct {
+	BatchTimeout time.Duration
```
mathnogueira, line 19: When connecting to the server, we gonna receive information about how the agent must behave. One of the behaviors that will be configured by the server is how long the agent must batch spans before sending the entire batch when it is acting as a collector.

agent/client/client.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type Client struct {
+	conn          *grpc.ClientConn
+	config        Config
+	sessionConfig *SessionConfig
+	done          chan bool
+
+	triggerListener func(*proto.TriggerRequest) error
```
mathnogueira, line 28: Decided to use a callback instead of a channel because this way we don't freeze the agent when there's more than one request at the same time.

xoscar (approved):
GJ @mathnogueira!

PR #2971, opened 2023-07-20
feat(server): refactor runner channels into abstract queues

schoren:
This PR introduces the concept of test runner pipelines, which decouples the steps of the pipeline from each other. The pipeline consists of several steps, including trigger, poll trace, analyze result, and assert. Instead of each step needing to know what step comes next, it only needs to have a reference to an "outputQueue". This outputQueue is the "input" queue for the next step, which allows for greater flexibility and modularity in the codebase.

The main goal of this PR is to remove all the manual queues that were implemented using Go channels. Previously, each step had a copy/pasted worker code that relied on channels, which made it hard to replace the queue mechanism. The new pipeline queues rely on an abstract queue driver, which allows for different types of queues to be created independently. Currently, we only implement an in-memory driver using Go channels, so we're not changing behavior, just architecture.

With everything decoupled, we can independently create different types of queues, such as a PostgreSQL listen/notify queue. This also adds flexibility to modify the pipeline, such as adding steps or changing the order, without needing to modify any existing steps.

Explanation of the Queue:
https://www.loom.com/share/8209a55d500c4766a0f137f0c723bf22

Explanation of the Pipeline:
https://www.loom.com/share/94aca45c8e8f4f17ab06403e1f65833a

How those fit in the application:
https://www.loom.com/share/43f74fccecb042f69b89f5f9098e56d6

## Changes

- Introduced the concept of test runner pipelines
- Decoupled the steps of the pipeline using output queues
- Removed manual queues implemented using Go channels
- Added an abstract queue driver for greater flexibility
- Added an in-memory driver using Go channels
- Updated the pipeline steps to use the new queue mechanism

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/test/run_repository.go
```diff
@@ -602,7 +603,7 @@ func (r *runRepository) GetTransactionRunSteps(ctx context.Context, id id.ID, ru
 WHERE transaction_run_steps.transaction_run_id = $1 AND transaction_run_steps.transaction_run_transaction_id = $2
 ORDER BY test_runs.completed_at ASC
 `
-	query, params := sqlutil.Tenant(ctx, query, runID, id)
+	query, params := sqlutil.Tenant(ctx, query, strconv.Itoa(runID), id)
```
schoren, line 606: this is required for the new pgx db  driver

server/testdb/test_run_event.go
```diff
@@ -136,10 +135,6 @@ func readTestRunEventFromRows(rows *sql.Rows) (model.TestRunEvent, error) {
 	)
 
 	if err != nil {
-		if errors.Is(err, sql.ErrNoRows) {
```
schoren, line 139: the controller compares with `errors.Is(err, sql.ErrNoRows)` now. this func checks the wrapped errors, so it will correctly assert that the error is `sql.ErrNoRows` without the need to return a custom error

server/app/test_pipeline.go
```diff
@@ -0,0 +1,106 @@
...
+
+	return executor.NewTestPipeline(
+		pipeline,
+		subscriptionManager,
+		pipeline.GetQueueForStep(3), // assertion runner step
+		runRepo,
+		trRepo,
+		ppRepo,
+		dsRepo,
+	)
```
danielbdias, line 105: Can we move this magic index `3` (for the assertion runner step) to a variable (or even a constant)?

```suggestion
	assertionRunnerStepIndex := 3

	return executor.NewTestPipeline(
		pipeline,
		subscriptionManager,
		pipeline.GetQueueForStep(assertionRunnerStepIndex), // assertion runner step
		runRepo,
		trRepo,
		ppRepo,
		dsRepo,
	)
```

server/app/test_pipeline.go
```diff
@@ -0,0 +1,106 @@
...
+		WithSubscriptor(subscriptionManager).
+		WithDataStoreGetter(dsRepo).
+		WithPollingProfileGetter(ppRepo).
+		WithTestGetter(testRepo).
+		WithRunGetter(runRepo)
+
+	pipeline := executor.NewPipeline(queueBuilder,
+		executor.PipelineStep{Processor: runner, Driver: executor.NewInMemoryQueueDriver("runner")},
+		executor.PipelineStep{Processor: tracePoller, Driver: executor.NewInMemoryQueueDriver("tracePoller")},
+		executor.PipelineStep{Processor: linterRunner, Driver: executor.NewInMemoryQueueDriver("linterRunner")},
```
danielbdias, line 91: One thing to consider in the future: since we are considering renaming the internals of `environment` and `transaction` soon, we can also add the `linter` to the packet and change it to `analyzer` explicitly. Does it make sense?

server/config/server.go
```diff
@@ -80,23 +82,18 @@ func (c *AppConfig) PostgresConnString() string {
 	defer c.mu.Unlock()
 
 	if postgresConnString := c.vp.GetString("postgresConnString"); postgresConnString != "" {
-		return postgresConnString
+		fmt.Println("ERROR: postgresConnString was discontinued. Migrate to the new postgres format")
```
danielbdias, line 85: Something to think about in the future: we should move these server logs from `fmt` to something else, like `zap` or `logrus`, that we can centralize and configure later.
Does it make sense to create a ticket for that? (I can do that if needed!)

server/executor/poller_executor_test.go
```diff
@@ -457,57 +457,55 @@ func executeAndValidatePollingRequests(t *testing.T, pollerExecutor executor.Pol
...
+		Strategy: pollingprofile.Periodic,
+		Periodic: &pollingprofile.PeriodicPollingConfig{
+			RetryDelay: retryDelay.String(),
+			Timeout:    maxWaitTimeForTrace.String(),
+		},
+	}
 
-		finished, finishReason, anotherRun, err := pollerExecutor.ExecuteRequest(request)
-		run = anotherRun // should store a run to use in another iteration
+	job := executor.NewJob()
```
danielbdias, line 471: On every call to `NewJob`, we set `Test`, `Run`, and `PollingProfile`. Does it make sense to add them as parameters on `NewJob`?

server/executor/queue.go
```diff
@@ -0,0 +1,474 @@
...
+	if q.transactionRuns == nil {
+		return transaction.TransactionRun{}
+	}
+
+	tranRun, err := q.transactionRuns.GetTransactionRun(ctx, job.Transaction.ID, job.TransactionRun.ID)
+	if errors.Is(err, sql.ErrNoRows) {
+		return transaction.TransactionRun{}
+	}
+	if err != nil {
+		panic(err)
```
danielbdias, line 361: Philosophical question: if an error happens with a Job, should we just `panic`, bringing down the engine or just drop the `job` and add it to a `dead letter` queue/storage?

server/executor/queue.go
```diff
@@ -0,0 +1,474 @@
...
+func (r *InMemoryQueueDriver) SetQueue(q *Queue) {
+	r.q = q
+}
+
+func (r InMemoryQueueDriver) Enqueue(job Job) {
+	r.queue <- job
+}
+
+func (r InMemoryQueueDriver) Start() {
+	for i := 0; i < 5; i++ {
```
danielbdias, line 455: The number `5` here is the number of internal workers for this driver, isn't it? Can we move it to a constant? (and maybe later to a tracetest config?)

server/pkg/id/generator.go
```diff
@@ -56,12 +52,12 @@ func (g randGenerator) ID() ID {
 
 func (g randGenerator) TraceID() trace.TraceID {
 	tid := trace.TraceID{}
-	g.rand.Read(tid[:])
+	rand.New(rand.NewSource(time.Now().UnixNano())).Read(tid[:])
```
danielbdias, line 55: A small suggestion (feel free to ignore!): we can isolate the seed from the generator for readability:
```suggestion
	rndSeed := rand.NewSource(time.Now().UnixNano())
	rand.New(rndSeed).Read(tid[:])
```

server/pkg/id/generator.go
```diff
@@ -56,12 +52,12 @@ func (g randGenerator) ID() ID {
...
 	return tid
 }
 
 func (g randGenerator) SpanID() trace.SpanID {
 	sid := trace.SpanID{}
-	g.rand.Read(sid[:])
+	rand.New(rand.NewSource(time.Now().UnixNano())).Read(sid[:])
```
danielbdias, line 61: Same here:
```suggestion
	rndSeed := rand.NewSource(time.Now().UnixNano())
	rand.New(rndSeed).Read(sid[:])
```

server/test/run_repository.go
danielbdias: Question: what is the idea of using `pgx` instead of `pq` directly?

testing/server-tracetesting/features/http_test/08_rerun_http_test.yml
```diff
@@ -22,9 +22,6 @@ spec:
   - selector: span[name = "query INSERT"]
     assertions:
     - attr:tracetest.selected_spans.count = 1
-  - selector: span[name = "exec UPDATE"]
-    assertions:
-    - attr:tracetest.selected_spans.count = 1
```
danielbdias, line 27: What is the rationale for removing this assertion? Don't we need to update the test anymore?

server/test/run_repository.go
schoren: pgx has support for `LISTEN` / `NOTIFY`. Originally this PR was going to do everything but it got larger than expected so I'll implement that in a separated PR, but we already did the driver change at the beginning. The idea was to check that everything worked with the new driver

server/app/test_pipeline.go
schoren: I agree, but I saw that there's a distinction between `linter` and `analyzer`. I didn't dig too much into it, but that looks like a good time to normalize this kinds of things

server/config/server.go
schoren: +1, do you mind creating the ticket @danielbdias ?

server/executor/pipeline.go
```diff
@@ -0,0 +1,89 @@
...
+func (p *Pipeline) GetQueueForStep(i int) *Queue {
+	if i < 0 || i >= len(p.queues) {
+		return nil
+	}
+
+	return p.queues[i]
+}
+
+func (p *Pipeline) Begin(ctx context.Context, job Job) {
+	p.queues[0].Enqueue(ctx, job)
```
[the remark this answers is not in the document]
schoren: makes sense. I'll add it

server/executor/poller_executor_test.go
schoren: This is not the case for transactions, where it only passes `Transaction` and `TransactionRun`.

server/executor/queue.go
schoren: the only reason for this to happen is because we don't have access to the DB for some reason. if we lost access to DB, or it's locked, or queries are wrong, it feels reasonable to have a panic and restart the service.  I'm not married to this idea though.

While I think it might be a good idea to have a death queue, it might add complexity that we don't neccesariliy need at the moment IMO

server/executor/queue.go
schoren: I'll add it as a constant, but I don't think this will have a lot of use outside testing

testing/server-tracetesting/features/http_test/08_rerun_http_test.yml
schoren: Good question. This was an RunUpdate made after creating the new run when rerunning a test, to update the state. I removed it incorrectly. Readding it

server/config/server.go
danielbdias: done! Issue: https://github.com/kubeshop/tracetest/issues/2996

PR #2975, opened 2023-07-21
feat: send trigger requests from agent

mathnogueira:
This PR allows the agent to send trigger requests to a service and send the trigger response back to the control plane server.

### Implemented flow
```mermaid
sequenceDiagram
	participant SUT as Service Under Test
	participant Agent
	participant Server
	
	Server ->> Agent: Send trigger request
	Server ->> Server: Waits for trigger response from agent
	Agent ->> SUT: Triggers operation
	SUT ->> Agent: Send response from operation
	Agent ->> Server: Forwards response from Service Under Test
```

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


agent/client/client.go
```diff
@@ -25,8 +25,8 @@ type Client struct {
 	sessionConfig *SessionConfig
 	done          chan bool
 
-	triggerListener func(*proto.TriggerRequest) error
-	pollListener    func(*proto.PollingRequest) error
+	triggerListener func(context.Context, *proto.TriggerRequest) error
```
mathnogueira, line 28: I'm adding these contexts so we can add instrumentation to the agent as well. It will be important to be able to collect metrics and some traces from agents if users allow it

agent/initialization/start.go
```diff
@@ -1,12 +1,37 @@
...
 )
 
 // Start the agent with given configuration
 func Start(config config.Config) {
 	fmt.Println("Starting agent")
+	ctx := context.Background()
+
+	client, err := client.Connect(ctx, config.ServerURL,
+		client.WithAPIKey(config.APIKey),
+		client.WithAgentName(config.AgentName),
```
schoren, line 21: is this agent name configurable by user? does it act as an ID or is it just for displaying purposes?

agent/workers/trigger.go
```diff
@@ -0,0 +1,219 @@
...
+		return nil
+	}
+
+	headers := make([]trigger.HTTPHeader, 0, len(pt.Headers))
+
+	for _, header := range pt.Headers {
+		headers = append(headers, trigger.HTTPHeader{Key: header.Key, Value: header.Value})
+	}
+
+	return &trigger.HTTPRequest{
```
schoren, line 82: would it make sense to use `DeepCopy` instead of manually mapping fields?

schoren (approved):
looks good. left a few questions

agent/initialization/start.go
mathnogueira: only for displaying purposes. By default, it's the host name, but the user might override it with a flag: `tracetest agent start --name my-agent`

agent/workers/trigger.go
mathnogueira: I thought about it. However, it would require adding the `dc` tag to some fields in the server's `trigger.Trigger` struct. I'm not convinced the server core should be aware of these details about the agent.

PR #2979, opened 2023-07-24
feat: agent polling

mathnogueira:
This PR adds the polling capability on the agent

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


xoscar (approved):

agent/initialization/start.go
```diff
@@ -25,11 +25,16 @@ func Start(config config.Config) {
...
 	triggerWorker := workers.NewTriggerWorker(client)
+	pollingWorker := workers.NewPollerWorker(client)
 
 	client.OnTriggerRequest(func(ctx context.Context, tr *proto.TriggerRequest) error {
 		return triggerWorker.Trigger(ctx, tr)
 	})
 
+	client.OnPollingRequest(func(ctx context.Context, pr *proto.PollingRequest) error {
+		return pollingWorker.Poll(ctx, pr)
+	})
```
danielbdias, line 36: Stupid question, since we are just forwarding parameters to `Poll` why not pass it directly to `client.OnPolingRequest`?

```suggestion
	client.OnPollingRequest(pollingWorker.Poll)
```

danielbdias (approved):

agent/initialization/start.go
mathnogueira: I didn't think about it. Makes total sense, I'll do it

PR #2980, opened 2023-07-24
feature(backend) Tenant ID support

xoscar:
This PR adds support for tenant ID to all resources in the application

## Changes

- Adds tenant id column to all resources
- Includes middleware adding it to the context
- Adds tests and validations

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


server/pkg/sqlutil/tenant.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/kubeshop/tracetest/server/http/middleware"
+)
+
+func Tenant(ctx context.Context, query string, params ...any) (string, []any) {
```
xoscar, line 11: From here we can grab add the tenant query to the operation and grab the tenant id from the context

server/http/middleware/tenant.go
```diff
@@ -0,0 +1,44 @@
...
+
+type key string
+
+var (
+	TenantIDKey key = "tenantID"
+)
+
+const HeaderTenantID = "X-Tracetest-TenantID"
+
+func TenantMiddleware(next http.Handler) http.Handler {
```
xoscar, line 19: Validates that the tenant id has the proper format, if all is correct, we add it to the context so the rest of the app can use it

server/migrations/28_add_tenant_id.up.sql
```diff
@@ -0,0 +1,69 @@
+BEGIN;
```
mathnogueira, line 1: You should also create an index for those columns:

```sql
CREATE INDEX idx_config_tenant_id
ON config(tenant_id);
```

Otherwise this will be very costly to query after a few weeks of saas being released.

server/pkg/sqlutil/tenant.go
```diff
@@ -0,0 +1,58 @@
...
+	paramNumber := len(params) + 1
+	condition := fmt.Sprintf(" %s %stenant_id = $%d)", queryPrefix, prefix, paramNumber)
+
+	return query + condition, append(params, tenantID)
+}
+
+func TenantID(ctx context.Context) *string {
+	tenantID := ctx.Value(middleware.TenantIDKey)
+
+	if tenantID == "" || tenantID == nil {
```
mathnogueira, line 41: You are creating each resource with:

```sql
COLUMN tenant_id uuid DEFAULT '00000000-0000-0000-0000-000000000000';
``` 

Shouldn't you inject `00000000-0000-0000-0000-00000000000` into the context when user passes no tenant information?

xoscar: I removed the default value because that might cause problems with backward compatibility, right now I think is better to have it always be `NULL`. Good catch!

mathnogueira (approved):
LGTM

PR #2997, opened 2023-07-27
feat: agent internal collector

mathnogueira:
This PR allows the agent to start a HTTP and gRPC server to receive traces from a local application. This will be used by the local `dev mode` we will introduce.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


agent/collector/http_server.go
```diff
@@ -0,0 +1,204 @@
...
+	return s.parseProtoBuf(protoBody)
+}
+
+func (s httpServer) parseBody(reqBody io.ReadCloser, contentType string) (*pb.ExportTraceServiceRequest, error) {
+	var body []byte
+	if b, err := io.ReadAll(reqBody); err == nil {
+		body = b
+	} else {
+		return nil, err
+	}
```
danielbdias, line 114: Is there a reason to use this `if/else`, or can we simplify it using an early return, like this?
```suggestion
	var body []byte
	body, err := io.ReadAll(reqBody)
	
	if err != nil {
		return nil, err
	}
```

agent/collector/http_server.go
```diff
@@ -0,0 +1,204 @@
...
+		w:           w,
+		contentType: contentType,
+	}
+}
+
+func (r httpResponse) send(statusCode int, message proto.Message) error {
+	body, err := r.paseResponseBody(message)
+	if err != nil {
+		fmt.Println("Could not attach body to response", err.Error())
+		return err
```
danielbdias, line 143: Does it make sense to wrap this error with this message too?
```suggestion
		fmt.Println("Could not attach body to response", err.Error())
		return fmt.Errorf("could not attach body to response: %w", err)
```

danielbdias (approved):

server/otlp/ingester.go
```diff
@@ -16,21 +16,32 @@ import (
 	v1 "go.opentelemetry.io/proto/otlp/trace/v1"
 )
 
+type Ingester interface {
```
schoren, line 19: is this change required? looks like no other package is referencing the interface. maybe I'm missing something though

agent/collector/ingester.go
```diff
@@ -0,0 +1,130 @@
...
+	"time"
+
+	"github.com/kubeshop/tracetest/server/otlp"
+	pb "go.opentelemetry.io/proto/otlp/collector/trace/v1"
+	v1 "go.opentelemetry.io/proto/otlp/trace/v1"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type ingester interface {
```
schoren, line 17: this interface also seems unused, maybe it can be removed?

schoren (approved):
looks awesome! If i'm not mistaken, the server ingestor interface can be left unexported and the agent's ingestor interface is not used, but I might be missing something

agent/collector/ingester.go
mathnogueira: changed it to `stoppable` because we might have to stop the ingester.

server/otlp/ingester.go
mathnogueira: after I deleted the `collector.ingester` interface, we needed this to be exported.

PR #2998, opened 2023-07-27
feat(frontend): prepare api base url for submodule

jorgeepc:
This PR introduces changes to our `api`/`redux` functionality in the frontend app in order to support it as a git submodule.

## Changes

- refactor tracetest API redux toolkit to support dynamic `baseUrl`
- implement the [injectEndpoints](https://redux-toolkit.js.org/rtk-query/usage/code-splitting) method for RTK to inject endpoints into the original Tracetest API 
- update references to the new api module

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


web/src/redux/apis/Tracetest/Tracetest.api.ts
```diff
@@ -0,0 +1,27 @@
...
+import Env from 'utils/Env';
+
+const customBaseQuery: BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError> = async (
+  args,
+  api,
+  extraOptions
+) => {
+  const baseUrl = Env.get('baseApiUrl');
+
+  return fetchBaseQuery({
```
xoscar, line 12: Thinking ahead a little bit, could we use something like this to inject the tokens or other things coming from the top source?

web/src/redux/apis/Tracetest/endpoints/DataStore.endpoint.ts
```diff
@@ -0,0 +1,49 @@
...
+        url: `/config/connection`,
+        method: HTTP_METHOD.POST,
+        body: connectionTest,
+      }),
+      transformResponse: (result: TRawConnectionResult) => ConnectionResult(result),
+      transformErrorResponse: ({data: result}) => ConnectionResult(result as TRawConnectionResult),
+    }),
+  }),
+});
+
```
xoscar, line 47: I like this approach, if we want it to be even more flexible we could have it be a function which receives an API instance to inject the endpoints to, but this is not needed atm

xoscar (approved):
Great job @jorgeepc!

web/src/redux/apis/Tracetest/Tracetest.api.ts
jorgeepc: Yes, that is the idea. Use this customBaseQuery to inject tokens, tenantId, etc.

PR #3009, opened 2023-07-31
fix(frontend): fix build warnings related to typescript version

jorgeepc:
This PR fixes some build warnings in our web app related to a typescript conflict with a nested package.

**warnings:**
```
TypeError: Cannot set property mark of #<Object> which has only a getter
TypeError: Cannot set property mark of #<Object> which has only a getter
    at Object.connectTypeScriptPerformance
```

See https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/797

## Changes

- remove storybook packages causing issues

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


xoscar (approved):

danielbdias (approved):

web/package.json
```diff
@@ -81,8 +81,6 @@
     "cy:run": "cypress run",
     "cy:ci": "cypress run --parallel --record --key $CYPRESS_RECORD_KEY",
     "prettier": "prettier --write ./src",
-    "storybook": "start-storybook -p 6006 -s public",
-    "build-storybook": "build-storybook -s public",
```
danielbdias, line 85: Sad to see the storybook getting rid of the project (I used it a lot in other projects before to test component behaviors), but right now, that makes sense.
Today it is easier to build a component and test it with our UI pointing to an example API than creating a story and mocking many things.

xoscar: Exactly, storybook is great but is not useful at this point, in the future if makes sense we can add it back again

PR #3018, opened 2023-08-01
chore(cli): Adding support for Variable Sets

xoscar:
This PR updates the CLI and fixes all of the automated tests to work with the new variable set changes

## Changes

- Updates CLI resources from environments to variable sets
- Changes the environment flag for the vars flag when used when running a test
- Fixes CLI e2e tests and trace-based testing

## Fixes

- #3014 

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

https://www.loom.com/share/ccd34203368e4b9ba921ce8530cfb258


cli/cmd/resource_run_cmd.go
```diff
@@ -63,7 +63,7 @@ func init() {
 
 	runCmd.Flags().StringVarP(&runParams.DefinitionFile, "file", "f", "", "path to the definition file")
 	runCmd.Flags().StringVar(&runParams.ID, "id", "", "id of the resource to run")
-	runCmd.Flags().StringVarP(&runParams.EnvID, "environment", "e", "", "environment file or ID to be used")
+	runCmd.Flags().StringVarP(&runParams.VarsID, "vars", "", "", "variable set file or ID to be used")
```
danielbdias, line 66: Does it make sense to be `variableset` here?
```suggestion
	runCmd.Flags().StringVarP(&runParams.VarsID, "variableset", "", "", "variable set file or ID to be used")
```

danielbdias (commented):
The overall PR is ok, but I'm missing one CLI e2e test: we should have a scenario to run a test using vars in the old format and the new one, just to guarantee that we are not breaking existing tests for our customers.
One example of test that does that is here: https://github.com/kubeshop/tracetest/blob/main/testing/cli-e2etest/testscenarios/test/run_test_test.go#L40

cli/cmd/middleware.go
```diff
@@ -84,8 +82,8 @@ func (p *resourceParameters) Validate(cmd *cobra.Command, args []string) []error
 
 	p.ResourceName = args[0]
 
-	_, err := resources.Get(p.ResourceName)
-	if errors.Is(err, resourcemanager.ErrResourceNotFound) {
+	exists := resources.Exists(p.ResourceName)
```
xoscar, line 85: I changed this to use a different function because `Get` now writes the warning to the console if deprecated

cli/cmd/resource_run_cmd.go
xoscar: As we were using a shorthand for environments, I thought we could continue using the same approach

cli/cmd/resource_run_cmd.go
```diff
@@ -63,10 +67,16 @@ func init() {
...
 	runCmd.Flags().BoolVarP(&runParams.SkipResultWait, "skip-result-wait", "W", false, "do not wait for results. exit immediately after test run started")
 	runCmd.Flags().StringVarP(&runParams.JUnitOuptutFile, "junit", "j", "", "file path to save test results in junit format")
 	runCmd.Flags().StringSliceVar(&runParams.RequriedGates, "required-gates", []string{}, "override default required gate. "+validRequiredGatesMsg())
+
+	//deprecated
```
xoscar, line 75: Still supporting this in case users need it

cli/cmd/resources.go
```diff
@@ -35,16 +35,20 @@ var (
...
+		"variableset", "variablesets",
 		resourcemanager.WithTableConfig(resourcemanager.TableConfig{
 			Cells: []resourcemanager.TableCellConfig{
 				{Header: "ID", Path: "spec.id"},
 				{Header: "NAME", Path: "spec.name"},
 				{Header: "DESCRIPTION", Path: "spec.description"},
 			},
 		}),
+		resourcemanager.WithResourceType("VariableSet"),
+		resourcemanager.WithApplyPreProcessor(variableSetPreprocessor.Preprocess),
```
xoscar, line 50: Preprocessor to switch `Environment` to `VariableSet` and DeprecatedAlias to pass the `resourceType` validations

cli/cmd/resources.go
```diff
@@ -120,6 +124,13 @@ var (
 		resourcemanager.WithApplyPreProcessor(transactionPreprocessor.Preprocess),
 	)
 
+	// deprecated resources
+	deprecatedEnvironmentClient = resourcemanager.NewClient(
+		httpClient, cliLogger,
+		"environment", "environments",
+		resourcemanager.WithProxyResource("variableset"),
```
xoscar, line 131: Used to create an entry in the registry that will point to the variable set client

cli/pkg/resourcemanager/registry.go
```diff
@@ -30,13 +31,33 @@ func (r *Registry) Get(resourceName string) (Client, error) {
 		return Client{}, ErrResourceNotFound
 	}
 
+	if c.options.proxyResource != "" {
```
xoscar, line 34: If proxy, check if the proxied resource exists

cli/pkg/resourcemanager/registry.go
```diff
@@ -30,13 +31,33 @@ func (r *Registry) Get(resourceName string) (Client, error) {
...
+
+	return true
+}
+
 func (r *Registry) List() []string {
 	var resources []string
-	for k := range r.resources {
-		resources = append(resources, k)
+	for k, c := range r.resources {
+		if c.options.proxyResource == "" {
```
xoscar, line 58: Deleting the proxy resources from the list

cli/preprocessor/variableset.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (vs variableSet) Preprocess(ctx context.Context, input fileutil.File) (fileutil.File, error) {
+	var resource generic
+	err := yaml.Unmarshal(input.Contents(), &resource)
+	if err != nil {
+		vs.logger.Error("error parsing test", zap.String("content", string(input.Contents())), zap.Error(err))
+		return input, fmt.Errorf("could not unmarshal test yaml: %w", err)
+	}
+
+	if resource.Type == "Environment" {
```
xoscar, line 35: Updates Environment to VariableSet

testing/cli-e2etest/Makefile
```diff
@@ -17,6 +18,14 @@ test: # run tests for this application
 	go clean -testcache; \
 	go test -v -timeout 300s -p 1 ./...;
 
+test/scenario: # run tests for this application
```
xoscar, line 21: This will allows us execute a single testscenario

testing/cli-e2etest/testscenarios/test/run_test_test.go
```diff
@@ -85,20 +85,47 @@ func TestRunTestWithHttpTriggerAndEnvironmentFile(t *testing.T) {
 		require.Equal("https://assets.pokemon.com/assets/cms2/img/pokedex/full/143.png", environmentVars.Spec.Values[1].Value)
 	})
 
-	t.Run("should pass when using environment id", func(t *testing.T) {
+	t.Run("should pass when using the deprecated environment definition file", func(t *testing.T) {
```
xoscar, line 88: Adding a test to validate that the old way still works as expected

xoscar:
@danielbdias I added this test
https://github.com/kubeshop/tracetest/pull/3018/files#diff-5b0c58bde933c931dc8fbefd748fa0859adbb4fc222e7856ed2f35f4461b2898R88

To validate that the old way of using envs works as expected

cli/preprocessor/variableset.go
danielbdias: Does it make sense to add a `//TODO` comment saying that this mapping will be removed in future versions?

testing/cli-e2etest/Makefile
danielbdias: Good idea! 🎉

testing/cli-e2etest/testscenarios/variableset/apply_variableSet_test.go
danielbdias, line 1: nit: Can we name these files with smallcaps? (variableset instead of variableSet)

danielbdias (approved):

PR #3022, opened 2023-08-02
feat: agent shutdown

mathnogueira:
This PR allows the server to send a shutdown signal to the agent and make it shutdown remotely.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


agent/client/workflow_listen_for_poll_requests.go
```diff
@@ -10,20 +10,27 @@ import (
...
 	stream, err := client.RegisterPollerAgent(ctx, request)
 	if err != nil {
+		cancelCtx()
 		return fmt.Errorf("could not open agent stream: %w", err)
 	}
 
+	go func() {
+		<-c.done
+		cancelCtx()
+	}()
```
mathnogueira, line 32: Initially, I was going to use the same goroutine for that, however, `stream.RecvMsg(&resp)` is a blocking operation. Meaning that if we got a shutdown command before receiving a polling request, it would just freeze because the server would probably remove this agent from the agent poll.

By using another goroutine, this doesn't happen. We can cancel the context as soon as we receive the shutdown message.

agent/client/workflow_listen_for_poll_requests.go
```diff
@@ -10,20 +10,27 @@ import (
 )
 
 func (c *Client) startPollerListener() error {
-	ctx := context.Background()
+	ctx, cancelCtx := context.WithCancel(context.Background())
```
mathnogueira, line 13: I cannot `defer cancelCtx` here because this function ends while the goroutine keeps going on. If I defer the call, the goroutine will break.

danielbdias: Can you add a comment explaining this situation on the code?

agent/client/workflow_listen_for_poll_requests.go
danielbdias: Does it make sense to add a comment here too?

agent/client/workflow_shutdown.go
```diff
@@ -0,0 +1,44 @@
...
+			if err == io.EOF {
+				return
+			}
+
+			if err != nil {
+				log.Fatal("could not get message from trigger stream: %w", err)
+			}
+
+			// TODO: get context from request
+			c.shutdownListener(context.Background(), &resp)
```
danielbdias, line 40: Do you plan to do that in another PR?

danielbdias (approved):

agent/client/workflow_shutdown.go
mathnogueira: yes, several places require this. I want to have a PR to instrument the agent

agent/client/workflow_listen_for_poll_requests.go
mathnogueira: managed to solve this issue in a better place, so no explanation is needed.

Issue #3020, opened 2023-08-02
Add Kafka trigger to Tracetest


danielbdias:
Following the research done on Spike https://github.com/kubeshop/tracetest/issues/2983, on this task we will develop the trigger for Kafka streams.

- [Specifications](https://github.com/kubeshop/tracetest/issues/2983#issuecomment-1657181053)
- [UX Design](https://github.com/kubeshop/tracetest/issues/2983#issuecomment-1658132767)
- [Feature branch](https://github.com/kubeshop/tracetest/tree/feature/kafka-trigger)

```[tasklist]
### Tasks
- [x] Add OpenAPI specs for Kafka trigger
- [x] Add Kafka trigger on Tracetest Server
- [x] Add support for Kafka trigger on Tracetest CLI
- [x] Add support for Kafka trigger on Tracetest WebUI
- [x] Add support for Kafka on Pokeshop
- [x] Setup Tracetest instalation to have a quick example for Kafka
- [x] Add trace based test for Kafka trigger
- [x] Add CLI e2e tests for Kafka trigger
- [ ] Add Kafka tests on OTel demo test
- [ ] Add documentation
```

xoscar:
Hey @danielbdias is this something we can close?

danielbdias:
Yes! @adnanrahic made the documentation about Kafka, and I'll add these tests on OTel soon (creating a ticket just for it).

PR #3024, opened 2023-08-02
feature(server): add kafka trigger

danielbdias:
This PR adds the Kafka trigger to OpenAPI specs and Tracetest Server.

## Changes

- Updated OpenAPI specs
- Added Kafka trigger

## Checklist

- [ ] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [ ] added a test


schoren (approved):
This looks awesome! Can you add a trace-based test for this? Not sure if we have a way to do it with the current test infra, but would be an awesome addition

danielbdias:
> This looks awesome! Can you add a trace-based test for this? Not sure if we have a way to do it with the current test infra, but would be an awesome addition

Sure @schoren ! I will do that as separated task on https://github.com/kubeshop/tracetest/issues/3020

PR #3028, opened 2023-08-03
feat(server): move otlp server data from runs to its own table

schoren:
This PR makes the OTLP server save data on its own table instead of directly updating the `run` table with the received spans.

This avoids bugs caused by moving the queues from in memory to external.

There wasn't a good place to put the `traces` repository needed, so I took the resources approach, moved the `Trace` out of the `model` package and into the `traces` package, and added the repo there. So, there are a lot of files changes that are just change the reference to the new `traces.Trace` and related

## Changes

- moved `model.Trace` and related structs into `traces.Trace`
- added a `traces.TraceRepository` used to store and access otlp server db data
- updated `otlp.Ingestor` to interface with the `traces.TraceRepository`

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/default_poller_executor.go
```diff
@@ -196,15 +201,15 @@ func (pe DefaultPollerExecutor) testConnection(ctx context.Context, traceDB trac
...
+func (pe DefaultPollerExecutor) donePollingTraces(job *Job, traceDB tracedb.TraceDB, trace traces.Trace) (bool, string) {
 	if !traceDB.ShouldRetry() {
 		return true, "TraceDB is not retryable"
 	}
 
 	maxTracePollRetry := job.PollingProfile.Periodic.MaxTracePollRetry()
 	// we're done if we have the same amount of spans after polling or `maxTracePollRetry` times
 	log.Printf("[PollerExecutor] Test %s Run %d: Job count %d, max retries: %d", job.Test.ID, job.Run.ID, job.EnqueueCount(), maxTracePollRetry)
-	if job.EnqueueCount() == maxTracePollRetry {
+	if job.EnqueueCount() >= maxTracePollRetry {
```
schoren, line 212: there was a weird bug where a job got incorrectly requeued after hitting the max retries limit. Since the new `EnqueueConunt` was bigger than `maxTracePollRetry` and the check was for equality, it got forever stuck in a polling loop

server/migrations/30_add_traces.up.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+CREATE TABLE "otlp_traces" (
+  "trace_id" varchar not null primary key,
+  "tenant_id" uuid,
+  "trace" JSONB,
+  "created_at" timestamp NOT NULL DEFAULT NOW()
```
schoren, line 7: this field gets autofilled on create. the rows are always inserted. for updates we use a delete+insert approach so we'll always have the correct value.

I think this might be useful for deleting old records from time to time

server/model/traceid.go
schoren, line 1: the code in this file was not being used

server/model/traces_test.go
schoren, line 1: this file was merged with this `server/traces/traces_test.go`

server/otlp/ingester.go
```diff
@@ -55,69 +83,54 @@ func (i ingester) Ingest(ctx context.Context, request *pb.ExportTraceServiceRequ
...
+		// trace is not part of any known test run, no need to notify
 		return nil
 	}
-
 	if err != nil {
-		return fmt.Errorf("could not find test run with traceID %s: %w", traceID.String(), err)
-	}
-
-	if run.State != test.RunStateAwaitingTrace {
-		// test is not waiting for trace, so we can completely ignore those as they might
```
schoren, line 96: test integrity is now safe, since the results are saved to another table. the ingestor has no impact on the runs table

server/traces/otel_converter.go
```diff
@@ -6,37 +6,40 @@ import (
...
 	for _, resource := range resourceSpans {
 		for _, scopeSpans := range resource.ScopeSpans {
 			flattenSpans = append(flattenSpans, scopeSpans.Spans...)
 		}
 	}
 
+	return FromSpanList(flattenSpans)
+}
+
+func FromSpanList(input []*v1.Span) Trace {
```
schoren, line 29: this was extracted as an exported function so it could be reused by the ingestor more easily

server/traces/traces_test.go
```diff
@@ -6,13 +6,253 @@ import (
...
-	"github.com/kubeshop/tracetest/server/model"
 	"github.com/kubeshop/tracetest/server/pkg/id"
+	"github.com/kubeshop/tracetest/server/traces"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/otel/trace"
+	v1 "go.opentelemetry.io/proto/otlp/trace/v1"
 )
 
+func TestTraces(t *testing.T) {
```
schoren, line 17: this was merged from `server/model/traces_test.go`

xoscar (approved):
Great job @schoren I really like that you moved the traces model and logic to its own package!

server/traces/trace_repository.go
```diff
@@ -0,0 +1,110 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to delete trace: %w", err)
+	}
+
+	_, err = tx.ExecContext(
+		ctx,
+		`INSERT INTO otlp_traces (tenant_id, trace_id, trace) VALUES ($1, $2, $3)`,
+		sqlutil.TenantID(ctx),
+		trace.ID.String(),
```
danielbdias, line 96: Philosophical question: should we rely on each vendor to have a unique TraceID?

danielbdias (approved):

jorgeepc (approved):

PR #3030, opened 2023-08-04
feat(server): implement postgres queue driver

schoren:
This PR implements a postgres queue driver

## Changes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/queue_driver_in_memory.go
```diff
@@ -0,0 +1,61 @@
...
+	"log"
+)
+
+type loggerFn func(string, ...any)
+
+func newLoggerFn(name string) loggerFn {
+	return func(format string, params ...any) {
+		log.Printf("[%s] %s", name, fmt.Sprintf(format, params...))
+	}
+}
```
danielbdias, line 14: This is useful even outside of this package. Perhaps we could move it to a specific infra package in the future?

server/executor/queue_driver_postgres.go
```diff
@@ -0,0 +1,182 @@
...
+		Channel: ch.name,
+		Job:     job,
+	})
+
+	if err != nil {
+		ch.log("error marshalling pgJob: %s", err.Error())
+		return
+	}
+
+	ctx, cancelCtx := context.WithTimeout(context.Background(), 500*time.Millisecond)
```
danielbdias, line 165: Can we add this `500ms` timeout as a constant? Something like `enqueueTimeout`?

danielbdias (approved):

server/executor/queue_driver_in_memory.go
schoren: yes, I found the pattern very useful, but if we want to improve our logging I would make it a bit more sophisticated

server/executor/queue_driver_postgres.go
schoren: done!

xoscar (approved):

PR #3033, opened 2023-08-07
feat(backend): Updating transactions to test suites

xoscar:
This PR updates the database, controllers, services, and all of that to use the new test suites instead of transactions.

## Changes

- Adds migration scripts to update tables and columns
- Updates models and repositories with the new naming
- Updates unit tests to work with new naming
- Changes open API specs to reflect naming changes

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


api/openapi.yaml
```diff
@@ -36,138 +36,138 @@ paths:
...
+  /testsuites/{testSuiteId}:
     get:
       tags:
         - resource-api
       parameters:
-        - $ref: "./parameters.yaml#/components/parameters/transactionId"
-      summary: "get transaction"
-      description: "get transaction"
-      operationId: getTransaction
+        - $ref: "./parameters.yaml#/components/parameters/testsuiteId"
```
schoren, line 77: if we're using `TestSuite`, shouldn't the parm be `testSuitId`? looks more consistent IMO

server/http/controller.go
```diff
@@ -579,20 +579,20 @@ func (c *controller) DeleteTransactionRun(ctx context.Context, transactionId str
 func (c *controller) GetResources(ctx context.Context, take, skip int32, query, sortBy, sortDirection string) (openapi.ImplResponse, error) {
 	// TODO: this is endpoint is a hack to unblock the team quickly.
 	// This is not production ready because it might take too long to respond if there are numerous
-	// transactions and transaction.
+	// transactions and testsuite.
```
schoren, line 582:
```suggestion
	// tests and testsuite.
```

schoren (approved):

api/openapi.yaml
xoscar: Makes sense, I'm going to fix it as part of the frontend PR so I can merge this already into the feature branch

PR #3047, opened 2023-08-09
fix(docs): fixing docs build and variable sets

xoscar:
This PR fixes the docs build as it was failing due to some recent changes. It also updates the environment references to variable sets

## Changes

- Fixes build
- Moves environment pages to variable sets
- Adds new screenshots

## Fixes

- docs build

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

https://www.loom.com/share/e8d4feeda9404a2eac319ce41c7d98b4


docs/docs/cli/creating-variable-sets.md
```diff
@@ -0,0 +1,31 @@
...
+      value: mysecret
+```
+
+In order to apply this configuration to your Tracetest instance, make sure to have your [CLI configured](./configuring-your-cli.md) and run:
+
+```sh
+tracetest apply variableset -f <variableset.yaml>
+```
+
+> If the file contains the property `spec.id`, the operation will be considered a variable set update. If you try to apply a variable set and you get an error: `could not apply variableset: 404 Not Found`, it means the provided id doesn't exist. Either update the id to reference an existing variable set, or just remove the property from the file, so Tracetest will create a new variable set and a new id.
```
jfermi, line 31:
```suggestion
> If the file contains the property `spec.id`, the operation will be considered a variable set update. If you try to apply a variable set and you get the error: `could not apply variableset: 404 Not Found`, it means the provided ID doesn't exist. Either update the ID to reference an existing variable set or remove the property from the file. Tracetest will create a new variable set and a new ID.
```

docs/docs/concepts/transactions.md
```diff
@@ -29,7 +29,7 @@ outputs:
       expression: attr:tracetest.time.start
 ```
 
-This would create an output called `TIME_CANCEL_SUBSCRIPTION_MESSAGE_OBTAINED` that is obtained by reading the attribute `tracetest.time.start` from the span with `name` equal to `Process request from cancel subscription queue`. This value would would then be injected into the environment variables of that transaction to be accessed by other tests within the same transaction run.
+This would create an output called `TIME_CANCEL_SUBSCRIPTION_MESSAGE_OBTAINED` that is obtained by reading the attribute `tracetest.time.start` from the span with `name` equal to `Process request from cancel subscription queue`. This value would would then be injected into the variables of that transaction to be accessed by other tests within the same transaction run.
```
jfermi, line 32:
```suggestion
This would create an output called `TIME_CANCEL_SUBSCRIPTION_MESSAGE_OBTAINED` that is obtained by reading the attribute `tracetest.time.start` from the span with `name` equal to `Process request from cancel subscription queue`. This value would then be injected into the variables of that transaction to be accessed by other tests within the same transaction run.
```

mathnogueira (approved):

PR #3072, opened 2023-08-16
fix(ui): kafka trigger with null header blocks test render on WebUI

danielbdias:
This PR fixes a UI bug caused by tests with Kafka trigger and no headers.

Loom: https://www.loom.com/share/f246f488f58d4aefa2fe34a67bb6b6e4?sid=8753e331-70b6-4462-92a5-b1116ed0c23a

## Changes

- Add guard clause on KafkaRequest model
- Removed unused field that is blocking `demo` deploy

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/models/KafkaRequest.model.ts
+++ b/web/src/models/KafkaRequest.model.ts
@@ -19,10 +19,6 @@ const KafkaRequest = ({
   messageKey = '',
   messageValue = '',
 }: TRawKafkaRequest): KafkaRequest => {
-  if (!headers) {
-    headers = []; // guard clause to avoid null header
-  }
-
   return {
     brokerUrls,
     topic,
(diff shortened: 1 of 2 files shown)


web/src/models/KafkaRequest.model.ts
```diff
@@ -19,6 +19,10 @@ const KafkaRequest = ({
   messageKey = '',
   messageValue = '',
 }: TRawKafkaRequest): KafkaRequest => {
+  if (!headers) {
```
xoscar, line 22: We try to avoid side effects like this, I think it is better to have something like
```
headers: !headers ? [] : ...
```

danielbdias: It makes sense. I've updated this code to something similar, using the notation `(headers || [])` when building the model: 
```js

    headers: (headers || []).map(({key = '', value = ''}) => ({
```

xoscar (approved):

PR #3098, opened 2023-08-23
fix panic when under pressure

schoren:
This PR fixes the performace issues detected by running k6 tests

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/queue_driver_postgres.go
```diff
@@ -111,7 +111,7 @@ func (qd *PostgresQueueDriver) worker(conn *pgxpool.Conn) {
 	}
 
 	// spin off so we can keep listening for jobs
-	go channel.q.Listen(job.Job)
+	channel.listener.Listen(job.Job)
 	qd.log("spun off job for channel: %s, runID: %d", job.Channel, job.Job.Run.ID)
```
danielbdias, line 115: 🙃 

```suggestion
	qd.log("spin off job for channel: %s, runID: %d", job.Channel, job.Job.Run.ID)
```

testing/load/load-test.js
```diff
@@ -0,0 +1,68 @@
...
+  thresholds: {
+    http_req_duration: ["p(95)<500"],
+  },
+};
+
+const tracetest = Tracetest({
+  serverUrl: "http://localhost:11633",
+});
+const testId = "kc_MgKoVR";
+let pokemonId = 6; // charizad
```
danielbdias, line 22: nit: Just because my OCD: 

```suggestion
let pokemonId = 6; // charizard
```

testing/load/run.bash
```diff
@@ -0,0 +1,51 @@
...
+DOCKER_LOG=/tmp/docker-log
+# printDockerLog() {
+#   $DOCKER_COMPOSE kill
+#   echo
+#   echo "Error occured. Printing docker log ($DOCKER_LOG):"
+#   cat $DOCKER_LOG
+# }
+
+# print docker log on any error
+# trap 'printDockerLog' ERR
```
danielbdias, line 30: Does it make sense to keep this code yet?

testing/load/run.bash
```diff
@@ -0,0 +1,51 @@
...
+$TRACETEST apply test -f tracetest-test.yaml
+
+rm -f ./k6
+currentDir=$(pwd)
+dir=$(mktemp -d)
+cd $dir
+go install go.k6.io/xk6/cmd/xk6@latest
+xk6 build v0.42.0 --with github.com/kubeshop/xk6-tracetest \
+  --replace go.buf.build/grpc/go/prometheus/prometheus=buf.build/gen/go/prometheus/prometheus/protocolbuffers/go@latest \
+  --replace go.buf.build/grpc/go/gogo/protobuf=buf.build/gen/go/gogo/protobuf/protocolbuffers/go@latest
```
danielbdias, line 45: It could be good to add a comment here explaining that we must compile this code outside of `tracetest` dir to avoid a build error due `go.work`.

danielbdias (approved):

server/subscription/manager.go
```diff
@@ -4,54 +4,51 @@ import "sync"
...
+	m.subscriptions[resourceID] = subscribers
+}
 
-	array, exists := m.subscriptions[resourceID]
-	if !exists {
-		return
-	}
+func (m *Manager) Subscribe(resourceID string, subscriber Subscriber) {
+	subscribers := append(m.getSubscribers(resourceID), subscriber)
+	m.setSubscribers(resourceID, subscribers)
```
mathnogueira, line 31: Don't you need to lock the mutex here to prevent read-write race conditions here? Go will panic in case of read-write concurrency: https://stackoverflow.com/questions/36167200/how-safe-are-golang-maps-for-concurrent-read-write-operations

Same for PublishUpdate

server/executor/queue_driver_postgres.go
```diff
@@ -140,20 +141,20 @@ func (qd *PostgresQueueDriver) Channel(name string) *channel {
...
 }
 
-func (ch *channel) SetQueue(q *Queue) {
-	ch.q = q
+func (ch *channel) SetListener(l Listener) {
+	ch.listener = l
 }
 
-const enqueueTimeout = 500 * time.Millisecond
+const enqueueTimeout = 5 * time.Minute
```
danielbdias, line 154: This line owes me two days of life and one year of therapy.

mathnogueira (changes requested):
small thing but it can panic the server

jorgeepc (approved):
LGTM!

server/subscription/manager.go
schoren: Originally each function that needed to access the map would manage the lock itself. That code was causing issues for some reason, so I moved the usage of the lock to the [getSubscribers](https://github.com/kubeshop/tracetest/pull/3098/files/52bc8cfbad5e4cf9d39692d3cf5c4f9744b75957#diff-e5da02bda163f4561d8b7d1d92e295305327f3afd4235a8de026fd2a5f31da7aR17-R19) function and the [setSubscribers](https://github.com/kubeshop/tracetest/pull/3098/files/52bc8cfbad5e4cf9d39692d3cf5c4f9744b75957#diff-e5da02bda163f4561d8b7d1d92e295305327f3afd4235a8de026fd2a5f31da7aR23-R25) function, so effectively every read and write is lock protected

mathnogueira (approved):
Lgtm

PR #3113, opened 2023-08-31
feat(queue): send test version to queue

mathnogueira:
This PR sends the test version to the executor queue

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/executor/queue.go
+++ b/server/executor/queue.go
@@ -319,11 +319,16 @@ func (q Queue) Enqueue(ctx context.Context, job Job) {
 		propagator().Inject(ctx, propagation.MapCarrier(*job.Headers))
 		job.Headers.Set("InstanceID", q.instanceID)
 
+		version := 1
+		if job.Test.Version != nil {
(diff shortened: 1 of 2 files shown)


server/executor/queue.go
```diff
@@ -319,11 +319,16 @@ func (q Queue) Enqueue(ctx context.Context, job Job) {
...
+		if job.Test.Version != nil {
+			version = *job.Test.Version
+		}
+
 		newJob := Job{
 			Headers: job.Headers,
 
 			Test: test.Test{ID: job.Test.ID},
-			Run:  test.Run{ID: job.Run.ID},
+			Run:  test.Run{ID: job.Run.ID, TestVersion: version},
```
mathnogueira, line 331: We need the test version to ensure we are loading the correct test spec when running the test.

xoscar (approved):

jorgeepc (approved):

PR #3120, opened 2023-09-04
chore(server): split trigger worker into 3 workers

mathnogueira:
This PR splits the trigger worker into 3 workers:

1. Trigger resolution
2. Trigger execution
3. Trigger result processing

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/executor/poller_executor_test.go
```diff
@@ -540,7 +540,7 @@ func getRunRepositoryMock(t *testing.T) *testdb.MockRepository {
 
 	testDB := new(testdb.MockRepository)
 	testDB.Test(t)
-	testDB.On("CreateTestRunEvent", mock.Anything).Return(noError)
+	testDB.On("CreateTestRunEvent", mock.Anything).Return(nil)
```
schoren, line 543: isn't this the same? `noError` is a pretty name for a nil error IIRC

schoren (approved):

PR #3121, opened 2023-09-04
feat(cli): cleanup and dynamic endpoint support

xoscar:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

https://www.loom.com/share/650c1c542a734e8c962e9300526f8958

### DIFF
--- a/cli/cmd/resources.go
+++ b/cli/cmd/resources.go
@@ -38,33 +38,6 @@ var (
 var (
 	httpClient = &resourcemanager.HTTPClient{}
 
-	organizationsClient = resourcemanager.NewClient(
-		httpClient, cliLogger,
-		"organization", "organizations",
-		resourcemanager.WithTableConfig(resourcemanager.TableConfig{
-			Cells: []resourcemanager.TableCellConfig{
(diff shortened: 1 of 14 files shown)


cli/cmd/resources.go
```diff
@@ -286,6 +259,31 @@ var (
...
+				{Header: "ID", Path: "id"},
+				{Header: "NAME", Path: "name"},
+			},
+		}),
+		resourcemanager.WithListPath("elements"),
+	)
+
+	environmentClient = resourcemanager.NewClient(
+		httpClient, cliLogger,
+		"env", "environments",
```
jorgeepc, line 277: is `env` correct here?

jorgeepc (approved):
Looks good. One question, are we protecting the Configure cmd with the flag, or is not needed?

cli/cmd/resources.go
xoscar: Yes, if we use envrionment it collides with the other resource 🥴

PR #3135, opened 2023-09-08
chore(server): make queue drivers generic

schoren:
This PR extracts the Queue module including its drivers into a reusable package through the use of generics. This sets the field to allow easy reuse in other modules


agent/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubeshop/tracetest/agent
 
-go 1.20
+go 1.21
```
danielbdias, line 3: ❤️

danielbdias (approved):

server/pkg/pipeline/pipeline.go
```diff
@@ -1,49 +1,60 @@
...
-	queues []*Queue       // N + 1
+type loggerFn func(string, ...any)
+
+func newLoggerFn(name string) loggerFn {
+	return func(format string, params ...any) {
+		log.Printf("[%s] %s", name, fmt.Sprintf(format, params...))
+	}
+}
+
+type Pipeline[T any] struct {
```
danielbdias, line 18: I liked this approach of turning the Pipeline generic. This will help us to deal with other queues in the future.

jorgeepc (approved):

PR #3137, opened 2023-09-08
fix: move trigger state check to post process run and fix small issues

mathnogueira:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


api/tests.yaml
```diff
@@ -28,7 +28,7 @@ components:
       type: object
       properties:
         id:
-          type: string
+          type: integer
```
mathnogueira, line 31: I was getting an error using the tracetest client because of this.

server/test/trigger/trigger.go
```diff
@@ -15,7 +15,7 @@ type (
 
 	TriggerResult struct {
 		Type    TriggerType      `json:"type"`
-		HTTP    *HTTPResponse    `json:"httpRequest,omitempty"`
+		HTTP    *HTTPResponse    `json:"http,omitempty"`
```
mathnogueira, line 18: `httpRequest` is only used in `Trigger`, here is the response.

danielbdias (approved):

PR #3161, opened 2023-09-15
feat(server): add options for test pipeline

danielbdias:
This PR adds options to enable/disable trace fetch and trigger execution.

## Changes

- Added option for test pipeline

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/server/executor/tracepollerworker/fetcher_worker.go
+++ b/server/executor/tracepollerworker/fetcher_worker.go
@@ -16,6 +16,7 @@ import (
 type traceFetcherWorker struct {
 	state       *workerState
 	outputQueue pipeline.Enqueuer[executor.Job]
+	enabled     bool
 }
 
 func NewFetcherWorker(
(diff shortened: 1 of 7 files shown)


server/executor/tracepollerworker/fetcher_worker.go
```diff
@@ -47,6 +49,11 @@ func (w *traceFetcherWorker) SetOutputQueue(queue pipeline.Enqueuer[executor.Job
 }
 
 func (w *traceFetcherWorker) ProcessItem(ctx context.Context, job executor.Job) {
+	if !w.enabled {
+		w.outputQueue.Enqueue(ctx, job)
```
schoren, line 53: shouldn't this be just a skip? if we have an external queue processor that also listens for this step and then passes it to the next, the item will be duplicated in the queue

danielbdias: I agree with you, changing it now.

PR #3166, opened 2023-09-18
chore(server): Updating data store test to use queues

xoscar:
This PR updates the data test connection logic to use queues

## Changes

- Implements ds test connection pipeline
- Updates logic to be async

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


danielbdias (approved):

danielbdias (approved):

agent/proto/orchestrator.proto
```diff
@@ -170,6 +176,32 @@ message TraceIdResponse {
     string id = 1;
 }
 
+message DataStoreConnectionTestRequest {
+    string requestID = 1;
+    DataStore datastore = 5;
```
jorgeepc, line 181:
```suggestion
    DataStore datastore = 2;
```
??

jorgeepc (approved):

server/http/controller.go
```diff
@@ -78,6 +83,13 @@ type transactionRunner interface {
 	Run(context.Context, testsuite.TestSuite, test.RunMetadata, variableset.VariableSet, *[]testrunner.RequiredGate) testsuite.TestSuiteRun
 }
 
+type dataStoreTestRunner interface {
```
schoren, line 86: name suggestion
```suggestion
type dataStoreTester interface {
```

server/http/controller.go
```diff
@@ -672,27 +688,26 @@ func takeResources(transactions []testsuite.TestSuite, tests []test.Test, take,
...
-	if err := ds.Validate(); err != nil {
-		return openapi.Response(http.StatusBadRequest, err.Error()), err
-	}
+	job := c.dsTestPipeline.NewJob(ds)
 
-	tdb, err := c.newTraceDBFn(ds)
-	if err != nil {
-		return openapi.Response(http.StatusBadRequest, err.Error()), err
-	}
+	job.Headers.Set(string(middleware.TenantIDKey), middleware.TenantIDFromContext(ctx))
```
schoren, line 693: this is not required

server/testconnection/pipeline.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	job.ID = uuid.New().String()
+	job.DataStore = datastore
+	job.Headers[string(middleware.TenantIDKey)] = middleware.TenantIDFromContext(ctx)
+
+	return job
+}
+
+func (p *DataStoreTestPipeline) Run(ctx context.Context, job Job) {
+	spew.Dump(job)
```
schoren, line 62:
```suggestion
```

server/testconnection/pipeline.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (p *DataStoreTestPipeline) NewJob(ctx context.Context, datastore datastore.DataStore) Job {
+	job := Job{
+		Headers: make(map[string]string),
+	}
+	job.ID = uuid.New().String()
+	job.DataStore = datastore
+	job.Headers[string(middleware.TenantIDKey)] = middleware.TenantIDFromContext(ctx)
+
+	return job
```
schoren, line 57:
```suggestion
	return Job{
		ID: uuid.New().String()
		DataStore: datastore
		Headers:= map[string]string{
			string(middleware.TenantIDKey): middleware.TenantIDFromContext(ctx)
		}
	}
```

server/testconnection/pipeline.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/kubeshop/tracetest/server/http/middleware"
+	"github.com/kubeshop/tracetest/server/model"
+	"github.com/kubeshop/tracetest/server/pkg/pipeline"
+	"github.com/kubeshop/tracetest/server/tracedb"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type Job struct {
+	Headers    map[string]string      `json:"headers"`
```
schoren, line 18: are the headers used for anything other than passing the tenant ID? if so, it might be simpler to just have a `TenantID string`  field

xoscar: I wanted to keep using the same approach we have for the rest, but yeah we can just change it for a new field, let me check..

server/testconnection/listener.go
```diff
@@ -0,0 +1,38 @@
...
+	defer m.mutex.Unlock()
+
+	m.subscriptions[jobID] = append(m.subscriptions[jobID], notifier)
+}
+
+func (m *Listener) Unsubscribe(jobID string) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	delete(m.subscriptions, jobID)
```
schoren, line 31: for each job ID there is a slice of subscribers, implying that you can have multiple subscribers for a single job. When any client calls `Unsubscribe`, since this function is deleting **all** registered subscribers, all subscribers will stop being notified.

It might make more sense to have a single listener per job, and have some logic to avoid subscribing to a job that already has a subscriber. Another, more complex option is to assign an ID for every subscriber func and track that for unsubscriptions, but seems overkill for this usecase

xoscar: This is meant to be a 1-1 so I'll remove the string slice for the moment

server/testconnection/request.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (w *dsTestConnectionRequest) ProcessItem(ctx context.Context, job Job) {
+	if !w.enabled {
+		return
+	}
+
+	ctx, pollingSpan := w.tracer.Start(ctx, "dsTestConnectionRequest.ProcessItem")
+	defer pollingSpan.End()
+
+	traceDB, err := getTraceDB(job.DataStore, w.newTraceDBFn)
```
schoren, line 42: If i understand this correctly, this function's only effect is to wrap the error with a message. I'd recommend to just have this behavior inlined here so it's easier to follow

server/testconnection/listener.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (m *Listener) Notify(job Job) {
+	for _, sub := range m.subscriptions[job.ID] {
+		sub(job)
+	}
+}
```
schoren, line 38: If not using locks for reading there can still be race conditions

```suggestion
func (m *Listener) Notify(job Job) {
	m.mutex.Lock()
	defer m.mutex.Unlock()

	for _, sub := range m.subscriptions[job.ID] {
		sub(job)
	}
}
```

server/testconnection/listener.go
schoren: make sense

schoren (changes requested):
My comments are mostly optiona, except for the one about the race conditoin. that needs to be fixed or it might cause panics on prod

schoren (approved):

PR #3175, opened 2023-09-21
feat(agent): keep track of spans that were generated by tests

mathnogueira:
This PR makes the agent's internal collector keep track of spans that are generated by tests triggered by the agent.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/agent/workers/trigger.go
+++ b/agent/workers/trigger.go
@@ -4,25 +4,25 @@ import (
 	"context"
 	"fmt"
 
-	gocache "github.com/Code-Hex/go-generics-cache"
 	"github.com/kubeshop/tracetest/agent/client"
+	"github.com/kubeshop/tracetest/agent/collector"
 	"github.com/kubeshop/tracetest/agent/proto"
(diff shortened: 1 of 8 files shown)


agent/workers/trigger.go
```diff
@@ -42,6 +62,10 @@ func (w *TriggerWorker) Trigger(ctx context.Context, triggerRequest *proto.Trigg
 		return fmt.Errorf("invalid traceID was received in TriggerRequest: %w", err)
 	}
 
+	// Set traceID to cache so the collector starts watching for incoming traces
+	// with same id
+	w.traceCache.Set(triggerRequest.TraceID, []*v1.Span{})
```
xoscar, line 67: I guess we don't know what ds is used at this point right?

xoscar (approved):

agent/workers/trigger.go
mathnogueira: yeah, we don't :(

PR #3176, opened 2023-09-21
feat(agent): in-memory datastore to support otlp datastores

mathnogueira:
This PR makes the agent use an in-memory datastore that is populated by the collector when receiving traces in the OTLP endpoint.

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


xoscar (approved):
Amazing job @mathnogueira!

PR #3177, opened 2023-09-21
fix(agent): updating support for otlp datastore

xoscar:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/agent/collector/ingester.go
+++ b/agent/collector/ingester.go
@@ -2,8 +2,6 @@ package collector
 
 import (
 	"context"
-	"fmt"
-	"log"
 	"sync"
 	"time"
 
@@ -11,8 +9,6 @@ import (
 	"go.opencensus.io/trace"
 	pb "go.opentelemetry.io/proto/otlp/collector/trace/v1"
 	v1 "go.opentelemetry.io/proto/otlp/trace/v1"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials/insecure"
 )
 
 type stoppable interface {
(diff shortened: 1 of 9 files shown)


cli/pkg/starter/starter.go
```diff
@@ -100,6 +100,10 @@ func (s *Starter) StartAgent(ctx context.Context, endpoint, name, agentApiKey, u
 		ServerURL: endpoint,
 		APIKey:    agentApiKey,
 		Name:      name,
+		OTLPServer: agentConfig.OtlpServer{
```
xoscar, line 103: This needs to be configurable from the CLI

server/app/app.go
```diff
@@ -230,8 +230,8 @@ func (app *App) Start(opts ...appOption) error {
 
 	tracedbFactory := tracedb.Factory(tracesRepo)
 
-	eventEmitter := executor.NewEventEmitter(testDB, subscriptionManager)
-	registerOtlpServer(app, tracesRepo, runRepo, eventEmitter, dataStoreRepo, tracer)
+	// eventEmitter := executor.NewEventEmitter(testDB, subscriptionManager)
```
xoscar, line 233: We might need a flag to enable/disable this (?)

agent/collector/ingester.go
```diff
@@ -28,13 +24,6 @@ func newForwardIngester(ctx context.Context, batchTimeout time.Duration, remoteI
 		traceCache:     remoteIngesterConfig.traceCache,
 	}
 
-	err := ingester.connectToRemoteServer(ctx)
```
xoscar, line 31: Having a connect here was failing the initialization and we are not connecting to a server "yet"

mathnogueira: We might need to disable all the removed code based on a flag. 

The removed code is important in the future when we decide to forward traces from the agent to an OTLP-compatible endpoint. But for now, I think we can remove it.

server/app/app.go
mathnogueira: Yes. If we cant have it on cloud, we should disable it via configuration

mathnogueira (approved):

PR #3180, opened 2023-09-22
fix(agent): fixing connection error check

xoscar:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/agent/workers/poller.go
+++ b/agent/workers/poller.go
@@ -10,9 +10,9 @@ import (
 	"github.com/fluidtruck/deepcopy"
 	"github.com/kubeshop/tracetest/agent/client"
 	"github.com/kubeshop/tracetest/agent/proto"
-	"github.com/kubeshop/tracetest/agent/workers/datastores/connection"
 	"github.com/kubeshop/tracetest/server/datastore"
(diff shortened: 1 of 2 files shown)


agent/workers/poller.go
```diff
@@ -10,9 +10,9 @@ import (
 	"github.com/fluidtruck/deepcopy"
 	"github.com/kubeshop/tracetest/agent/client"
 	"github.com/kubeshop/tracetest/agent/proto"
-	"github.com/kubeshop/tracetest/agent/workers/datastores/connection"
 	"github.com/kubeshop/tracetest/server/datastore"
 	"github.com/kubeshop/tracetest/server/tracedb"
+	"github.com/kubeshop/tracetest/server/tracedb/connection"
```
xoscar, line 15: We are pointing everything to the server trace db logic, in the future we have to kill that package to use the one from the agent, the polling error was caused because of the errorTraceNotFound instance was different between server/tracedb and agent/datastores package

mathnogueira (approved):

schoren (approved):

jorgeepc (approved):

PR #3189, opened 2023-09-26
feat(agent): docker image

mathnogueira:
This PR adds the dockerfile for the agent
## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -1,4 +1,4 @@
-FROM kubeshop/tracetest:v0.13.9-rc.7
+FROM kubeshop/tracetest:latest
 WORKDIR /app
 
 ENV TRACETEST_API_KEY ""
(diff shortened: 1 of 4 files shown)


agent/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM kubeshop/tracetest:v0.13.9
```
schoren, line 1: this looks like a very easy thing to forget to update on every release. can we make it dynamic somehow?

mathnogueira: do you think it's worth adding it to the pipeline now and make it build the same way our main image?

schoren: yea, totally, no time like the present

danielbdias (approved):

agent/Dockerfile
mathnogueira: Done!

PR #3190, opened 2023-09-27
feat(agent): k8s deployment script

mathnogueira:
This PR adds the k8s manifest to deploy the agent in a cluster

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/k8s/agent/deploy-agent.yaml
+++ b/k8s/agent/deploy-agent.yaml
@@ -1,0 +1,53 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tracetest-agent
+  labels:
+    app: tracetest-agent
+spec:
+  selector:
+    matchLabels:
+      app: tracetest-agent
+  template:
+    metadata:
+      labels:
+        app: tracetest-agent
+    spec:
+      containers:
+        - name: tracetest-agent
+          image: kubeshop/tracetest-agent:v0.13.9
+          env:
+            - name: TRACETEST_API_KEY
+              valueFrom:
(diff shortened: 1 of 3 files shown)


k8s/agent/deploy-agent.yaml
```diff
@@ -0,0 +1,53 @@
...
+    matchLabels:
+      app: tracetest-agent
+  template:
+    metadata:
+      labels:
+        app: tracetest-agent
+    spec:
+      containers:
+        - name: tracetest-agent
+          image: kubeshop/tracetest-agent:v0.13.9
```
mathnogueira, line 19: Should we use `latest` here?

danielbdias: It may be a good idea. I believe that it is easy to forget to update this version.

danielbdias, line 19: Thinking also about the version: is there a way to make this version configurable? My idea is to think a way to install a specific version of the agent in a cluster.

danielbdias (approved):

xoscar (approved):

k8s/agent/deploy-agent.yaml
mathnogueira: Done

mathnogueira: By default, `kubeshop/tracetest-agent:latest` will be used. 

If the user provides a third option with the version, that version will be used instead:

```
curl <path> | bash -s -- <namespace> <api-key> (<version>?)
```

PR #3196, opened 2023-09-28
fix(frontend): update header back button logic

### LINKED ISSUE #3178: Fix arrow behaviour ASAP
if user open a specific version of test from test detail page we should back user on this page, when user open a test from main page we should back user on the main page when he click on back butoon

Image

jorgeepc:
This PR fixes the back button behavior in the test run detail page.

## Changes

- back button in test run detail page

## Fixes

- fixes #3178

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/web/src/components/RunActionsMenu/RunActionsMenu.tsx
+++ b/web/src/components/RunActionsMenu/RunActionsMenu.tsx
@@ -14,9 +14,10 @@ interface IProps {
   isRunView?: boolean;
   testSuiteId?: string;
   testSuiteRunId: number;
+  origin?: string;
 }
 
-const RunActionsMenu = ({resultId, testId, testSuiteId, testSuiteRunId, isRunView = false}: IProps) => {
+const RunActionsMenu = ({resultId, testId, testSuiteId, testSuiteRunId, isRunView = false, origin}: IProps) => {
   const {getIsAllowed} = useCustomization();
   const {onJUnit} = useFileViewerModal();
   const {navigate} = useDashboard();
@@ -52,7 +53,7 @@ const RunActionsMenu = ({resultId, testId, testSuiteId, testSuiteRunId, isRunVie
               data-cy="automate-test-button"
               key="automate-test"
               onClick={() => {
-                navigate(`/test/${testId}/run/${resultId}/automate`);
+                navigate(`/test/${testId}/run/${resultId}/automate`, {state: {origin}});
               }}
             >
               Automate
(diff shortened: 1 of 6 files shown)


web/src/components/RunDetailLayout/RunDetailLayout.tsx
```diff
@@ -42,6 +46,12 @@ const RunDetailLayout = ({test: {id, name, trigger}, test}: IProps) => {
   const [prevState, setPrevState] = useState(run.state);
   useDocumentTitle(`${name} - ${run.state}`);
 
+  const location = useLocation();
+  const [origin] = useState(() => {
+    const state = location.state as CustomizedState;
```
xoscar, line 51: Is there a need for the use state here?

web/src/components/RunActionsMenu/RunActionsMenu.tsx
```diff
@@ -52,7 +53,7 @@ const RunActionsMenu = ({resultId, testId, testSuiteId, testSuiteRunId, isRunVie
               data-cy="automate-test-button"
               key="automate-test"
               onClick={() => {
-                navigate(`/test/${testId}/run/${resultId}/automate`);
+                navigate(`/test/${testId}/run/${resultId}/automate`, {state: {origin}});
```
xoscar, line 56: What happens if the origin is empty or undefined?

xoscar (approved):
Hey @jorgeepc I left some questions, let me know what you think

web/src/components/RunDetailLayout/RunDetailLayout.tsx
jorgeepc: We need the state because we only want to extract the `origin` value the first time the page is loaded. Every time the `route` is updated the location object changes too.

web/src/components/RunActionsMenu/RunActionsMenu.tsx
jorgeepc: Nothing. We default it to the root page when we do the navigation.

web/src/components/RunDetailLayout/RunDetailLayout.tsx
xoscar: That's very interesting, what would be the advantages and disadvantages from `useMemo`?

jorgeepc:
close in favor of #3199

PR #3197, opened 2023-09-29
fix(server): Fixing nil Trace

xoscar:
This PR adds a check to validate if a trace is `nil` before starting to manipulate it as part of the evaluation process

## Changes

- Adds trace nil validation check

## Fixes

- https://github.com/kubeshop/tracetest-cloud/issues/140

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/executor/tracepollerworker/evaluator_worker.go
+++ b/server/executor/tracepollerworker/evaluator_worker.go
@@ -142,6 +142,11 @@ func (w *tracePollerEvaluatorWorker) ProcessItem(ctx context.Context, job execut
 	log.Printf("[TracePoller] Test %s Run %d: Done polling. (%s)", job.Test.ID, job.Run.ID, reason)
 
 	log.Printf("[TracePoller] Test %s Run %d: Start Sorting", job.Test.ID, job.Run.ID)
+	if job.Run.Trace == nil {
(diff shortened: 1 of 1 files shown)


server/executor/tracepollerworker/evaluator_worker.go
```diff
@@ -142,6 +142,11 @@ func (w *tracePollerEvaluatorWorker) ProcessItem(ctx context.Context, job execut
 	log.Printf("[TracePoller] Test %s Run %d: Done polling. (%s)", job.Test.ID, job.Run.ID, reason)
 
 	log.Printf("[TracePoller] Test %s Run %d: Start Sorting", job.Test.ID, job.Run.ID)
+	if job.Run.Trace == nil {
+		newTrace := traces.NewTrace(job.Run.TraceID.String(), []traces.Span{})
+		job.Run.Trace = &newTrace
+	}
```
xoscar, line 148: There was a problem caused by this change:
https://github.com/kubeshop/tracetest/pull/3170/files#diff-8477c1b1ce0272a42d729b310d1dacfd4452b274c8de5e0737d701593965b150R84

Which make every incoming trace without spans to be marked as nil, but for no tracing mode, this is expected

jorgeepc (approved):

PR #3219, opened 2023-10-05
fix(agent): Graceful Stop

xoscar:
This PR fixes the problem when trying to stop the agent from the CLI option

## Changes

- Adds missing channel creation
- Validates if the error is context stopped to not yield the errors to the console

## Fixes

- https://github.com/kubeshop/tracetest-cloud/issues/147

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

https://www.loom.com/share/d0bf6ba20a074b16811e5524b20bb36f

### DIFF
--- a/agent/client/client.go
+++ b/agent/client/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/kubeshop/tracetest/agent/proto"
@@ -38,6 +39,7 @@ func (c *Client) Start(ctx context.Context) error {
 		return err
 	}
 
+	c.done = make(chan bool)
 	ctx, cancel := context.WithCancel(ctx)
 	go func() {
 		<-c.done
@@ -137,4 +139,8 @@ func (c *Client) getName() (string, error) {
 	}
 
 	return hostname, nil
+}
+
+func isCancelledError(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "context canceled")
 }
--- a/agent/client/workflow_listen_for_ds_connection_tests.go
+++ b/agent/client/workflow_listen_for_ds_connection_tests.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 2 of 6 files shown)


agent/client/client.go
```diff
@@ -138,3 +140,7 @@ func (c *Client) getName() (string, error) {
 
 	return hostname, nil
 }
+
+func isCancelledError(err error) bool {
```
xoscar, line 144: I didn't find a good way of doing this, the rpc message contains a wrapped error that when I tried to unwrap it it was returning a plane `interface{}` object 🤷🏾

mathnogueira: Did you try: 

```
errors.Is(err, context.Canceled)
```

mathnogueira (approved):

agent/client/workflow_shutdown.go
```diff
@@ -21,12 +22,13 @@ func (c *Client) startShutdownListener(ctx context.Context) error {
...
-			if err == io.EOF {
+
+			if errors.Is(err, io.EOF) || isCancelledError(err) {
 				return
 			}
 
 			if err != nil {
-				log.Fatal("could not get message from trigger stream: %w", err)
+				log.Fatal("could not get shutdown listener: %w", err)
 			}
```
danielbdias, line 32: Does forcing a shutdown here in case of an error make sense? Something like using: `os.Exit(-1)`.

agent/client/workflow_listen_for_ds_connection_tests.go
```diff
@@ -21,12 +22,12 @@ func (c *Client) startDataStoreConnectionTestListener(ctx context.Context) error
...
 			req := proto.DataStoreConnectionTestRequest{}
 			err := stream.RecvMsg(&req)
-			if err == io.EOF {
+			if errors.Is(err, io.EOF) || isCancelledError(err) {
 				return
 			}
 
 			if err != nil {
-				log.Fatal("could not get message from trigger stream: %w", err)
+				log.Fatal("could not get message from ds connection stream: %w", err)
```
danielbdias, line 30:
```suggestion
				log.Fatal("could not get message from data store connection stream: %w", err)
```

danielbdias (approved):

agent/client/workflow_shutdown.go
xoscar: I believe `log.Fatal` is already doing that

```
// Fatal is equivalent to Print() followed by a call to os.Exit(1).
func Fatal(v ...any) {
	std.Output(2, fmt.Sprint(v...))
	os.Exit(1)
}
```

PR #3255, opened 2023-10-12
feat(cli): Token Support

xoscar:
This PR adds token support for the `tracetest start` command

## Changes

- Adds new env variable to support the TOKEN
- Adds config to replace jwt and token from the given flags

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video
https://www.loom.com/share/a1405279a47a429cae4fe0bb8df25233

### DIFF
--- a/cli/cmd/start_cmd.go
+++ b/cli/cmd/start_cmd.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"os"
 
 	agentConfig "github.com/kubeshop/tracetest/agent/config"
 	"github.com/kubeshop/tracetest/cli/config"
@@ -10,8 +11,9 @@ import (
 )
 
 var (
-	start      = starter.NewStarter(configurator, resources)
-	saveParams = &saveParameters{}
+	start        = starter.NewStarter(configurator, resources)
+	defaultToken = os.Getenv("TRACETEST_TOKEN")
+	saveParams   = &saveParameters{}
 )
 
 var startCmd = &cobra.Command{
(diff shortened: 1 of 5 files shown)


cli/cmd/start_cmd.go
```diff
@@ -49,6 +52,7 @@ func init() {
 	startCmd.Flags().StringVarP(&saveParams.organizationID, "organization", "", "", "organization id")
 	startCmd.Flags().StringVarP(&saveParams.environmentID, "environment", "", "", "environment id")
 	startCmd.Flags().StringVarP(&saveParams.agentApiKey, "api-key", "", "", "agent api key")
+	startCmd.Flags().StringVarP(&saveParams.tokenApiKey, "token", "", defaultToken, "token api key")
```
jorgeepc, line 55: `cli-api-key`instead of `token`?

mathnogueira: I'm kind of late to this discussion. But I think that `cli-api-key` is too close to `api-key` and it's hard to understand what's the difference between them. 

I would suggest using either `token` or a more meaningful name for this flag (which I have no suggestions yet). The idea behind using `token` is that the user will get the value from the `Token` tab, so will be more intuitive for the user to create a connection between the value and the flag they need to use.

mathnogueira (approved):
Left a comment on the flag name discussion. It's not a blocker

cli/cmd/start_cmd.go
xoscar: I agree with this statement too, I think we can go with token as it matches the SaaS cloud, then in the future we can decide to change it

PR #3279, opened 2023-10-18
chore(agent): make docker image multiarch

schoren:
This PR makes the agent docker image mutliarch. For simplicity, it uses the same approach as the main tracetest image. Meaning the building happens through goreleaser. Because of how goreleaser works, it was not practical to keep the agent image "inheriting" from the main tracetest image, instead the easiest way was to have a dockerfile that simply imports the already built cli binary and set an entrypoint.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.github/workflows/release-version.yml
+++ b/.github/workflows/release-version.yml
@@ -102,20 +102,6 @@ jobs:
           ANALYTICS_BE_KEY: ${{ secrets.ANALYTICS_BE_KEY }}
           TRACETEST_DEFAULT_CLOUD_ENDPOINT: ${{ secrets.TRACETEST_DEFAULT_CLOUD_ENDPOINT }}
 
-      # release agent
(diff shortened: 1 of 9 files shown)


.github/workflows/release-version.yml
```diff
@@ -102,20 +102,6 @@ jobs:
           ANALYTICS_BE_KEY: ${{ secrets.ANALYTICS_BE_KEY }}
           TRACETEST_DEFAULT_CLOUD_ENDPOINT: ${{ secrets.TRACETEST_DEFAULT_CLOUD_ENDPOINT }}
 
-      # release agent
```
schoren, line 105: this is now handled by goreleaser on the main release step

danielbdias (approved):

PR #3295, opened 2023-10-25
feat: use semver for version check

schoren:
This PR changes how the CLI checks for version. Instead of an exact mach, it does a semver comparision for [major].[minor], alowing any patch and prereleases to be considered valid.

### DIFF
--- a/cli/config/version.go
+++ b/cli/config/version.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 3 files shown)


cli/config/version.go
```diff
@@ -31,15 +33,48 @@ Server: Not Configured`, false
 Server: Failed to get the server version - %s`, err.Error()), false
 	}
 
+	versionMatch := false
 	version := meta.GetVersion()
-	isVersionMatch := version == Version
-	if isVersionMatch {
+	if !isSemver(version) || !isSemver(Version) {
```
schoren, line 38: this is the old behaviour. useful for CI and development, where versions are not semantic (like `dev` or some commit sha)

xoscar (approved):

PR #3326, opened 2023-11-01
feat(frontend): add trigger type modal

jorgeepc:
This PR adds the Trigger type selection modal that should be part of the new Test creation flow.

## Changes

- Trigger type modal

## Fixes

- fixes https://github.com/kubeshop/tracetest-cloud/issues/253

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Screenshot

Screenshot 2023-10-31 at 19 03 02

### DIFF
--- a/web/src/components/TriggerTypeModal/TriggerTypeModal.tsx
+++ b/web/src/components/TriggerTypeModal/TriggerTypeModal.tsx
@@ -1,0 +1,45 @@
+import {TriggerTypeToPlugin} from 'constants/Plugins.constants';
+import {useDashboard} from 'providers/Dashboard/Dashboard.provider';
+import CreateTestAnalyticsService from 'services/Analytics/CreateTestAnalytics.service';
+import TriggerTypeCard from './TriggerTypeCard';
+import * as S from './TriggerTypeModal.styled';
+
+const pluginList = Object.values(TriggerTypeToPlugin);
+
+interface IProps {
+  isOpen: boolean;
+  onClose(): void;
+}
+
+const TriggerTypeModal = ({isOpen, onClose}: IProps) => {
+  const {navigate} = useDashboard();
+
+  return (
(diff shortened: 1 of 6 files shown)


web/src/pages/Home/TestsList.tsx
```diff
@@ -98,7 +98,7 @@ const Tests = () => {
         </Pagination>
       </S.Wrapper>
 
-      <CreateTestModal isOpen={isCreateTestOpen} onClose={() => setIsCreateTestOpen(false)} />
```
xoscar, line 101: Do we plan to keep this component around?

web/src/components/TriggerTypeModal/TriggerTypeModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+      footer={null}
+      title={<S.Title level={2}>Create a new test</S.Title>}
+      visible={isOpen}
+      width={625}
+    >
+      <S.Title level={3} $marginBottom={16}>
+        What kind of trigger do you want to use to initiate this Tracetest?
+      </S.Title>
+
+      <S.CardList>
```
xoscar, line 29: I don't see the coming soon cards, is this what we want?

web/src/pages/Home/TestsList.tsx
jorgeepc: It's going to die. We'll do a big cleanup after we have everything in place

web/src/components/TriggerTypeModal/TriggerTypeModal.tsx
jorgeepc: We only have the `OpenAPI` as coming soon and I think that is going to be an import, not a create flow. Anyway we can add it if we want

PR #3342, opened 2023-11-08
feat: make agent reconnect in case of connection dropping

mathnogueira:
This PR makes the agent reconnect to the server in case of a grpc failure (server disconnects or any other error). If not possible to reconnect, the agent will exit.

## Checklist

- [x] tested locally
- [x] added new dependencies
- [ ] updated the docs
- [x] added a test


agent/client/workflow_ping.go
```diff
@@ -9,11 +9,15 @@ import (
 
 func (c *Client) startHearthBeat(ctx context.Context) error {
 	client := proto.NewOrchestratorClient(c.conn)
-	ticker := time.NewTicker(2 * time.Minute)
+	ticker := time.NewTicker(c.config.PingPeriod)
```
mathnogueira, line 12: Changed the ping to `30s`

agent/client/client.go
```diff
@@ -144,3 +154,32 @@ func (c *Client) getName() (string, error) {
 func isCancelledError(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "context canceled")
 }
+
+func (c *Client) reconnect() error {
+	// connection is not working. We need to reconnect
+	err := retry.Do(func() error {
+		return c.connect(context.Background())
+	}, retry.Attempts(3), retry.Delay(1*time.Second))
```
schoren, line 162: this will do 3 attempts with 1sec between them, right? Is that correct? Maybe we want to make it longer? like at least 1 min total?

agent/client/workflow_listen_for_ds_connection_tests.go
```diff
@@ -26,8 +27,22 @@ func (c *Client) startDataStoreConnectionTestListener(ctx context.Context) error
 				return
 			}
 
+			if err != nil && isConnectionError(err) {
+				err = retry.Do(func() error {
+					return c.reconnect()
+				})
+				if err == nil {
+					// everything was reconnect, so we can exist this goroutine
```
schoren, line 35: what does this comment mean? I don't understand it :S

agent/client/client.go
danielbdias, line 162: Can we add these numbers as constants in the code? Something like:

```go
// somewhere
const reconnectRetryAttempts = 3
const reconnectRetryAttemptDelay = 1 * time.Second

// here 
	err := retry.Do(func() error {
		return c.connect(context.Background())
	}, retry.Attempts(reconnectRetryAttempts), retry.Delay(reconnectRetryAttemptDelay))
```

agent/client/connector.go
```diff
@@ -13,16 +13,28 @@ import (
 )
 
 func Connect(ctx context.Context, endpoint string, opts ...Option) (*Client, error) {
-	conn, err := connect(ctx, endpoint)
-	if err != nil {
-		return nil, err
+	config := Config{
+		PingPeriod: 30 * time.Second,
```
danielbdias, line 17: Same thing here. Can we add this config as a constant?

danielbdias (approved):

agent/client/workflow_listen_for_trigger_requests.go
```diff
@@ -22,12 +21,26 @@ func (c *Client) startTriggerListener(ctx context.Context) error {
...
+				})
+				if err == nil {
+					// everything was reconnect, so we can exist this goroutine
+					// as there's another one running in parallel
+					return
+				}
+
+				log.Fatal(err)
+			}
+
```
schoren, line 40: all this functiosn are copy/pasted. can't they be generalizaed? like:

```
func (c Client) handleDisconnectionError(inputErr error) error {
    if inputErr == nil {
        return  nil
    }
	  if isConnectionError(inputErr) {
        return retry.Do(func() error {
					return c.reconnect()
				})
		}
    return inputErr
}
```

and the 
```suggestion
      err = c.handleConnectionError(err)
			if err != nil {
				log.Fatal(err)
			}
```

agent/client/client.go
mathnogueira: Thats delay doubles every attempt. So that would be 7 seconds total. If we set attempts as 6 we would get a bit longe than 1 minute

agent/client/workflow_listen_for_ds_connection_tests.go
mathnogueira: When we call `reconnect`. It needs to restart the client. Which calls this function again with amother goroutine

agent/client/workflow_listen_for_trigger_requests.go
mathnogueira: I used the core idea of your suggestion.

I didn't follow it 100% for two reasons:

1. I cannot exit the process if any error happens, only if it's a connection error and we can't connect.
2. I still need to exit the goroutine because when `c.reconnect()` is called, a new goroutine will spawn with a new `stream` object. When reconnecting to a grpc server, we also need to get a new `stream` using the new connection. So I have to have a way of ensuring that the old goroutine will not run again. That's why we have this:

```go
if err == nil {
  return
}
```

It means that the reconnection was sucessfull and now we have a new `grpc.Conn` and new `grpc streams` as well.

schoren (approved):

PR #3378, opened 2023-11-20
chore: Migrate start actions from CLI to agent

danielbdias:
This PR starts to decouple the agent from the CLI. Here, we are removing the old `cmd` command for the agent (because this responsibility should be to the CLI), and the `starter` and `config flags` are being moved to the agent because it is related to the agent business logic.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


agent/ui/pterm.go
```diff
@@ -0,0 +1,192 @@
+package ui
```
jorgeepc, line 1: Question, why not reusing what we have in `cli/ui`?

jorgeepc (approved):
Great job @danielbdias

agent/ui/pterm.go
danielbdias: Good question! I'm not using `cli/ui` to avoid a dependency of `cli -> agent`.  I'll unify them in the next PR, moving to a shared package.

PR #3429, opened 2023-12-06
fix(server): add context event properties

jorgeepc:
This PR adds event properties in the context so it can be accessed from workers.

## Changes

-

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/app/app.go
+++ b/server/app/app.go
@@ -7,7 +7,6 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"regexp"
 
 	"github.com/gorilla/handlers"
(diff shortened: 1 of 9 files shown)


server/app/app.go
```diff
@@ -372,42 +371,6 @@ func (app *App) Start(opts ...appOption) error {
 	return nil
 }
 
-var (
```
xoscar, line 375: good cleanup!

xoscar (approved):

PR #3481, opened 2023-12-27
fix: skip open attribute list step on e2e

danielbdias:
This PR fixes our Cypress e2e tests that are breaking due to a behavior change on our UI. Since we already open the Attribute list on the Trace and Test tabs, there is no need of having of clicking on the expand button to show the list.

## Fixes

- Update e2e to skip open attribute list step

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -163,7 +163,7 @@ jobs:
 
       - name: Setup node
         # if cache miss
-        if: steps.cache-web-build.outputs.cache-hit != 'true'
+        # if: steps.cache-web-build.outputs.cache-hit != 'true'
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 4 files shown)


schoren (approved):

.github/workflows/pull-request.yaml
```diff
@@ -146,80 +146,45 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: set cache keys
```
danielbdias, line 149: Our cache is causing false positives and negatives during our tests, causing our tests not to capture changes that should break tests.

This allowed us to merge a PR with a false negative test (the PR that changed the behavior of the Trace tab) and blocked this PR with a false positive.

Issue #3507, opened 2024-01-09
[bug] Polling for traces from Grafana Cloud Tempo does not work


adnanrahic:
**Describe the bug**
Polling for traces from Grafana Cloud Tempo does not work

**To Reproduce**
1. Go to 'settings > tracing backends'
2. Click on 'tempo'
3. Add Grafana Cloud Tempo creds
4. Connection test fails and test run fails to fetch traces

**Expected behavior**

**Screenshots**

**Browser (please complete the following information):**
 - Browser: [Firefox]
 - Version [e.g. 99]

**Additional context**

xoscar:
@adnanrahic I was able to get it working with this setup
![Screen Shot 2024-01-09 at 5 00 32 p m](https://github.com/kubeshop/tracetest/assets/11051031/1ecf7337-a201-4833-ba04-e46501ae2839)

PR #3515, opened 2024-01-10
fix(backend): tempo cloud connection string

xoscar:
This PR fixes a confusing way of dealing with the tempo cloud http connection, as pasting the given endpoint from the webapp uses `:443` instead of `https`

## Changes

- Defaults `:443` port to `https`

## Fixes

- https://github.com/kubeshop/tracetest/issues/3507

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


server/tracedb/datasource/http.go
```diff
@@ -27,6 +27,10 @@ type HttpClient struct {
 
 func NewHttpClient(name string, config *datastore.HttpClientConfig, callback HttpCallback) DataSource {
 	endpoint, _ := urlx.Parse(config.Url)
+	if endpoint.Port() == "443" {
+		endpoint.Scheme = "https"
+	}
```
xoscar, line 32: With a url like this:

`tempo-us-central1.grafana.net:443`

The parsing method returns

`http://tempo-us-central1.grafana.net:443`

Which breaks the `getTraceById` logic.

server/tracedb/tempodb.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func tempoDefaultPorts() []string {
-	return []string{"9095", ""}
+	return []string{"9095", "443", "1"}
```
xoscar, line 26: 443 and 1 (using https://tempo-us-central1.grafana.net) should be also valid ports

jorgeepc (approved):

server/tracedb/tempodb.go
schoren: I've never seen port 1 being used. Is there any docs about this port?

xoscar: For some reason the url parser uses `1` as port when using something like `https://tempo-us-central1.grafana.net/`

xoscar: Screen Shot 2024-01-10 at 1 57 31 p m

schoren: That looks like a bug in the URL parsing library. Do you know why we're using an URL parser other than the go's stdlib one?

xoscar: Yeah the standard makes it worse in some cases, here's a table
Screen Shot 2024-01-10 at 2 00 28 p m

schoren: I see. That makes sense. What I would do is handle the case of port "1" and replace it by a correct default port, for example:

```
if port == 1 {
  if schema == "http" {
    port = 80
  }
  if schema == "https" {
    port = 43
  }
}
```

adnanrahic:
This is awesome. When will this be ready for me to try it?

PR #3534, opened 2024-01-16
fix: make test connection async on trace polling step

danielbdias:
This PR changes the Test Connection action on the Trace Polling start to be async.

## Fixes

- Update Test Connection action on the Trace Polling start to be async.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/server/executor/tracepollerworker/starter_worker.go
+++ b/server/executor/tracepollerworker/starter_worker.go
@@ -5,20 +5,31 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"sync"
 
 	"github.com/kubeshop/tracetest/server/datastore"
 	"github.com/kubeshop/tracetest/server/executor"
+	"github.com/kubeshop/tracetest/server/model"
 	"github.com/kubeshop/tracetest/server/model/events"
 	"github.com/kubeshop/tracetest/server/pkg/pipeline"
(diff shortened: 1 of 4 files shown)


server/executor/tracepollerworker/starter_worker.go
```diff
@@ -85,29 +100,52 @@ func (w *tracePollerStarterWorker) ProcessItem(ctx context.Context, job executor
...
+	})
+
 	if err != nil {
-		return fmt.Errorf("could not get current datastore: %w", err)
+		return nil, err
 	}
 
-	emitEvent(ctx, w.state, events.TracePollingStart(job.Test.ID, job.Run.ID, string(ds.Type), endpoints))
+	w.dsTestPipeline.Run(ctx, job)
+	wg.Add(1)
```
mathnogueira, line 146:
```suggestion
   	wg.Add(1)
	w.dsTestPipeline.Run(ctx, job)
```

This is important because if you run `wg.Done()` before `wg.Add()` is executed, the app will panic. I know this should not happen in normal cases, but we never know. Better be safe than sorry.

https://go.dev/play/p/NHIkZ00KhGx

mathnogueira (approved):
Only one comment

PR #3544, opened 2024-01-18
feat: agent collector statistics report

mathnogueira:
This PR makes the agent report some statistics to the server when requested.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test


agent/client/workflow_listen_for_otlp_connection_tests.go
```diff
@@ -0,0 +1,47 @@
...
+				return
+			}
+
+			if err != nil {
+				log.Println("could not get message from otlp connection stream: %w", err)
+				time.Sleep(1 * time.Second)
+				continue
+			}
+
+			// TODO: Get ctx from request
```
danielbdias, line 39: Philosophical question: is there a way to get the context from the request?
My impression is that there is no way to get these contexts due to the passive nature of getting the messages.

Could we create a context in the future and change it by monitoring the stream connection?

danielbdias (approved):

agent/client/workflow_send_ds_connection_result.go
```diff
@@ -7,14 +7,14 @@ import (
 	"github.com/kubeshop/tracetest/agent/proto"
 )
 
-func (c *Client) SendDataStoreConnectionResult(ctx context.Context, response *proto.DataStoreConnectionTestResponse) error {
+func (c *Client) SendOTLPConnectionResult(ctx context.Context, response *proto.OTLPConnectionTestResponse) error {
```
xoscar, line 10: Is the renaming of this function correct? seems that there is a new file about oltp

xoscar (approved):
GJ!

agent/client/workflow_send_ds_connection_result.go
mathnogueira: Not correct, nice catch. Unfortunately this endpoint isn't covered by a test, that's why I didn't get this error :(

agent/client/workflow_listen_for_otlp_connection_tests.go
mathnogueira: We would need to propagate headers through the messages to be able to create a context from the request (propagate the otel headers)

agent/client/workflow_send_ds_connection_result.go
mathnogueira: fixed the mess and also added tests to ensure it will not break in the future

PR #3552, opened 2024-01-22
feat: handle stop run and skip polling during polling step

schoren:
This PR fixes the polling step in the agent and core server so it can correctly handle skip polling and stop requests.

## Loom video

https://www.loom.com/share/2e6c88596633449f87928dae87b89250


Makefile
```diff
@@ -6,12 +6,12 @@ PROJECT_ROOT=${PWD}
...
-	@printf "\033[0;31m Tracetest requires goreleaser pro installed (licence not necessary for local builds)\033[0m\n\n"
-	@printf "\033[0;33m See https://goreleaser.com/install/ \033[0m\n\n"
-	@exit 1
-endif
+# ifneq "$(CURRENT_GORELEASER_VERSION)" "$(GORELEASER_VERSION)"
+# 	@printf "\033[0;31m Bad goreleaser version $(CURRENT_GORELEASER_VERSION), please install $(GORELEASER_VERSION)\033[0m\n\n"
+# 	@printf "\033[0;31m Tracetest requires goreleaser pro installed (licence not necessary for local builds)\033[0m\n\n"
+# 	@printf "\033[0;33m See https://goreleaser.com/install/ \033[0m\n\n"
+# 	@exit 1
+# endif
```
mathnogueira: this was the worst check we had :smile:

xoscar (approved):
@schoren thanks for the video it looks great, I didn't see something concerning or out of place while reviewing, so I'm approving!

server/executor/test_pipeline.go
```diff
@@ -159,14 +162,21 @@ type runCancelHandlerFn func(ctx context.Context, run test.Run) error
...
+	return (run.TriggerResult.Error != nil &&
+		ErrorMessageIsUserCancelled(run.TriggerResult.Error.ErrorMessage)) ||
+		(run.LastError != nil && ErrorMessageIsUserCancelled(run.LastError.Error()))
 }
 
 func ErrorMessageIsUserCancelled(msg string) bool {
 	return msg == ErrUserCancelled.Error()
 }
 
+func ErrorMessageIsSkipTraceColleciton(msg string) bool {
```
mathnogueira, line 176: I think you wanted `ErrorMessageIsSkipTraceCollection` (there's a small typo)

server/executor/test_pipeline.go
```diff
@@ -159,14 +162,21 @@ type runCancelHandlerFn func(ctx context.Context, run test.Run) error
 var ErrUserCancelled = fmt.Errorf("cancelled by user")
 
 func RunWasUserCancelled(run test.Run) bool {
-	return run.TriggerResult.Error != nil &&
-		ErrorMessageIsUserCancelled(run.TriggerResult.Error.ErrorMessage)
+	// depeending on when the Run was cancelled (which step was being executed)
+	// the error might be set on different filelds
```
mathnogueira, line 166: `filelds` == `fields`?

server/http/controller.go
```diff
@@ -773,8 +773,10 @@ func (c *controller) UpdateTestRun(ctx context.Context, testID string, runID int
 	existingRun.TriggerResult = run.TriggerResult
 	existingRun.Trace = traces.MergeTraces(existingRun.Trace, run.Trace)
 	existingRun.State = run.State
+	existingRun.LastError = errors.New(testRun.LastErrorState)
 
 	if executor.RunWasUserCancelled(existingRun) {
+		existingRun.LastError = nil
```
mathnogueira, line 779: So even if something went wrong, we gonna ignore it if the process was canceled, right?

mathnogueira (approved):
Looks good, just a few minor typos and questions

server/http/controller.go
schoren: If there was an error, it would happen before the cancellation, so the test would be already stopped and the cancelation would be ignored before this

PR #3600, opened 2024-02-06
chore(frontend): indentifying performance improvements

xoscar:
This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video


web/src/models/DAG.model.ts
```diff
@@ -13,6 +13,7 @@ function getNodesDatumFromSpans(spans: Span[], type: NodeTypesEnum): INodeDatum<
 }
 
 function DAG(spans: Span[], type: NodeTypesEnum) {
+  // TODO: this runs twice for the list of spans
```
xoscar, line 16: We might be able to use the tree data structure of the trace here, instead of sorting, by running a recursive `flatMap` instead

PR #3611, opened 2024-02-08
feat(server): add endpoint to search spans in a trace

schoren:
This PR adds an endpoint that allows searching spans within a trace. 

The query is transmitted via post body to avoid possible issues  with URL length if the query is a bit large.

The plain text implementation is a bit naive, it just does a `string.contains` over a json encoded version of each span, both the query and the json lowercased for improved matching. 
While this is a pretty terrible implementation, it's easy and it works. In the future if needed we could implement a more complex solution, like [Bleve](https://github.com/blevesearch/bleve), ElasticSearch, etc, but at this time this seems like a good enough approach.

### Search by query language

```
POST http://localhost:11633/api/tests/{{testID}}/run/{{runID}}/search-spans

{"query": "span[tracetest.span.name = \"Delete Report from cache\"]"}
```

### Search by text

```
POST http://localhost:11633/api/tests/{{testID}}/run/{{runID}}/search-spans

{"query": "Delete Report from cache"}
```

## Loom video

https://www.loom.com/share/e2912a4fde554a85a8dfb043fca983fc


xoscar (approved):
I understand the idea of having this done in a simpler way for the moment, great job on the quick turn around @schoren!

PR #3618, opened 2024-02-08
chore(examples): Adding Cypress and Playwright examples

xoscar:
This PR adds examples for both Cypress and Playwright, so we have them as part of the main repo not only on the pokeshop repo.

## Changes

- Adds Cypress example
- Adds Playwright example

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


adnanrahic (commented):
Very nice!!!

I see in the readme you say it's using OpenTelemetry Collector but there is a Jaeger URL in the `collector.config.yaml`. Is this what you want or did you forget removing something?

PR #3643, opened 2024-02-15
feat(cli): reauthenticate user in case of invalid token

schoren:
This PR improves the error handling when the server does not accept the JWT token for whatever reason (corrupted config file, expired token, anything that makes the server return a 401 code).
When an auth error happens, the authentication flow begins, and the issued command is automatically rerun with the new credentials.

For the case of the `start` cmd, this new behaviour was causing the cli to show 2 times the org/env selector: once for the new auth and another one for the `start` command (it shows the selector when started unless org/env are passed via flags).
To overcome this issue, I added some context magic to pass the values around in a relatively seamless way so this command can pick the values and not ask the user 2 times.

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

https://www.loom.com/share/cf956f7d668f475baa22678d4ac6a7bf

### DIFF
--- a/cli/cmd/middleware.go
+++ b/cli/cmd/middleware.go
@@ -17,27 +17,9 @@ type RunFn func(ctx context.Context, cmd *cobra.Command, args []string) (string,
 type CobraRunFn func(cmd *cobra.Command, args []string)
 type MiddlewareWrapper func(RunFn) RunFn
 
-// type keyCmd struct{}
-
-// var cmdKey keyCmd
-
-// func ContextWithCmd(cmd *cobra.Command, args []string) context.Context {
-// 	return context.WithValue(
-// 		cmd.Context(),
-// 		cmdKey,
-// 		buildCmdLine(cmd, args),
-// 	)
-// }
-
-// func ContextGetCmd(ctx context.Context) string {
-// 	v := ctx.Value(cmdKey)
-// 	if v == nil {
-// 		return ""
-// 	}
-// 	return v.(string)
-// }
-
 func rootCtx(cmd *cobra.Command) context.Context {
+	// cobra does not correctly progpagate rootcmd context to sub commands,
+	// so we need to manually traverse the command tree to find the root context
 	if cmd == nil {
 		return nil
 	}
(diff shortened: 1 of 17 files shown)


cli/cmd/start_cmd.go
```diff
@@ -37,6 +37,17 @@ var startCmd = &cobra.Command{
...
+		// override organization and environment id from context.
+		// this happens when auto rerunning the cmd after relogin
+		if orgID := config.ContextGetOrganizationID(ctx); orgID != "" {
+			flags.OrganizationID = orgID
+		}
+		if envID := config.ContextGetEnvironmentID(ctx); envID != "" {
+			flags.EnvironmentID = envID
+		}
+
+		spew.Dump(flags)
```
xoscar, line 49: spew 🔫

cli/cmd/middleware.go
```diff
@@ -29,6 +59,28 @@ func WithResultHandler(runFn RunFn) CobraRunFn {
...
+	reqErr := resourcemanager.RequestError{}
+	if errors.As(err, &reqErr) && reqErr.IsAuthError {
+		handleAuthError(ctx)
+	} else {
+		OnError(err)
+	}
+}
+
+func handleAuthError(ctx context.Context) {
+	ui.DefaultUI.Warning("Your authentication token has expired, please log in again.")
```
xoscar, line 72: Does this handles environment tokens expiration?

xoscar (approved):
The code looks good, I'd suggest @kdhamric taking a look at the video to see if it matches what he wants

jorgeepc (approved):

cli/cmd/middleware.go
schoren: That's a good question. I didn't test that case, but this mechanism is triggered whenever the server returns a 401. So if an expired token throws a 401, then this will be triggered.

kdhamric:
I watched the video - looks awesome @schoren !

PR #3645, opened 2024-02-15
feat: propagete trace context in queues

mathnogueira:
This PR makes the context be propagated to all queues

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

### DIFF
--- a/server/pkg/pipeline/queue.go
+++ b/server/pkg/pipeline/queue.go
@@ -20,7 +20,7 @@ type Listener[T any] interface {
 }
 
 type QueueDriver[T any] interface {
-	Enqueue(T)
+	Enqueue(context.Context, T)
 	SetListener(Listener[T])
 }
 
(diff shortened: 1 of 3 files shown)


server/pkg/pipeline/queue.go
```diff
@@ -78,6 +80,12 @@ func (q Queue[T]) Enqueue(ctx context.Context, item T) {
 		return
 	}
 
+	workerCtx := context.Background()
+	if propagator := otel.GetTextMapPropagator(); propagator != nil {
+		var carrier propagation.HeaderCarrier
+		workerCtx = propagator.Extract(workerCtx, carrier)
+	}
```
mathnogueira, line 87: I need a new context here because `ctx` will be canceled right after this function returns.

schoren (approved):

PR #3656, opened 2024-02-16
fix(cli): fix merge issues

schoren:
This PR fixes some issues caused by an incorrect merge

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

### DIFF
--- a/cli/cmd/start_cmd.go
+++ b/cli/cmd/start_cmd.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 
+	"github.com/davecgh/go-spew/spew"
 	agentConfig "github.com/kubeshop/tracetest/agent/config"
 	"github.com/kubeshop/tracetest/agent/runner"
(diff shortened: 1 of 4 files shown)


mathnogueira (approved):

cli/cmd/start_cmd.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 
+	"github.com/davecgh/go-spew/spew"
```
danielbdias, line 7: Should we keep `spew`? I noticed that we tried to remove it on [this commit](https://github.com/kubeshop/tracetest/commit/7d6bf5f6fb0424bb4ca23c7aa496416e45259282).

schoren: ah damn, good catch. fixed on main

PR #3657, opened 2024-02-16
fix(frontend): fix bug in test specs virtual list

jorgeepc:
This PR fixes a bug with the implementation of the virtual list for `Test Specs`. A test spec item has a variable height, and for that reason we need to implement the `VariableSizeList` component with a strategy to provide the item height as part of the render cycle.

## Changes

- fix test specs list

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Before
- Double scroll and 🐛


web/src/components/TestSpec/TestSpec.styled.ts
```diff
@@ -57,7 +57,10 @@ export const Selector = styled.div`
 
 export const Title = styled(Typography.Text)`
   color: ${({theme}) => theme.color.text};
+  font-family: monospace;
```
xoscar, line 60: Why do we need this font-family change here?

web/src/components/TestSpecs/TestSpecs.tsx
```diff
@@ -13,37 +17,67 @@ interface IProps {
 }
 
 const TestSpecs = ({assertionResults, onDelete, onEdit, onOpen, onRevert}: IProps) => {
+  const hiddenElementRef = useRef<HTMLDivElement>(null);
+  const specs = useAppSelector(state => TestSpecsSelectors.selectSpecs(state));
+
+  const getItemSize = useCallback(
+    index => {
+      const item = assertionResults?.resultList?.[index];
```
xoscar, line 25: I believe we should add some comments about what is going here, why do we need this etc

xoscar (approved):
Hey @jorgeepc GJ I added some non-blocking comments!

PR #3659, opened 2024-02-16
fix(cli): configure for non interactive environments

schoren:
This PR fixes an issue on the CLI preventing the `configure` command to work on automated environments. The command was asking for user input (asking for server) even when the `--token` flag was passed. This flag allows us to infer a non interactive environment so we can use a reasonable fallback (cloud url or previously configured value from config file) if no `--server-url` is provided.

On top of this, there was some incorrect handling of the flags within the configurator. Flags were saved as an instance variable by value instead of reference, so when one method modifies a flag value, the change is lost at the end of the function scope (see comments on PR)

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

### DIFF
--- a/cli/config/configurator.go
+++ b/cli/config/configurator.go
@@ -21,7 +21,7 @@ type Configurator struct {
 	ui             cliUI.UI
 	onFinish       onFinishFn
 	errorHandlerFn errorHandlerFn
-	flags          agentConfig.Flags
+	flags          *agentConfig.Flags
 	finalServerURL string
 }
 
@@ -38,7 +38,7 @@ func NewConfigurator(resources *resourcemanager.Registry) Configurator {
 		errorHandlerFn: func(ctx context.Context, err error) {
 			ui.Exit(err.Error())
 		},
-		flags: agentConfig.Flags{},
+		flags: &agentConfig.Flags{},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cli/config/configurator.go
```diff
@@ -55,11 +55,19 @@ func (c Configurator) WithErrorHandler(fn errorHandlerFn) Configurator {
 }
 
 func (c Configurator) Start(ctx context.Context, prev *Config, flags agentConfig.Flags) error {
-	c.flags = flags
-	serverURL, err := c.getServerURL(prev, flags)
-	c.finalServerURL = serverURL
-	if err != nil {
-		return err
+	c.flags = &flags
```
schoren, line 58: the original `flags` value was passed around in functions call, on top of being saved to the instance. this was confusing to follow and created 2 different copies of the same value, so modifications were almost impossible to track

danielbdias (approved):

cli/config/selector.go
```diff
@@ -17,12 +17,12 @@ type Entry struct {
 func (c Configurator) organizationSelector(ctx context.Context, cfg Config, prev *Config) (string, error) {
 	resource, err := c.resources.Get("organization")
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("cannot create organizations client: %w", err)
```
schoren, line 20: some extra error info helped me debug the issue

PR #3662, opened 2024-02-19
feat(cli): add open telemetry support for Tracetest agent

danielbdias:
This PR adds support for OTel Collector on Tracetest Agent.

## Changes

- Added capability of setting verbose mode via env vars
- Added option to inform the address of an OTel Collector to send observability telemetry

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/cli/config/configurator.go
+++ b/cli/config/configurator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 
 	"github.com/golang-jwt/jwt"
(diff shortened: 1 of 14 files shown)


cli/config/configurator.go
```diff
@@ -162,6 +162,19 @@ func (e invalidServerErr) Render() {
 }
 
 func (c Configurator) populateConfigWithVersionInfo(ctx context.Context, cfg Config) (_ Config, _ error, isOSS bool) {
+	cliVersion := Version
+	if cliVersion == "dev" {
```
danielbdias, line 166: Helper code to help us to run the agent locally when the cliVersion is `dev`

schoren (approved):

cli/config/configurator.go
xoscar: Can we point the agent to a different environment from a dev build too? I want to make sure this will allows us to be flexible enough when debugging

danielbdias: Yes! Added an update to allow these config to be configured by env vars.

PR #3671, opened 2024-02-21
fix(cli): automatically use cloud endpoint if token is passed

schoren:
This PR makes the CLI not ask interactively from the user the desired server URL, and instead automatically select one. 

When passing `--token` or `--api-key` we can no longer asume an interactive environment. Those tokens are usually used in a CI or otherwise automated environment, where user interaction is not available. 

In this case, server url is chose as:
1. if `--server-url` flag is passed, used that.
2. if we have a pre existing config, use that value.
3. otherwise use cloud endpoint.

This behaviour was already implemented for the `--token` flag, but not for the `--api-key` flag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ ifneq "$(CURRENT_GORELEASER_VERSION)" "$(GORELEASER_VERSION)"
 	@printf "\033[0;31m Bad goreleaser version $(CURRENT_GORELEASER_VERSION), please install $(GORELEASER_VERSION)\033[0m\n\n"
 	@printf "\033[0;31m Tracetest requires goreleaser pro installed (licence not necessary for local builds)\033[0m\n\n"
 	@printf "\033[0;33m See https://goreleaser.com/install/ \033[0m\n\n"
-	@exit 1
 endif
 
 
--- a/cli/config/configurator.go
+++ b/cli/config/configurator.go
@@ -59,10 +59,9 @@ func (c Configurator) Start(ctx context.Context, prev *Config, flags agentConfig
 	c.flags = &flags
 	var serverURL string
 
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -11,7 +11,6 @@ ifneq "$(CURRENT_GORELEASER_VERSION)" "$(GORELEASER_VERSION)"
 	@printf "\033[0;31m Bad goreleaser version $(CURRENT_GORELEASER_VERSION), please install $(GORELEASER_VERSION)\033[0m\n\n"
 	@printf "\033[0;31m Tracetest requires goreleaser pro installed (licence not necessary for local builds)\033[0m\n\n"
 	@printf "\033[0;33m See https://goreleaser.com/install/ \033[0m\n\n"
-	@exit 1
```
schoren, line 14: i want to know if we're outdated, but let's keep the builds rolling.

cli/config/configurator.go
```diff
@@ -187,8 +190,8 @@ func (c Configurator) populateConfigWithDevConfig(ctx context.Context, cfg *Conf
 }
 
 func (c Configurator) populateConfigWithVersionInfo(ctx context.Context, cfg Config) (_ Config, _ error, isOSS bool) {
-	cliVersion := Version
-	if cliVersion == "dev" {
+	useDevVersion := os.Getenv("TRACETEST_AGENT_DEV_CONFIG") == "true"
```
schoren, line 193: it took me a good 20 mins of debuggin to figure out this was happening. So I changed the behavior to be opt in by passing an env var. Otherwise it might hide issues with our config logic during development

jorgeepc (approved):

xoscar (approved):

PR #3675, opened 2024-02-22
feat(agent): add debug logs

schoren:
This PR adds debug logs so it's easier to identify issues, mainly when talking with the control plane.

It also fixes the logger configuration to correctly display debug logs when passing the `-v` flag

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

### DIFF
--- a/agent/client/workflow_listen_for_ds_connection_tests.go
+++ b/agent/client/workflow_listen_for_ds_connection_tests.go
@@ -8,13 +8,18 @@ import (
 
 	"github.com/kubeshop/tracetest/agent/proto"
 	"github.com/kubeshop/tracetest/server/telemetry"
+	"go.uber.org/zap"
 )
 
 func (c *Client) startDataStoreConnectionTestListener(ctx context.Context) error {
+	logger := c.logger.Named("dataStoreConnectionTestListener")
+	logger.Debug("Starting")
+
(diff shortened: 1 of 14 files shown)


agent/client/workflow_listen_for_ds_connection_tests.go
```diff
@@ -8,42 +8,55 @@ import (
...
+				logger.Error("could not get message from data store connection stream", zap.Error(err))
+			}
 			if isEndOfFileError(err) || isCancelledError(err) {
+				logger.Debug("data store connection stream closed")
 				return
 			}
 
 			reconnected, err := c.handleDisconnectionError(err)
 			if reconnected {
+				logger.Debug("reconnected to data store connection stream")
```
mathnogueira, line 40: I think this one should be a warning instead of debug, especially because I want to know that the error was handled.

mathnogueira (changes requested):
One comment that is valid for all workers that reconnect

agent/client/workflow_listen_for_ds_connection_tests.go
schoren: fixed

danielbdias (approved):

jorgeepc (approved):

PR #3689, opened 2024-02-26
feat(frontend): add getPublicInfo for tracing backend

jorgeepc:
This PR adds the getPublicInfo for the tracing backend service.

## Changes

- new getPublicInfo method

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


xoscar (approved):
Looks good, the filtering might be a little bit weird as the function says getPublicInfo but we are pretty much getting everything from the BE, we can decide later on if there is something we can do on that end

PR #3826, opened 2024-04-24
fix: don't use buffer and run things async

mathnogueira:
This PR updates the agent collector. Now it's not a `proxy` for open telemetry anymore. It only collects traces for tests and stores them. It also discards the test traces once they are sent back to the server to reduce the memory footprint.

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

### DIFF
--- a/agent/collector/ingester.go
+++ b/agent/collector/ingester.go
@@ -108,7 +108,7 @@ func (i *forwardIngester) SetSensor(sensor sensors.Sensor) {
 }
 
 func (i *forwardIngester) Ingest(ctx context.Context, request *pb.ExportTraceServiceRequest, requestType otlp.RequestType) (*pb.ExportTraceServiceResponse, error) {
-	go i.ingestSpans(ctx, request)
+	go i.ingestSpans(request)
 
 	return &pb.ExportTraceServiceResponse{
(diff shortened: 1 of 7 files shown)


agent/collector/ingester.go
```diff
@@ -45,14 +45,14 @@ func newForwardIngester(ctx context.Context, batchTimeout time.Duration, cfg rem
...
+	// if startRemoteServer {
+	// 	err := ingester.connectToRemoteServer(ctx)
+	// 	if err != nil {
+	// 		return nil, fmt.Errorf("could not connect to remote server: %w", err)
+	// 	}
 
-		go ingester.startBatchWorker()
-	}
+	// 	go ingester.startBatchWorker()
+	// }
```
schoren, line 55: can this be removed?

schoren (approved):
LGTM!

agent/collector/ingester.go
mathnogueira: I just cleaned the whole thing and removed all unused methods as well.

danielbdias (approved):

PR #3834, opened 2024-05-01
docs: Add Run Groups, Runs and Tests to Concepts Documentation

jfermi:
Also reduce the size of an image that was affecting SEO.

This PR...

## Changes

-

## Fixes

-

## Checklist

- [ ] tested locally
- [ ] added new dependencies
- [x] updated the docs
- [ ] added a test

## Loom video


docs/docs/concepts/run-groups.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+Now you also can specify the run group id when running a test. This would make the test part of the run group.
+
+```
+tracetest run test -f test1.yaml --group my-pipeline
+tracetest run test -f test2.yaml --group my-pipeline
+```
+
+With the `--group` argument, both tests would be run in parallel but would be seen inside the same run group.
```
kdhamric, line 62: I asked the team about this in slack. I would expect you still need the --skip-result-wait flag in order for the second test to process before tthe first test is totally finished, but perhaps setting the -group flag implicitly means 'and dont wait' (which might make sense).

kdhamric (approved):

PR #3856, opened 2024-05-16
fix(cli): add valid environmenttokens and invites resource name plural

jorgeepc:
This PR fixes a bug with the plural resource name for `environmenttokens` and `invites`.

## Changes

- fix plural resource names

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/cmd/resources.go
+++ b/cli/cmd/resources.go
@@ -57,12 +57,13 @@ var (
 
 	envTokensClient = resourcemanager.NewClient(
 		httpClient, cliLogger,
-		"token", "tokens",
+		"token", "environmenttokens",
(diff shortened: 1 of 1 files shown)


cli/cmd/resources.go
```diff
@@ -57,12 +57,13 @@ var (
 
 	envTokensClient = resourcemanager.NewClient(
 		httpClient, cliLogger,
-		"token", "tokens",
+		"token", "environmenttokens",
```
schoren, line 60: shouldn't singular and plural be the same? like token/tokens or environmenttoken/environmenttokens?

schoren (approved):

PR #3859, opened 2024-05-20
feat(cli): enabling environment provisioning

xoscar:
This PR adds the ability for users to provision environments from a definition file

## Changes

- Adds post process for environments
- Adds logic to apply multiple resources at once

## Fixes

-

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

https://www.loom.com/share/e22fb323f70e40f28d47ff3525ed3ce8?focus_title=1&muted=1&from_recorder=1


api/environments.yaml
```diff
@@ -0,0 +1,134 @@
+openapi: 3.0.1
```
xoscar, line 1: Had to add the stuff from cloud here

cli/cmd/resources.go
```diff
@@ -39,7 +39,7 @@ var (
 var (
 	httpClient = &resourcemanager.HTTPClient{}
 
-	variableSetPreprocessor = preprocessor.VariableSet(cliLogger)
+	variableSetPreprocessor = processor.VariableSet(cliLogger)
```
xoscar, line 42: preprocessor is now called just `processor`

cli/pkg/resourcemanager/apply.go
```diff
@@ -165,5 +165,13 @@ func (c Client) Apply(ctx context.Context, inputFile fileutil.File, requestedFor
 		}
 	}
 
+	if c.options.applyPostProcessor != nil {
+		var err error
```
xoscar, line 169: New post-processor implementation for apply

cli/processor/environment.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (e environment) Postprocess(ctx context.Context, input fileutil.File) (fileutil.File, error) {
+	var env openapi.EnvironmentResource
+	err := yaml.Unmarshal(input.Contents(), &env)
+	if err != nil {
+		e.logger.Error("error parsing test suite", zap.String("content", string(input.Contents())), zap.Error(err))
+		return input, fmt.Errorf("could not unmarshal test suite yaml: %w", err)
+	}
+
+	if env.GetSpec().Id != nil {
```
xoscar, line 35: Switch to new env

cli/processor/environment.go
```diff
@@ -0,0 +1,87 @@
...
+		err = e.updateEnvFn(ctx, *env.GetSpec().Id)
+		if err != nil {
+			return input, fmt.Errorf("could not update environment: %w", err)
+		}
+	}
+
+	if env.GetSpec().Resources != nil {
+		err = e.mapResources(ctx, input, *env.GetSpec().Resources)
```
xoscar, line 43: And create new resources

cli/cmd/resources.go
```diff
@@ -57,7 +57,7 @@ var (
 
 	envTokensClient = resourcemanager.NewClient(
 		httpClient, cliLogger,
-		"token", "environmenttokens",
+		"environmenttoken", "environmenttokens",
```
jorgeepc, line 60: do we want this change? same for `organizationinvite`?

xoscar: So, we need to either change the cloud BE to use the same nomenclature of the CLI or keep it like this, the problem is that the registry won't find the client when the definition is `type: OrganizationInvite` or `type: EnvironmentToken` as it is registered as `token` and `invite`.

We might need to change that tho

xoscar: I think for the environment tokens we need to do this, but for invites, it should work!

jorgeepc: 🤔 Got it! I would prefer to match the cmd with the resource type (`EnvironmentToken` and `Invite`). I'm not sure if changing the cloud BE at point makes sense.

jorgeepc (approved):

PR #3861, opened 2024-05-21
fix(cli): agent connection

xoscar:
This PR fixes the agent connection using the new features, it currently fails when connecting an agent using the `start` command without an API key due to a change on the environment endpoint and the lack of the API key on the response.

## Changes

- Uses the agent endpoint to grab the key
- Removes the usage of the env id when fetching the agent token from an environment

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


agent/runner/runner.go
```diff
@@ -96,17 +96,17 @@ func (s *Runner) onStartAgent(ctx context.Context, cfg config.Config) {
...
 	}
 
-	if env.AgentApiKey == "" {
+	if agentToken == "" {
 		s.ui.Error("You are attempting to start the agent in an environment you do not have admin rights to. Please ask the administrator of this environment to grant you admin rights.")
 		return
 	}
 
-	err = s.StartAgent(ctx, cfg.AgentEndpoint, env.AgentApiKey, cfg.UIEndpoint, cfg.EnvironmentID)
+	err = s.StartAgent(ctx, cfg.AgentEndpoint, agentToken, cfg.UIEndpoint, "")
```
xoscar, line 109: The agent tokens fetched by the agent API are already tied to an env, there is no need to add it here, if we add it it will break due to a control plane validation for organization tokens

agent/runner/agent.go
```diff
@@ -0,0 +1,56 @@
...
+	"net/http"
+
+	"go.uber.org/zap"
+)
+
+type agentResponse struct {
+	Key string `json:"key"`
+}
+
+func (s *Runner) getAgentToken(ctx context.Context, endpoint, orgID, envID, token string) (string, error) {
```
xoscar, line 17: Instead of using the environment endpoint, we grab the agent token from its own API

schoren (approved):

PR #3874, opened 2024-05-27
fix(agent): datastore headers proto type

xoscar:
This PR fixes the header type for data store, instead of using `HTTPHeader` with a key-value pair struct, using the `map[string]string` to make it work

## Changes

- Updates the HTTP header types

## Fixes

- Fixes connection to DS when headers were required (Grafana Cloud Tempo)

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


agent/proto/orchestrator.proto
```diff
@@ -315,7 +315,7 @@ message SumoLogicConfig {
 
 message HttpClientSettings {
     string url = 1;
-    repeated HttpHeader headers = 2;
```
xoscar, line 318: This is wrong, the http triggers use a struct like:
```
{
 key: string,
 value: string,
}
```

but Datastores use `map[string]string`

mathnogueira (approved):

PR #3895, opened 2024-06-04
fix(cli): adding validation when trying to run resource that does not exist

danielbdias:
This PR adds a validation on the CLI to inform the user that they tried to run a test / testsuite that doesn't exist. 

Example:
image

## Changes

- Add validation on `run` command to check if definition files exists

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

### DIFF
--- a/cli/cmd/resource_run_cmd.go
+++ b/cli/cmd/resource_run_cmd.go
@@ -20,7 +20,7 @@ var (
 	runCmd    *cobra.Command
 )
 
-const ExitCodeResourceNotFound = -1
+const ExitCodeResourceNotFound = 3
 
 func init() {
 	runCmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


xoscar:
@danielbdias I think this would be a good candidate for creating a small video recording the new CLI behavior, what do you think?

cli/cmd/resource_run_cmd.go
```diff
@@ -102,3 +121,20 @@ func validRequiredGatesMsg() string {
 
 	return "valid options: " + strings.Join(opts, ", ")
 }
+
+func hasValidResourceFiles() bool {
+	if len(runParams.DefinitionFiles) == 0 {
```
xoscar, line 126: Should an empty list be valid?

xoscar (commented):
Hey @danielbdias I left a small comment

cli/cmd/resource_run_cmd.go
danielbdias: Yes, in the case of the OSS test. (We can use --id for that case.)
Perhaps we should have something to validate when using run groups, too?

danielbdias:
> @danielbdias I think this would be a good candidate for creating a small video recording the new CLI behavior, what do you think?

@xoscar instead of adding a video, I added a small image showing the behavior, does it make sense?

xoscar (approved):

PR #3991, opened 2024-08-28
fix: oauth path building

mathnogueira:
This PR fixes how we build oauth endpoint paths by replacing `fmt.Sprintf` with `path.Join`

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

### DIFF
--- a/cli/pkg/oauth/oauth.go
+++ b/cli/pkg/oauth/oauth.go
@@ -6,7 +6,7 @@ import (
 	"io"
 	"net"
 	"net/http"
-	"path"
+	neturl "net/url"
 	"sync"
 
 	"github.com/kubeshop/tracetest/cli/ui"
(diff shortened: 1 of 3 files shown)


xoscar (approved):

cli/pkg/oauth/oauth.go
```diff
@@ -58,7 +59,7 @@ func (s *OAuthServer) GetAuthJWT() error {
 		return fmt.Errorf("failed to start oauth server: %w", err)
 	}
 
-	loginUrl := fmt.Sprintf("%soauth?callback=%s", s.frontendEndpoint, url)
+	loginUrl := path.Join(s.frontendEndpoint, fmt.Sprintf("oauth?callback=%s", url))
```
schoren, line 62:
```suggestion
	loginUrl := url.JoinPath(s.frontendEndpoint, fmt.Sprintf("oauth?callback=%s", url))
```

schoren (approved):

cli/pkg/oauth/oauth.go
mathnogueira: I had to create an alias for `url` as there is already an `url` var in this scope.

`neturl net/url`

PR #4008, opened 2024-09-03
fix: add Host to request in trigger

mathnogueira:
This PR makes both agent and opensource triggerer add the `Host` property to the `http.Request` before sending it to the server.

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

## Loom video

### DIFF
--- a/agent/workers/trigger/http_test.go
+++ b/agent/workers/trigger/http_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 
 	"github.com/kubeshop/tracetest/agent/workers/trigger"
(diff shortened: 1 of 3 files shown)


agent/workers/trigger/http_test.go
```diff
@@ -8,7 +8,6 @@ import (
 	"testing"
 
 	"github.com/kubeshop/tracetest/agent/workers/trigger"
-	triggerer "github.com/kubeshop/tracetest/agent/workers/trigger"
```
mathnogueira, line 11: those were the same import, but one was named.

danielbdias (approved):

PR #4022, opened 2024-09-17
feat: make spans fit an envelope size

mathnogueira:
This PR makes all polling responses to be added to an envelope and to fit a certain size. It tries to fit as many spans as possible in the envelope before sending it to the server. In case a span is let out of the envelope, it will be sent in the next polling iterations.

## Fixes

- grpc was breaking due to max message size

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [x] added a test

## Loom video

I could not record a loom because it is not supported in firefox :cry: But here is a screencast. In this case, I was running the agent in debug mode and an envelope max size of `5kb`, this is why you see spans increasing slowly. By default, we are going to use `3MB`, which would be more than enough for real-life spans.

[Screencast from 2024-09-17 16-56-18.webm](https://github.com/user-attachments/assets/b44ccf83-97e3-4eff-ab01-d933019e1ce0)

### DIFF
--- a/agent/workers/envelope/spans_envelope.go
+++ b/agent/workers/envelope/spans_envelope.go
@@ -1,0 +1,42 @@
+package envelope
+
+import (
+	agentProto "github.com/kubeshop/tracetest/agent/proto"
+	"google.golang.org/protobuf/proto"
+)
+
+// EnvelopeSpans get a list of spans and batch them into a packet that does not
+// surpasses the maxPacketSize restriction. When maxPacketSize is reached, no
+// more spans are added to the packet,
+func EnvelopeSpans(spans []*agentProto.Span, maxPacketSize int) []*agentProto.Span {
+	envelope := make([]*agentProto.Span, 0, len(spans))
+	currentSize := 0
+
+	// There's a weird scenario that must be covered here: imagine a span so big it is bigger than maxPacketSize.
+	// It is impossible to send a span like this, so in this case, we classify those spans as "large spans" and we allow
+	// `largeSpansPerPacket` per packet.
+	//
+	// It is important to ensure a limit of large spans per packet because if your whole trace is composed by
+	// large spans, this would mean a packet would hold the entiry trace and we don't want that to happen.
+	const largeSpansPerPacket = 1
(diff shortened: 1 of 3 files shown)


agent/workers/poller.go
```diff
@@ -24,6 +25,8 @@ import (
 	"go.uber.org/zap"
 )
 
+const maxEnvelopeSize = 4 * 1024 * 1024 // 4MB
```
mathnogueira, line 28: I wonder if we should increase this value. What do you think @xoscar, @schoren and @danielbdias ?

agent/workers/envelope/spans_envelope.go
```diff
@@ -0,0 +1,42 @@
...
+// EnvelopeSpans get a list of spans and batch them into a packet that does not
+// surpasses the maxPacketSize restriction. When maxPacketSize is reached, no
+// more spans are added to the packet,
+func EnvelopeSpans(spans []*agentProto.Span, maxPacketSize int) []*agentProto.Span {
+	envelope := make([]*agentProto.Span, 0, len(spans))
+	currentSize := 0
+
+	// There's a weird scenario that must be covered here: imagine a span so big it is bigger than maxPacketSize.
+	// It is impossible to send a span like this, so in this case, we classify those spans as "large spans" and we allow
+	// `largeSpansPerPacket` per packet.
```
xoscar, line 17: This is a great point, I believe there is an issue around this very same subject in one of the otel collector repos, I'm glad you thought about it too!

agent/workers/poller.go
xoscar: You mean increasing the value server side too?

xoscar (approved):
Great job!

agent/workers/poller.go
mathnogueira: No, just here. The server is already accepting a large-enough message. Increasing the message here means that in we can send more spans per polling cycle.

mathnogueira: We discussed and the limit on the server is 4MB, so I'm setting this to be `3MB` to prevent an overflow in case of a large span as discussed in the code.

PR #4026, opened 2024-09-19
chore(tempodb): Adding max size for received messages to 10MB

xoscar:
This PR temporary increases the max size of messages to 10MB, while we work on a way to paginate results from Tempo

## Changes

- Adds max received option to 10MB

## Fixes

- https://github.com/kubeshop/tracetest-cloud/issues/924

## Checklist

- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test


mathnogueira (approved):
LGTM. It's worth noting that the tempo server must be changed as well. So if we find any problem like that in prod, we need to ask the user to increase the max send size of the server to match our agent max size.

